PR #78, opened 2014-04-17
Fix for Issue 72 - projection recursion

### LINKED ISSUE #72: Projected Fit infinite recursion
If you project fit A onto fit B, then project fit B onto fit A, this will cause the fits to be recursively projected onto each other infinitely eventually hitting Python's recursion limit. This will break both fits until one is deleted, as well as any fit that either fit is projected onto.

Best way to try this is to create two new blank fittings and project each one onto the other.

blitzmann:
Fixes #72 - projection recursion. Creates a flag to turn on/off clearing of projected fit depending on who's calling the `clear`ing, breaking the recursion.

### DIFF
--- a/eos/saveddata/fit.py
+++ b/eos/saveddata/fit.py
@@ -727,7 +727,7 @@ class Fit(object):
         del self.__extraDrains[:]
 
         if self.ship is not None: self.ship.clear()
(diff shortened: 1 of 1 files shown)


blitzmann:
I used this fix in my recent work on this, but I'm just gonna close this

PR #232, opened 2014-12-19
Port custom wait dialog to wx.ProgressDialog

blitzmann:
I never liked the "in progress" animation for pyfa. It doesn't give you any information, and just doesn't look great. wxPython comes with a handy wx.ProgressDialog that can replace it.

Before (fit backup/import):
processing

After (backup/HTML export):
progress

After (importing):
processing2

waitDialog is still scattered around - most predominately while opening fits from previous pyfa session, but also when importing character and exporting skills needed (however these functions don't really take any time to do, so the wait dialog disappears so quickly). I am probably going to leave those in for now.

Need to test a bit more on Linux and OS X - from what I've tried it works great with importing multiple files and export as well, just need to thoroughly test.


blitzmann:
Tested on OS X, works well.

PR #241, opened 2015-01-02
Make tab tooltips optional

cockroach:
The current tooltips can be a bit irritating in some cases, eg. when using a tiling window manager which changes its layout when a tooltip appears. This branch fixes that for me by making the tooltips optional.


blitzmann:
Interesting.  I will take a look at this later next week when i get back holiday. What exactly happens with your window manager? Mind providing a screenshot?

DarkFenX:
They are also rather annoying for me (huge, rather sluggish and cause additional system-level popups on gnome3, because they are implemented as app windows which appear out of focus). If many people dislike them - might even consider disabling them as default.

blitzmann:
I didn't know they caused such issues on Linux. I've never really used them, nor do I really like them or see the value in them. I wonder how many folks actually use the information it contains... I will merge but keep it opt-out for now.

PR #242, opened 2015-01-04
Fix #233

### LINKED ISSUE #233: Selecting the Last Damage Pattern Does Nothing
This Bug seems to have cropped up recently. Selecting the last incoming damage pattern does nothing. If a fit is saved with the last pattern then the resists will show 0% for all damage types in the pattern and show invalid ehp.

blitzmann:
Fix for #233, need to confirm on Linux and OS X


blitzmann:
Confirmed, merged

PR #247, opened 2015-01-11
Implement GUI exception reporting, resolves #162

### LINKED ISSUE #162: Better error reporting
I have never seen pyfa's error logging actually work before. I went looking a little deeper, and quite honestly I don't know why pyfa doesn't log errors or output to their respective files. 

However, I found this gem in `wx.App()`:

> Parameters:
>     redirect - Should sys.stdout and sys.stderr be redirected? Defaults to True on Windows and Mac, False otherwise. If filename is None then output will be redirected to a window that pops up as needed. (You can control what kind of window is created for the output by resetting the class variable outputWindowClass to a class of your choosing.)

Interestingly enough, pyfa has this set to `False` (assuming because we wanted to log to file rather than have wxPython manage it). I think wx might take over `stderr` and, since this parameter is set to false, it resets any value we might have had before (namely, the actual log file).

So I set it to True and raised an exception:

stderr

I think this is exactly what we need - it will allow the user to copy/paste error message. And we can set the pop-up to a class of our own and customize the look a bit.

We don't want this to completely override `stderr`, especially for development. Windows and Mac release should definitely have this as they aren't normally run through a terminal. How about Linux? Should we provide a mechanism to enable this on Linux for those users who may use package software to install (and thus don't run in terminal?) How should this work?

blitzmann:
screenshot 2014-12-18 01 10 02

This will catch all unhandled exceptions that may crop up and present the user with it. Before, pyfa would simply crash or otherwise not work properly. When the exception is caught, it is sent to both the original `sys.stderr` (which will allow IDEs and debuggers to continue to work) and to the new error window. Fairly simple. wxPython has a similar implementation that this is based on.

This is simply a pull request to track current development. The following still must be completed before this is merged:
- Thoroughly test out on all platforms
- Rebuild the Windows skeleton with new pyfa.py from pull request, or modify distributive scripts to dynamically build skeletons. I believe OS X skeleton doesn't need modification

### DIFF
--- a/gui/errorDialog.py
+++ b/gui/errorDialog.py
@@ -1,0 +1,128 @@
+#===============================================================================
+# Copyright (C) 2010 Diego Duclos
+#
+# This file is part of pyfa.
+#
+# pyfa is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# pyfa is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with pyfa.  If not, see <http://www.gnu.org/licenses/>.
+#===============================================================================
+
+import wx
+import sys
+import gui.utils.fonts as fonts
+
+class ErrorFrame(wx.Frame):
+
+    def __init__(self, parent):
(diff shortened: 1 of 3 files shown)


kainz:
This is a neat idea, but why not use python logging? (yes, i know thats a bigger change)

blitzmann:
Well,  using explicit log messages (such as info, debug, warnings, etc) means to implement all throughout the source. pyfa wasn't built with logging in mind, and thus it tried to just dump output to a file. Problem is that wxPython takes over stdout/err so file logging wasn't working. Proper logging support will hopefully be included in the pyfa rewrite.

The goal of this feature wasn't to log messages, but rather to alert the user that something happened. At the moment, if an exception is raised in the program, pyfa simply crashes or the function doesn't work as intended, without giving the user any information to report. At least with this, the exception should pop up so that the user can report it and give us a little more information.

blitzmann:
I should really get this tested and implemented at some point

blitzmann:
Closing this because lol this isn't going to happen without some fundamental changes anyway. We'll see how beneficial the logging PR goes, and we can possibly hook into that to spawn an exception window

Issue #261, opened 2015-02-13
Activating/deactivating missile launcher modules affects capacitor stability when displaying a percentage


OISumeko:
When fiddling with a fit, I noticed that when I toggled the fitted 'Rocket Launcher II' module between active and inactive, the capacitor stability reading changed. This struck me as odd, as (to the extent of my knowledge) these modules shouldn't affect capacitor. The fit I was using when I first noticed this is as follows:

    [Jaguar, SampleFit]
    
    Gyrostabilizer II
    Internal Force Field Array I
    Micro Auxiliary Power Core I
    Overdrive Injector System II
    
    J5b Phased Prototype Warp Scrambler I
    Limited 1MN Microwarpdrive I
    Medium Ancillary Shield Booster, Navy Cap Booster 50
    Medium F-S9 Regolith Shield Induction
    
    150mm Light AutoCannon II, Republic Fleet Phased Plasma S
    150mm Light AutoCannon II, Republic Fleet Phased Plasma S
    150mm Light AutoCannon II, Republic Fleet Phased Plasma S
    Rocket Launcher II, Caldari Navy Nova Rocket
    
    Small Anti-Kinetic Screen Reinforcer I
    Small Projectile Burst Aerator II

With all modules enabled, and the 'All 5' character selected, the capacitor reads as Stable at 32.5%. When the Rocket Launcher is toggled off, the capacitor reads as Stable at 36.7%. The Capacitor total, as well as recharge/discharge rates do not change.

I tried multiple characters with the same fit, and found the same effect occurred when the stability was displayed as a percentage. However, if the stability was displayed as MM:SS (eg. 4m0s), then no change was seen when toggling the module. 

I tried fitting up a crow in a similar fashion, and encountered the same issue. In addition I fit up a caracal, and tried all available MIssile Launcher varieties (LML, RLML, TP, etc) as well as a number of named varieties of launchers -- all of these exhibit the same issue.

I also checked with a number of fits to see if this behaviour was present with projectile turrets, and did not note any issues, so this issue does seem to be limited to launchers.

blitzmann:
You seem to be describing #126, however that should have been fixed a while ago. I have not yet tested your fit or reproduction steps yet (thank you for being so detailed). In the meantime, can you confirm that you are using the latest pyfa version (v.1.8.1).

OISumeko:
I am using the latest release (v. 1.8.1) on Windows 7, 64 bit. Sorry for missing that on the original report!

Issue #211, opened 2014-12-02
Cap Boosters behaving oddly


JoshuaNotowitz:
i noticed this unusual behavior while fitting a dread:

With both cap boosters running the cap lasts: 7m18s
![screen shot 2014-12-01 at 8 25 22 am](https://cloud.githubusercontent.com/assets/8390690/5249897/f35d875a-793b-11e4-8ea2-901185e68472.png)

With one cap booster OH the caps lasts longer as expected at: 8m48s
![screen shot 2014-12-01 at 8 25 47 am](https://cloud.githubusercontent.com/assets/8390690/5249895/f35bf0ac-793b-11e4-8552-342ee78ce5e0.png)

Unusually with both cap boosters OH the cap lasts less that is does with one OH even though the cap regen is higher the cap only lasts:8m20s
![screen shot 2014-12-01 at 8 26 00 am](https://cloud.githubusercontent.com/assets/8390690/5249896/f35cdf4e-793b-11e4-8ec8-aaeaa6a8d909.png)

[edit-blitzmann] EFT:

    [Moros, Moros fit]

    Energized Adaptive Nano Membrane II
    Energized Adaptive Nano Membrane II
    Capital Inefficient Armor Repair Unit
    Magnetic Field Stabilizer II
    Magnetic Field Stabilizer II
    Magnetic Field Stabilizer II
    Damage Control II

    Heavy Capacitor Booster II, Navy Cap Booster 800
    Heavy Capacitor Booster II, Navy Cap Booster 800
    Tracking Computer II, Optimal Range Script
    Tracking Computer II, Optimal Range Script
    Sensor Booster II, Scan Resolution Script

    Siege Module II
    Limited Mega Ion Siege Blaster I, Shadow Antimatter Charge XL
    Limited Mega Ion Siege Blaster I, Shadow Antimatter Charge XL
    Limited Mega Ion Siege Blaster I, Shadow Antimatter Charge XL

    Capital Trimark Armor Pump I
    Capital Trimark Armor Pump I
    Capital Trimark Armor Pump I

OISumeko:
Disabling stagger on the cap simulator fixes this issue, giving it the expected behaviour in-line with what other fitting tools suggest. However disabling this does adversely affect some other fits, giving them much more pessimistic cap lives. 

[...]

PR #263, opened 2015-02-22
Fix for #261

OISumeko:
This addresses #261. Launchers do not have a "capacitorNeed" attribute (whilst projectile weapon systems do), so getModifiedItemAttr will return None rather than 0 when checking for cap-affecting modules. Adding an explicit check for None fixes the issue with the fit presented in the report.

### DIFF
--- a/eos/saveddata/fit.py
+++ b/eos/saveddata/fit.py
@@ -737,7 +737,7 @@ class Fit(object):
         capAdded = 0
         for mod in self.modules:
             if mod.state >= State.ACTIVE:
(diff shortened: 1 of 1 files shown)


blitzmann:
Nice find. However, I would prefer this syntax: `if (mod.getModifiedItemAttr("capacitorNeed") or 0) != 0:`. If you can change it, I'll go ahead and merge it. =D

OISumeko:
Updated with the requested change.

blitzmann:
Thanks =)

If you want to poke around at some more stuff, #211 has baffled me. =P

PR #276, opened 2015-04-22
Support fits exported from jEveAssets

kainz:
Support importing of fits exported from jEveAssets by implementing support for loading utf-16 XML.

Such XML is generated by software like jEveAssets's owned ships->fitting
export tool.

Without such detection, pyfa will go and try to import those as DNA
fits, with all the ensuing hilarity, thus the DNA import debug code as
well.

### DIFF
--- a/pyfa.py
+++ b/pyfa.py
@@ -78,6 +78,10 @@ if __name__ == "__main__":
         config.saveInRoot = True
     config.defPaths()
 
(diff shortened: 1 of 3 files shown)


DarkFenX:
Thank you. Looks clean.

Especially some logging is nice, that's where pyfa definitely doesn't shine.

Issue #277, opened 2015-04-27
Compatibility Issues after python2-sqlalchemy 1.0.0


MagicFish1990:
SQL processing failed, without any errors.

DarkFenX:
Sounds weird. I've been playing a bit with it (on windows, after merging wx3 branch) and 1.0.2 worked just fine for me.

DarkFenX:
Please specify at least exact version of sql alchemy and OS you're using.

MagicFish1990:
pyfa 1.11.0.r16.gbb8d5d0
python2-sqlalchemy 1.0.0-1.0.2
Archlinux

MagicFish1990:
Part of error

Traceback (most recent call last):
  File "/usr/share/pyfa/gui/sfBrowserItem.py", line 318, in OnLeftUp
    btn.DoCallback()
  File "/usr/share/pyfa/gui/sfBrowserItem.py", line 49, in DoCallback
    self.callback()
  File "/usr/share/pyfa/gui/shipBrowser.py", line 1608, in copyBtnCB
    self.copyFit()
  File "/usr/share/pyfa/gui/shipBrowser.py", line 1612, in copyFit
    fitID = sFit.copyFit(self.fitID)
  File "/usr/share/pyfa/service/fit.py", line 184, in copyFit
    eos.db.save(newFit)
  File "/usr/share/pyfa/eos/db/saveddata/queries.py", line 415, in save
    commit()
  File "/usr/share/pyfa/eos/db/saveddata/queries.py", line 426, in commit
    saveddata_session.commit()
  File "/usr/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 790, in commit
    self.transaction.commit()
  File "/usr/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 392, in commit
    self._prepare_impl()
  File "/usr/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 372, in _prepare_impl
    self.session.flush()
  File "/usr/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 2004, in flush
    self._flush(objects)
  File "/usr/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 2122, in _flush
    transaction.rollback(_capture_exception=True)
  File "/usr/lib/python2.7/site-packages/sqlalchemy/util/langhelpers.py", line 60, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
[...]

Issue #274, opened 2015-04-17
Pyfa crashes at start: "ValueError: None is not a valid value for ownerID"


mipesom:
```bash
Traceback (most recent call last):
  File "pyfa.py", line 97, in <module>
    MainFrame()
  File "/home/mipesom/Apps/pyfa/gui/mainFrame.py", line 132, in __init__
    self.fitMultiSwitch = MultiSwitch(self.FitviewAdditionsPanel)
  File "/home/mipesom/Apps/pyfa/gui/multiSwitch.py", line 26, in __init__
    gui.chromeTabs.PFNotebook.__init__(self, parent)
  File "/home/mipesom/Apps/pyfa/gui/chromeTabs.py", line 112, in __init__
    self.tabsContainer = PFTabsContainer(self, canAdd=canAdd)
  File "/home/mipesom/Apps/pyfa/gui/chromeTabs.py", line 682, in __init__
    self.sFit = service.Fit.getInstance()
  File "/home/mipesom/Apps/pyfa/service/fit.py", line 78, in getInstance
    cls.instance = Fit()
  File "/home/mipesom/Apps/pyfa/service/fit.py", line 85, in __init__
    self.character = Character.getInstance().all5()
  File "/home/mipesom/Apps/pyfa/service/character.py", line 162, in all5
    eos.db.commit()
  File "/home/mipesom/Apps/pyfa/eos/db/saveddata/queries.py", line 426, in commit
    saveddata_session.commit()
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/session.py", line 790, in commit
    self.transaction.commit()
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/session.py", line 392, in commit
    self._prepare_impl()
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/session.py", line 372, in _prepare_impl
    self.session.flush()
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/session.py", line 2004, in flush
    self._flush(objects)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/session.py", line 2122, in _flush
    transaction.rollback(_capture_exception=True)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/util/langhelpers.py", line 60, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
[...]

PR #292, opened 2015-05-27
fix issue copying fits with sqlalchemy 1.0

wolfwood:
addresses issue #274 / #277


wolfwood:
a quick git grep suggests there are no other places in the code where '"ownerID" : lambda val: isinstance(val, int)' occurs

PR #318, opened 2015-07-08
Fix logging

poettler-ric:
If started in a new installation savePath doesn't exist and crashes when the logging is configured.
I just moved the creation of savePath into a private method and used it before configuring logging.


blitzmann:
If this works, I <3 you. Double points if it works with the trouble I had with it on the mac as well. =D

Will look into it and merge today

PR #346, opened 2015-08-16
Context menu to swap modules for different variation

Will-W:
In the true spirit of open source - if you want a feature write it yourself.

This PR adds a right-click context menu (for modules) that brings up a list of meta-variations for the selected item(s). Selecting an item changes those modules to the selected meta-variation.
If multiple items are selected, menu only shows up if the variation list would be the same.

Saves a lot of clicking to meta-up or down a module to fit within skills or power budget etc.

I had a good look at the existing code, but I'm sure there are some style or architectural things you'll want me to fix.

(edit by blitzmann) Screenshot of how it looks:

untitled

### DIFF
--- a/gui/builtinContextMenus/metaSwap.py
+++ b/gui/builtinContextMenus/metaSwap.py
@@ -6,12 +6,6 @@ import service
 import wx
 import gui.globalEvents as GE
 
-# TODO:
-# Handle multiple selection better
-# Icons?
-# Submenu for officer?
-# Officer/Deadspace sorting
-
 class MetaSwap(ContextMenu):
     def __init__(self):
         self.mainFrame = gui.mainFrame.MainFrame.getInstance()
@@ -46,10 +40,15 @@ class MetaSwap(ContextMenu):
         def get_metalevel(x):
             return x.attributes["metaLevel"].value
 
+        def get_metagroup(x):
+            return x.metaGroup.ID if x.metaGroup is not None else 0
+
         m = wx.Menu()
 
+        # Sort items by metalevel, and group within that metalevel
         items = list(self.variations)
         items.sort(key=get_metalevel)
+        items.sort(key=get_metagroup)
 
         group = None
         for item in items:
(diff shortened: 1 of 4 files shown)


Will-W:
By the way - only tested on Windows, Can test on Linux tomorrow, have no access to a Mac.

gui/builtinContextMenus/metaSwap.py
```diff
@@ -0,0 +1,91 @@
...
+        items.sort(key=get_metalevel)
+        items.sort(key=get_metagroup)
+
+        group = None
+        for item in items:
+            # Apparently no metaGroup for the Tech I variant:
+            if item.metaGroup is None:
+                thisgroup = "Tech I"
+            else:
+                print item.metaGroup.ID
```
lunedis: Seems to work fine, I like it a lot. The context menu can really increase the speed you can work with pyfa.

Btw you still have a "print" in gui/builtinContextMenus/metaSwap.py:59

lunedis:
Seems to work fine, I like it a lot. The context menu can really increase the speed you can work with pyfa.

Btw you still have a "print" in https://github.com/Will-W/Pyfa/blob/master/gui/builtinContextMenus/metaSwap.py#L59

Will-W:
Ewwwww. That was careless. Removed.

Will-W:
There is an issue on Linux (as I was selecting from looking at the ammo selection code).
Will try and get that done later today.

blitzmann:
Hi,
Just tested this out, and I like it. Can you elaborate more on Linux issues? IIRC, ammo code (and any other 'sublist' selection) had issues between platforms simply due to bugs in the way wxPython handles events. I cannot remember how this manifests of how we've delt with it in the past, heh. Let me know!

Also, I can test in OS X once you submit linux fix

Will-W:
I only did a quick test to confirm, but in Linux the click event in the sub menu never gets handled. I think on Linux you have to register with the specific submenu but on msw you need to register with the root menu. That's what the ammo switcher code appeared to be doing, and on Linux clicking a module in a submenu doesn't do anything.

Can look at tonight.

Will-W:
Have now tested on Linux and Windows and it looks good. If you could test it on a Mac, @blitzmann, that would be great.

blitzmann:
Eh, my os x python is all fucked. I'll go ahead and merge this now and test with OS X once I can get it reinstalled correctly

PR #382, opened 2015-10-13
Add a simple usage note to projected view

Will-W:
Had a few people in corp not realise you could put a whole ship fitting in the 'Projected' view. Thought a basic usage hint might help.

cropped_screenshot

### DIFF
--- a/gui/projectedView.py
+++ b/gui/projectedView.py
@@ -27,18 +27,18 @@ from gui.contextMenu import ContextMenu
 import eos.types
 
 class ProjectedViewDrop(wx.PyDropTarget):
-        def __init__(self, dropFn):
-            wx.PyDropTarget.__init__(self)
-            self.dropFn = dropFn
-            # this is really transferring an EVE itemID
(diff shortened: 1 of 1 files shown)


Will-W:
Tested on Linux - wxGTK 2.8, and on Windows, both 2.8 and 3.0

blitzmann:
I have been wanting to do this ever since I added that note to the fleet thing, the only thing stopping me was the fact that the other views used the Display while the fleet views straight up used it's own panel. 

I can't check the code right now, but it seems like you insert the message into the Name column, yeah? I think a better solution would be to append the message to the bottom of the Display class, but it would take more work because you'll have to make a new panel and add the Display along with the message (in theory).

This looks like a good first step tho and I'll be happy to merge it once I have a chance to take a look. =)

Will-W:
I had a little think about the best way of doing this. Started just with tooltip but decided it wasn't obvious enough.

The nice thing about this approach is that it's really obvious without consuming any screen real-estate. It's also a metaphor that users will be familiar with: "drag thing to here".

PR #385, opened 2015-10-20
Check that python-dateutil is installed at startup

minlexx:
Check that python-dateutil is installed at startup in pyfa.py, same way as for wxPython and SQLAlchemy (except for version check).


blitzmann:
Thanks!

PR #390, opened 2015-10-31
Added option for disabling capacitor simulation staggering for certain modules

OISumeko:
Playing with the cap simulator, I noticed that turrets of the same type stack and have staggering applied when running through the cap simulator. This is not realistic behaviour as most (all?) pilots will
have grouped their turrets in-game, meaning they will all activate simultaneously rather than activating individually, offset from each other.

This change introduces a 'disable stagger' field to the drain tuple that is passed through to the cap simulator, allowing us to disable staggering on certain modules. This will enable us to add GUI options in the future to allow users to choose whether to stagger certain modules (eg, cap boosters, neuts) or have them all fire simultaneously. Currently this defaults to False (existing behavior) for all modules except for turrets, which will now behave more like in-game turrets for cap simulation purposes.

### DIFF
--- a/eos/capSim.py
+++ b/eos/capSim.py
@@ -59,7 +59,7 @@ class CapSimulator(object):
         return duration, capNeed
 
     def init(self, modules):
-        """prepare modules. a list of (duration, capNeed, clipSize) tuples is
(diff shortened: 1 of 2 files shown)


OISumeko:
Knew there had to be a better way :smile: That has been updated.

PR #414, opened 2015-11-07
Add "requests" to README.md

andrewd18:
CREST support requires the requests module which may not be included by default in all users' distributions.


blitzmann:
Thanks =)

PR #429, opened 2015-11-13
Added detection of Unicode BOM (Byte Order Mark) ...

Neugeniko:
To determine the encoding used in imported files.


blitzmann:
Yjsmld!

PR #472, opened 2015-12-10
Fixed display of neut and nos amount, and guidance disruptors

wendazhou:
Hi,

noticed that some stuff has been broken in the last patch, so gave a try at fixing them.

First issue: CCP renamed some of the groups in the database, so neuts, nos and tracking
disruptors were not displaying correctly anymore. This fixes that (simple name change).

Second issue: new guidance disruptors use a new set of attributes, this is now displayed
correctly in the miscellaneous column. (However, could maybe be more useful, e.g.
instead of computing missile velocity + flight time penalty, give combined range
penalty?)

I've included a screenshot of the change

Current:
current

The fix:
with fix

Hope it's useful. Love Pyfa by the way, I've found really useful.

### DIFF
--- a/gui/builtinViewColumns/misc.py
+++ b/gui/builtinViewColumns/misc.py
@@ -85,7 +85,7 @@ class Miscellanea(ViewColumn):
                 if n > 0:
                     info.append("{0}{1}".format(n, slot[0].upper()))
             return "+ "+", ".join(info), "Slot Modifiers"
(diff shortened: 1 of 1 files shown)


blitzmann:
Thanks so much, this is really useful and I totally missed these bugs. Also, you mentioned

> However, could maybe be more useful, e.g. instead of computing missile velocity + flight time penalty, give combined range penalty?

Could you elaborate more on that?

PR #497, opened 2016-01-12
Added commandline arguments to specify window title and path for savefiles

SpeedProg:
*  Settable Window Title using commandline arguments
  *  helps to distinguish between different instances
*  Settable savefile path using comandline arguments
  * allows for multiple instances to not mix the fittings/characters

I personally needed these changes because I use pyfa to check hundreds of fits a day, and having them mixed with my personal fits was a pain! (Deleting even more after that confirmation box was introduced)

### DIFF
--- a/config.py
+++ b/config.py
@@ -60,7 +60,7 @@ def __createDirs(path):
     if not os.path.exists(path):
         os.makedirs(path)
 
-def defPaths():
+def defPaths(savepath):
     global debug
     global pyfaPath
     global savePath
@@ -87,8 +87,11 @@ def defPaths():
     else:
         savePath = getattr(configforced, "savePath", None)
         if savePath is None:
-            savePath = unicode(os.path.expanduser(os.path.join("~", ".pyfa")),
+            if savepath is None: # savepath is not overriden
+                savePath = unicode(os.path.expanduser(os.path.join("~", ".pyfa")),
                                sys.getfilesystemencoding())
+            else:
+                savePath = savepath
 
     __createDirs(savePath)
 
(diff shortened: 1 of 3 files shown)


blitzmann:
> I personally needed these changes because I use pyfa to check hundreds of fits a day

Wait, what exactly are you doing that you need to check hundreds of fits per day? That's insane!

> (Deleting even more after that confirmation box was introduced)

Protip: Shift+click the delete to bypass the delete confirmation. =D

I'll take a look at this later on. Thanks!

SpeedProg:
well I fc/fleetcomp for incursions.
I know a lot of fits how they should look like to meet our tanking requirements but ppl are good at making weird tank mod compositions. So yeah if I do that for some hours a day there are easily a few hundred fits in pyfa :D

blitzmann:
Im intrigued. Why dont you just import the eft format and then delete it afterwards after checking? Do people supply you with their own database? 

Btw, took a quick look at the pull request, and it looks good. Ill test tonight and merge if no problems.

SpeedProg:
takes up to much time to delete the fit after checking in I don't really have more then a few seconds when it is a busy time.
So I normally just copy to clipboard in eve, paste to pyfa look at the values and move on :D

SpeedProg:
it might be that the command argument to unicode conversion of the title needs to happen in an other way to handle console specific (os specific) character sets.
I didn't know enough about python to do that :/

MonfortB:
It's probably not much but wouldn't it be better to rename your variable ```savepath``` to something like ```customPath``` in order to avoid any confusion with the existing global variable ```savePath```.  
Having two variables in the same method with names different only because of one uppercase letter doesn't strike me as a good idea.

SpeedProg:
xD might be a good idea to rename it to customSavePath yeah

PR #503, opened 2016-01-16
Use idiomatic osx preferences shortcut

resinneublem:
@blitzmann 

OSX uses "Command-," for application preferences


blitzmann:
Does it not work like this already? 

https://github.com/DarkFenX/Pyfa/issues/68 We closed an issue that notified us of this issue. The solution was to change the ID to a built in one, which allows the mac build to properly assign the hotkey

resinneublem:
Nah. Mac app is assigning `Command-P` to Preferences. This changes it to `Command-,`

blitzmann:
Maybe it's another difference of the wx3 backend? I dunno, but I know that it was previously addressed. I;ll take a look at this PR when I've got time to dabble in OSX and merge it. =)

Issue #368, opened 2015-09-27
Suggestion: New user setting to stop exporting ammo charges


resinneublem:
Often I wish I could export a fit without the loaded charges and only include stuff in the cargohold. Especially because I'll flip back and forth with ammo or scripts to see the numbers change.

I know you can empty the equipped charges before exporting, but that's a pain.

blitzmann:
So, instead of this

    [Hawk, 2x ASB]

    Damage Control II
    Co-Processor II

    1MN Afterburner II
    Medium Ancillary Shield Booster
    Warp Scrambler II
    Medium Ancillary Shield Booster
    Stasis Webifier II

    Rocket Launcher II, Caldari Navy Mjolnir Rocket
    Rocket Launcher II, Caldari Navy Mjolnir Rocket
    Rocket Launcher II, Caldari Navy Mjolnir Rocket
    Rocket Launcher II, Caldari Navy Mjolnir Rocket
    [Empty High slot]

    Small Anti-EM Screen Reinforcer I
    Small Anti-EM Screen Reinforcer I


    Scourge Rage Rocket x200
    Navy Cap Booster 50 x68

You want this:

    [Hawk, 2x ASB]

    Damage Control II
    Co-Processor II

    1MN Afterburner II
    Medium Ancillary Shield Booster
    Warp Scrambler II
    Medium Ancillary Shield Booster
    Stasis Webifier II

    Rocket Launcher II
    Rocket Launcher II
    Rocket Launcher II
    Rocket Launcher II
    [Empty High slot]

    Small Anti-EM Screen Reinforcer I
    Small Anti-EM Screen Reinforcer I


    Scourge Rage Rocket x200
    Navy Cap Booster 50 x68

ChrisKing517:
I second this!

resinneublem:
Sorry, I missed this notification. Yep that's exactly what I'm thinking.

It's not so bad with rockets, hybrids, or projectiles. It's mainly for lasers (don't need extra crystals) and cap boosters (sometimes has problems with cargo space).

PR #504, opened 2016-01-16
Make exporting charges a preference

resinneublem:
@blitzmann this is in reference to #368 

Just starting this PR off as a discussion, by all means let me know if there's anything you don't like about it.

This defaults the preference to the current settings so it shouldn't affect anyone unless they opt out of it.

For my specific use-case: I don't want to export equipped charges because it usually conflicts with the cargo hold (e.g. I don't need 2x of the same type of script or 2x of the same crystal).

I'm usually loading different charges to see the stats on the right.

I've also gotten caught with cap boosters in the cargo hold + equipped going over the actual cargo hold limit.

Let me know what you think.


resinneublem:
@blitzmann just bumping this to see if you have an opinion about it.

blitzmann:
Ah, sorry, forgot about this. I don't have a problem with the way it works now and will happily merge it, but I was envisioning an option on the actual export dialog that pops up rather than hidden in general settings. The problem with that is that it would require passing more arguments to the export functions, so it's a bit more involved.

I've actually been meaning to change up how import/export works and integrate the CREST export into the dialog, so maybe that would be the best time to do it.

resinneublem:
> but I was envisioning an option on the actual export dialog that pops up rather than hidden in general settings.

Ah no biggy man. Let me know if you get a working branch up and I'll try and move this around to accommodate.

Issue #262, opened 2015-02-20
Please sign source code tag in Git


AdamMajer:
Currently tags in git repository are not signed. I think it would be a benefit if the tags were cryptographically signed. You can do that with `git tag -s <tagname>`

blitzmann:
We use GitHub's web interface to create tags when we make a release, so it's not something that we do at the command level. What are the benefits of doing such a thing?

AdamMajer:
Signed tag essentially signs the entire archive as being authentic. No one can tamper with it.

I've noticed that it is possible to sign the tag, and then make a release with the GitHub interface. They do not seem to interfere. For example, I've made a Debian package for Pyfa and I was able to use a signed tag for that without problems.

  https://github.com/AdamMajer/Pyfa/releases

On another note, if you guys want, I can provide Debian package files for your releases.

DarkFenX:
I'm using git command line interface to create the tag. Will look into signing it next time i'm doing a release.

Regarding debian releases - would be cool, however i am not sure how we can integrate it to release everything together.

blitzmann:
Hey @AdamMajer,

With the v1.19.0 release, I went ahead and set up a key pair and signed the tag. Public key can be found here:

https://pgp.mit.edu/pks/lookup?op=vindex&search=0xA0702F954C53D05F

Maybe you can walk me through what else I can do? I know that keys are based on a web of trust, but I don't have any signatures on my public key considering it's brand new (and also that I'm somewhat new to actually using PK crypto). Let me know =)

AdamMajer:
Thank you very much! That's helpful.

Web of trusts is not as important here as just keeping that key consistent. So if you sign with same key going forward, that key is already imported for me, so I can check signatures and immediately know that next version was prepared/signed by same person (or at least same key). For example, no one at github can impersonate you anymore.

It's just a consistency check that makes tampering with repositories nearly impossible.

blitzmann:
Thanks. Lets hope I just remember to sign it for future releases. Been thinking of making a release script anyway that would automatically upload to github, so if that happens I can just integrate it with that hopefully. =)

PR #510, opened 2016-01-24
Debian specific files

AdamMajer:
These are debian specific files to make a .deb package. This then can be installed on Debian or derivatives or distributions that use .deb format (eg. Ubuntu).

I'm using Pyfa under Debian, and a package is useful.


blitzmann:
I am somewhat hesitant of merging this for the following reasons:

1) While pyfa supports Linux, I'm more inclined to keep the main repository distro-agnostic. This is because I cannot test most distro's, and honestly don't want to be responsible for maintaining packages for various distros. I do, however, support third-party packaging such as Arch's AUR or your .deb files.
2) I am not at all familiar with Debian packaging system to be comfortable including it in the repo right now. Besides adding debian-specific files to the repo so that user's can create their own package, what does this acomplish?

As a side note, and related to 2), I see the `debian/watch` file. How does this work? Does it automatically download and patch the package when a new release is available when updating the systems packages? Now that would be neat, but again, I don't know much about their system.

AdamMajer:
That's fine. If you don't mind, can you point Debian/Ubuntu users to my repository in your README.md? I maintain Debian package for pyfa, which should work fine with Ubuntu and others. RPM is on SUSE's Build Service - I'll add a link to that too from my repository.

`debian/watch` file is used by `uscan` to automatically download latest release versions. This automatically calls `uupdate` - It's mostly a convenience and automation utility. Many times it's enough to do,

    uscan
    cd ../<new_version>
    debuild

and you have your new package ready.

blitzmann:
I will definitely add it to the README file. I've been meaning to add a list of third party packages to it for ages.

Let me know if there is something I can do to make packaging smoother. =)

AdamMajer:
Thank you.

The only thing that you could do to make packaging smoother is not to make radical changes to the package, and I think you are already doing this :smile:

But you could sign release tags or changelog entries as per #262 . Signed archives allow people to verify that source code is authentic and not tampered with by a 3rd party.

altyr:
@AdamMajer In case you're not aware of it, there there are Pyfa packages in the playdeb repository: [http://www.playdeb.net/game/Python%20Fitting%20Assistant](http://www.playdeb.net/game/Python%20Fitting%20Assistant)

(assuming those aren't maintained by you already)

PR #534, opened 2016-03-11
Fixed ewar on damage graphs, 

Wozbo:
And added a quick formula for module falloff and applied it to Target Painters and Heavy Grapplers.

### DIFF
--- a/eos/graph/fitDps.py
+++ b/eos/graph/fitDps.py
@@ -40,11 +40,14 @@ class FitDpsGraph(Graph):
 
         for mod in fit.modules:
             if not mod.isEmpty and mod.state >= State.ACTIVE:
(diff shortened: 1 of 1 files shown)


blitzmann:
I meant to get to this sooner, but forgot about it. Will look into it this weekend. Unfortunately I missed the 1.20.1 release that fixed a cap bug. =(

Wozbo:
Has any progress been made on this?

blitzmann:
I haven't checked on sisi for accuracy because :life:, but I'm not going to worry about it. I'm going to merge it now and as long as it doesn't break due to recent changes it'll be in the release tomorrowish. If it doesn't work right, someone will let us know lol. Besides, the graphing in general needs major work especially now that we have falloff effectiveness for ewar and RR modules.

Issue #522, opened 2016-02-17
Dual prop enabled at the same time on import


geoff-kruss:
Importing a fitting that has both an afterburner and a microwarpdrive propulsion module causes both modules to be activated simultaneously, the usual check that happens when enabling a prop mod doesn't happen on import.

I have only tested this with EFT formatted fittings and only importing via the clipboard, though I would expect that the issue would exist even if importing via some other method. I can reproduce it easily, with varying ship and module combinations. 

In addition, the calculated velocity for the ship appears to take the base velocity, plus the percentage modifier of whichever prop mod is in the higher position, and then applies the next modifier on top of this new total creating really silly looking speeds.

pyfa 1.18.0 - this is where I initially spotted the issue
pyfa 1.19.1 - tested, bug still exists

Only able to test on Windows 10.

Thanks

EDIT: Perhaps a solution would be to pick whichever module is in the highest slot and enable that ? Or leave both disabled... 

EDIT: In issue #512 there is talk of working on validation during import, so this is more than likely related.

blitzmann:
You are correct with your edit, this seems to be a duplicate of #512. Hopefully I can get around to fixing this before the next release in March - it should be really simple.

Issue #512, opened 2016-01-26
Imported fits do not validate if modules fits


ALERTua:
pyfa 1.18.0 allows importing a fitting with 2 Medium Ancillary Armor Repairers fitted, ignoring "Max Modules Of This Group Allowed" = "1".

blitzmann:
Cannot reproduce. Please double check. If it's still the case with you, then provide fitting and screenshot.

Please note that you may be able to achieve this if you import a fit with two AAR's, as I don't believe pyfa checks these things during importing. So if you imported the fit originally, this may be why.

ALERTua:
you are right. this was an imported fit. editing ticket. thank you

blitzmann:
Changed the title to something a little more relevant to the problem. I will look into validating modules when importing.

Also, can you post the fit that you imported? I'm assuming it's EFT formatted...

ALERTua:
[Deimos, Deimos Nahkampf (EVEMon)]

Medium Ancillary Armor Repairer, Nanite Repair Paste
Armor Explosive Hardener II
Damage Control II
Magnetic Field Stabilizer II
Magnetic Field Stabilizer II
Medium Ancillary Armor Repairer, Nanite Repair Paste

Stasis Webifier II
Warp Scrambler II
50MN Microwarpdrive II
Medium Capacitor Booster II, Navy Cap Booster 400

Heavy Neutron Blaster II, Void M
Heavy Neutron Blaster II, Void M
Heavy Neutron Blaster II, Void M
Heavy Neutron Blaster II, Void M
Heavy Neutron Blaster II, Void M

Medium Ancillary Current Router I
Medium Auxiliary Nano Pump I


Hammerhead II x5

blitzmann:
Fixed with #538

PR #538, opened 2016-03-16
Check that modules fit on imported fits #512 #522

aacn500:
On import, modules will respect module quantity limits and hardpoints as well as activation limits.
Also minor bugfix to allow importing of XML loadouts via clipboard.

I only tested these with importing via clipboard, but they should work similarly for other styles. My test loadouts are here: https://gist.github.com/aacn500/e5ec75d4a2048506f4ae

### DIFF
--- a/service/fit.py
+++ b/service/fit.py
@@ -261,6 +261,11 @@ class Fit(object):
                 self.recalc(fit, withBoosters=True)
                 fit.fill()
 
+            # Check that the states of all modules are valid
+            sFit = Fit.getInstance()
+            sFit.recalc(fit)
+            sFit.checkStates(fit, None)
+
             eos.db.commit()
             fit.inited = True
         return fit
--- a/service/port.py
+++ b/service/port.py
@@ -189,11 +189,6 @@ class Port(object):
             except:
                 continue
 
-        # Check that the states of all modules are valid
-        sFit = service.Fit.getInstance()
-        sFit.recalc(f)
-        sFit.checkStates(f, None)
-
         return f
 
     @staticmethod
@@ -237,11 +232,6 @@ class Port(object):
                         m.owner = f
                         if m.isValidState(State.ACTIVE):
                             m.state = State.ACTIVE
-
-        # Check that the states of all modules are valid
-        sFit = service.Fit.getInstance()
-        sFit.recalc(f)
-        sFit.checkStates(f, None)
 
         return f
 
@@ -352,11 +342,6 @@ class Port(object):
             c = Cargo(sMkt.getItem(cargoName))
             c.amount = cargoMap[cargoName]
             fit.cargo.append(c)
-
-        # Check that there are no conflicts between active modules
-        sFit = service.Fit.getInstance()
-        sFit.recalc(fit)
-        sFit.checkStates(fit, None)
 
         return fit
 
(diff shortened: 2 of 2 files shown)


blitzmann:
Hi, 

thanks for the pull request, will look into it tonight probably. But can you elaborate on the XML import bug that you mentioned? I wasn't aware there was an issue with it...

blitzmann:
Also, not sure if we need this bit for everything:

```
               # Check that there are no conflicts between active modules
                sFit = service.Fit.getInstance()
                sFit.recalc(f)
                sFit.checkStates(f, None)
```

If we are making sure the module fits while we add it, then this should always return that the fit was fine, shouldn't it? The problem with this is that if someone imports a lot of fits at one (ie: from a backup), then this is going to recalculate all those fits, causing significant slowdown (untested, I can assume though =P). If there is a reason for it, sure, but I am unsure if this check would ever really do anything (this might be needed for t3 cruisers to check final module count? Unsure, really need tot est this later)

aacn500:
The XML input bug: there was an uncaught exception at https://github.com/pyfa-org/Pyfa/blob/45074952fd3030ede3769c5a5239783ea33378a8/service/port.py#L497 because importXml was being passed `encoding=None` when trying to import from clipboard. I doubt many people try to import xml fits from clipboard, so I suspect that is why it hasn't come up before.

The code block: I used that block of code to stop multiple prop mods being set active in an import, isValidState doesn't check that by itself.

blitzmann:
Nice catch with the bug.

I'm still hesitant with calculating the fit with every import however. I just think there is a smarter way to do it, we just gotta figure it out. Although, with my test database of 585 fits, it took 58 seconds to save vs 20 seconds without calculating them. It's not **too** bad (especially since it's only during large imports), but it's not great. 

We have `Module.canHaveState()` which might be able to help, but two caveats:

* In it's current form, it lets two of the prop mods to pass as active with one disabled, instead of the intended 1 active and 2 disabled. I don't know why this is, but the logic determining this may be funky.
* There's still some edge cases of effects modifying the maxActive attributes (ship hull bonuses to gang links come to mind). I don't believe there's a lot that affect it, but it's still a thing that happens. The only way to know would be to calculate the fit, which is what I'm trying to avoid...

blitzmann:
Actually, we might be able to look into using `checkStates` when the fit is *loaded* to make sure it comes back correctly. In other words, save it to the database with the dirty user-provided states, and check when loading them from the database (maybe through the actual append code such as I did [here](https://github.com/pyfa-org/Pyfa/blob/master/eos/effectHandlerHelpers.py#L134) or when we do a recalc(), or when we load the fit with getFit(), etc). Something to look into.

aacn500:
Moved the `checkStates()` calls into `getFit()` per your suggestion, seems to work nicely. I put the calls inside the `if inited is None or inited is False:` block, because as far as I can tell that is called only the first time a fit is loaded from the database, and after that we can assume that module states have been correctly set. It might be safer running every time a fit is loaded, but I couldn't think of any cases where it would change any states.

blitzmann:
The whole recalculation events are funky and there's no good way to do this right now with what we have. Unfortunately the engine is a bit clunky right now. It would be good if we can do some introspection  into the effect files to find out of a module affects a certain attribute, in which case we could then run that effect alone rather than the entire effect. Would be good for the `maxGroupActive` attribute such as this, but alas, it's not something that's currently supported.

Regardless, I've merged your changes and tweaked a bit - turns out we don't actually have to call recalc where you did as it's already called in the `if not projected` bit. Please continue to test and break for bugs - Thanks for the work!

PR #550, opened 2016-03-29
Update README

MarkNBroadhead:
I got confused when looking for a download link. I think a few more links, where appropriate, will really help, in addition to a few stylistic changes and punctuation corrections.


blitzmann:
Looks good! Thanks for the fixups =)

Issue #556, opened 2016-04-11
Cannot import T3 Cruiser fittings into Pyfa


ppfeufer:
Importing T3 Cruiser Fittings into Pyfa from EVE doesn't really work.

Subsystems and Rigs are fine, High, Med and Low Slots don't get fitted in Pyfa.
![screenie_303](https://cloud.githubusercontent.com/assets/2989985/14440685/63f4c8ac-0031-11e6-8e66-c4d70b5c010c.png)

aacn500:
Importing T3C fits worked in release 1.20.1, so I suspect I broke this with #538. :/ I think this is happening because EFT format lists subsystems at the end of the fit, so when adding high/med/low modules there aren't any slots available yet, so pyfa doesn't fit anything.

blitzmann:
@aacn500 You wanna tackle the fix? I think it would be pretty straightforward - simply parse the EFT text and store the modules and things in a list until the end. Once we have all the info we need, then we can assemble the fit, starting with subsystems.

Or you could possibly tell it to ignore module checking if it's a t3 cruiser. v0v

If you can't get to it, let me know.

blitzmann:
Closing this as it should have been resolved with the pull request.

PR #560, opened 2016-04-15
Fix importing T3 cruisers #556

aacn500:
When looping through items on the fit, subsystems are added immediately but modules are added to a list to be added in a subsequent loop. Tested and seems to work for EFT, XML, DNA and CREST formats, but I'm not certain on EftCfg.


blitzmann:
If EFT import works, EFTcfg should as well.

PR #562, opened 2016-04-17
Added website selection for HTML Export

DuLLSoN:
Recently I have seen some work on HTML Export tool here and wanted to contribute another improvement. I noticed a sizing issue while browsing fits on my laptop. Current online fitting viewer null-sec.com is not playing well with smaller screens so I decided to use another website called [Osmium](https://o.smium.org).

This commit adds an option to select preferred website while exporting fits. null-sec.com is the default option for compatibility but I suggest to swap them until null-sec fixes their page.

python_2016-04-17_17-46-54


blitzmann:
I have been wanting to switch over to O.smium for some time now. Thanks for the work =D

Issue #511, opened 2016-01-25
Damage graph doesn't consider sentry tracking/falloff


mmaucher:
Since a recent patch, there's just a flat line for damage done by sentry drones, independent from distance or target's signature radius or radial velocity.
Please bring back the old damage graph for sentry drones.

blitzmann:
You said since a recent patch its been broken. Do you know when this was caused? I don't think anybody has touched the graphing code for a while

mmaucher:
After checking out 52d19b8 (last commit 2016-01-09), the graphs seem to show the correct DPS.
After checking out 36ab224 (v1.18.0), the graphs are broken.

blitzmann:
Thanks. I'll have to double check, but if true, then that means there was a change in the EVE data regarding sentries, because nothing happened between those commits besides updating data. 

Will look into it when I can. Can you in the meantime provide a quick test fit, along with possibly graphs showing expected plotting vs current plotting?

mmaucher:
As a test fit, a Tristan with a single Garde I should suffice:
[Tristan, AAA]
Garde I x1

I've attached 4 graphs, showing the corresponding dps graphs against a target with and without angular velocity.
Graph 1: Expected output, no angular movement
Graph 2: Expected output, with angular movement
Graph 3: Current output, no angular movement
Graph 4: Current output, with angular movement

![expected_0_degrees](https://cloud.githubusercontent.com/assets/5999595/12607310/31c3fea4-c4d4-11e5-85ee-b3e061775d70.png)
![expected_90_degrees](https://cloud.githubusercontent.com/assets/5999595/12607318/3c6dae90-c4d4-11e5-825f-84d15c6e8ce8.png)
![current_0_degrees](https://cloud.githubusercontent.com/assets/5999595/12607324/4150756e-c4d4-11e5-901f-84f406b59f75.png)
![current_90_degrees](https://cloud.githubusercontent.com/assets/5999595/12607327/446ef50e-c4d4-11e5-8817-f8d1a2191c62.png)

PR #568, opened 2016-04-25
fix separation of moving/non-moving drones

mmaucher:
In a January 2016 patch, the velocity of sentry drones seems to have been increased to non-zero values.
Checking if the velocity surpasses 1 m/s should correctly separate sentries from non-sentries now.

This should fix issue #511.


blitzmann:
Thanks! I totally forgot about that issue <_<

I'll take a look at this during the week =)

Issue #576, opened 2016-05-01
DDAs not stacking penalized for fighters


lunedis:
![image](https://cloud.githubusercontent.com/assets/9085102/14939528/610fc7d0-0f4b-11e6-8a69-494da79e06f8.png)

Proposed fix: 
adding ", stackingPenalties=True" to the 3 lines in the effect

https://github.com/pyfa-org/Pyfa/blob/master/eos/effects/modulebonusdronedamageamplifier.py#L7

blitzmann:
For those interested in fixing this manually until a point release is made, replace contents of `pyfa/library.zip/eos/effects/modulebonusdronedamageamplifier.py` with this:
https://raw.githubusercontent.com/pyfa-org/Pyfa/a1d64acff4524af486df080bb1b427755c3a6670/eos/effects/modulebonusdronedamageamplifier.py

PR #578, opened 2016-05-01
Fixing Drone Damage Amplifier fighter stacking issue #576

lunedis:
Fixing the issue by adding `, stackingPenalties=True` to the end of the lines.


blitzmann:
Thanks!

PR #634, opened 2016-06-05
RAH follows damage profile

Ebag333:
The RAH will use the damage profile applied to adjust the resist on the module.  You can see the adjusted resists in the module attributes.

This does **NOT** take into account the time it takes to adjust (it assumes it adjusts instantly).  It is unlikely there is a clean way to do this that is not needlessly complex, and I'm not sure it'd be that useful anyway for what we're doing.  (Run into the same problem with passive shield HP/s, for example.)

### DIFF
--- a/config.py
+++ b/config.py
@@ -19,7 +19,7 @@ saveInRoot = False
 
 # Version data
 version = "1.21.1"
-tag = "git"
+tag = "Stable"
 expansionName = "Citadel"
 expansionVersion = "1.3"
 evemonMinVersion = "4081"
--- a/gui/builtinViewColumns/misc.py
+++ b/gui/builtinViewColumns/misc.py
@@ -484,6 +484,25 @@ class Miscellanea(ViewColumn):
             text = "{0} / {1}s".format(formatAmount(ehp, 3, 0, 9), formatAmount(duration, 3, 0, 3))
 
             return text, tooltip
+        elif itemGroup == "Armor Resistance Shift Hardener":
+            itemArmorResistanceShiftHardenerEM = (1-stuff.getModifiedItemAttr("armorEmDamageResonance"))*100
+            itemArmorResistanceShiftHardenerTherm = (1-stuff.getModifiedItemAttr("armorThermalDamageResonance")) * 100
+            itemArmorResistanceShiftHardenerKin = (1-stuff.getModifiedItemAttr("armorKineticDamageResonance")) * 100
+            itemArmorResistanceShiftHardenerExp = (1-stuff.getModifiedItemAttr("armorExplosiveDamageResonance")) * 100
+
+            text = "{0}% | {1}% | {2}% | {3}%".format(
+                formatAmount(itemArmorResistanceShiftHardenerEM, 3, 0, 3),
+                formatAmount(itemArmorResistanceShiftHardenerTherm, 3, 0, 3),
+                formatAmount(itemArmorResistanceShiftHardenerKin, 3, 0, 3),
+                formatAmount(itemArmorResistanceShiftHardenerExp, 3, 0, 3),
+            )
+            tooltip = "Resistances Shifted to Damage Profile:\n{0}% EM | {1}% Therm | {2}% Kin | {3}% Exp".format(
+                formatAmount(itemArmorResistanceShiftHardenerEM, 3, 0, 3),
+                formatAmount(itemArmorResistanceShiftHardenerTherm, 3, 0, 3),
+                formatAmount(itemArmorResistanceShiftHardenerKin, 3, 0, 3),
+                formatAmount(itemArmorResistanceShiftHardenerExp, 3, 0, 3),
+            )
+            return text, tooltip
         elif stuff.charge is not None:
             chargeGroup = stuff.charge.group.name
             if chargeGroup in ("Rocket", "Advanced Rocket", "Light Missile", "Advanced Light Missile", "FoF Light Missile",


Ebag333:
An example of a Sabre with a RAH active and Blood Raider (deadspace) resist profile applied:

[http://puu.sh/phk8A/41078a9330.png](url)

blitzmann:
On the surface it looks good. Was this tested on SISI / TQ for accuracy? I am probably not one to test it for accuracy (at least as accurate as it can be, considering it's not taking into consideration time to adjust) as I've never been a fitting guru, it would help if others could chime in with comments on the effects of this.

For those that wish to test on current stable version:

Take [this file](https://raw.githubusercontent.com/Ebag333/Pyfa/8f3370307afa1bf87412ff37d3868055ec5f78c6/eos/effects/adaptivearmorhardener.py) and save it to `pyfa/library.zip/eos/effects/adaptivearmorhardener.py` (overwrite). Then test the RAH module against various damage profiles.

@Ebag333, if you could add a misc column value for the RAH that shows the current resists it has calculated. So, for the example picture, it would look like this: `32.6/27.3/0/0` and have the tooltip be a little more verbose (eg: `32.6 EM / 27.3 THERM / etc`). I think this would be a good way to show folks what the values are currently without having to go into the item stats.

Ebag333:
I did some quick testing but will test further on TQ.  Smoke testing it looks good, so far I haven't seen any numbers that look wrong.

@blitzmann: So there's a couple places to look at.

In the item attributes, I think this is a separate ask from this particular enhancement, but it'd be nice to have a "base value" column.  So in the case of the RAH, "Value" might be 38.5%, but "Base Value" would always be 15%.  This would be a change globally for all items (many would be the same for both columns), but would give a clearer picture for how much your skills/fit affects the item (something I know I've desired more than once).

For the tooltip you speak of, there's already a tooltip over the resists.  It's not a particularly useful tooltip, to be honest.  I also had  a good discussion recently on #fits in tweetfleet slack about how the resists are a bit deceptive, since it's really a reduction rather than an add.  Thanks CCP for making that confusing.  So a tooltip showing the .85 instead of 15% would be nice. I'd like to see improvements to it, but that seems more along the lines of a full revamp of the tooltips for that particular section than RAH specific.

Playing with the GUI/tooltips is a bit beyond my ken at the moment, but I could create some new feature requests for the two items above (if there isn't already).  Seems like these changes are more global to this specific pull though.

If I'm totally missing what you were getting at, let me know.

blitzmann:
Having a "Base Value" and "Modified Value" column is something that has been brought up in the past. I don't remember why I never tackled it, but I think it's a little more difficult than was originally thought (or maybe not difficult, just a bigger change. original, unmodified values of attributes are already stored for safe keeping which could be used for this feature). But yeah, this is a different feature request all-together.

What I was talking about is the Misc column in the fitting window. See image.

capture

We use this column to display custom info based on the item group. See:

https://github.com/pyfa-org/Pyfa/blob/master/gui/builtinViewColumns/misc.py

What I was saying is a simple addition to account for RAHs that shows the current resist bonuses it's giving in a text representation in that column. ;)

Ebag333:
Mm.  I see.

That's a good idea, my only concern is pushing everything over as it could be quite large.  Worst case scenario would be something like 7 characters per resist (such as 38.243%), across 4 resists, so that'd end up being 31 characters wide.

Is the information that critical, since it's reflected on the resists window and in show info?  I just don't know if most people would care about how much their RAH is modifying it, vs the total resist.

blitzmann:
It's not super critical, no. But that's generally what the misc column is for - supplemental info that the user may be interested in. They may want to know what percentage of their resists is coming from the RAH against a certain damage profile. Length is a concern, but we can restrict the number of significant digits we return pretty easy.

Ebag333:
Can you point me at a good existing example of this being set in an effects file?

Hmm. If we run this in the effects file I'll have to do a case/switch (yes I know Python doesn't actually have this).  I was hoping to avoid that and keep it dynamic, but the resists don't run "in order".  It starts with Kin, not EM.

Ebag333:
Tested on TQ, and there is a small issue, but not one we can really solve easily.

RAH in practice doesn't do genuinely dynamic movements between resists.  So in the case of adjusting to 2 incoming damage types, it doesn't split them as they *SHOULD* be split. It will always keep 2 or 3 resists even.  So in the case of getting shot at by say Antimatter (which is 41.7% Therm 58.3% Kin), it should follow that split.  Instead you end up with something like this:
EM/Therm/Kin/Expl
15/15/15/15
12/18/18/12
09/21/21/09
07/23/23/07
03/27/27/03
00/30/30/00

So even though you'd think you'd see 0/24/36/0, you actually see 0/30/30/0.

In practice I don't think this is a big enough difference to matter, especially since often users will be checking against a single resist.

Some examples using a Vengeance with a RAH fitted:

Against Antimatter (Therm/Kin)
**Pyfa**:
77.6/81.2/83.9/86.8
**In game**:
79/84/84/88

Against Nova missiles (Exp)
**Pyfa**:
77.6/75/75.2/94.7
**In game**:
79/77/77/956

I'm not sure where the "free" few % points in EM/Therm/Kin show up from.  Pyfa seems to be slightly low on a number of fits, I suspect it's mostly just bad math on the built in fitting window but hard to say for sure.  Either way, 2% isn't enough to matter.

blitzmann:
Not the effects file, this file:

https://github.com/pyfa-org/Pyfa/blob/master/gui/builtinViewColumns/misc.py

Here's a quick implementation of what I'm talking about. Only tested to know it works, havent tested precisions or anything to see if it works well or anything. Feel free to tweak it.

```

        elif itemGroup == "Armor Resistance Shift Hardener":
            em = (1-stuff.getModifiedItemAttr("armorEmDamageResonance"))*100
            therm = (1-stuff.getModifiedItemAttr("armorThermalDamageResonance")) * 100
            kin = (1-stuff.getModifiedItemAttr("armorKineticDamageResonance")) * 100
            exp = (1-stuff.getModifiedItemAttr("armorExplosiveDamageResonance")) * 100

            text = "{0}/{1}/{2}/{3}".format(
                formatAmount(em, 3, 0, 3),
                formatAmount(therm, 3, 0, 3),
                formatAmount(kin, 3, 0, 3),
                formatAmount(exp, 3, 0, 3),
            )
            tooltip = "Modified resistances:\n{0}% EM / {1}% THERM / {2}% KIN / {3}% EXP".format(
                formatAmount(em, 3, 0, 3),
                formatAmount(therm, 3, 0, 3),
                formatAmount(kin, 3, 0, 3),
                formatAmount(exp, 3, 0, 3),
            )
            return text, tooltip
```

capture

Ebag333:
@blitzmann added the misc column stuff, if you want to take a look at it.

PR #643, opened 2016-06-11
Added sensor strength to handle scripting the SeBo

Ebag333:


Ebag333:
So this fixes the issue with scripting it....buuuuut......

Now it shows the ECM strength, because CCP decided to not make it the same.  So now it'll show 48% instead of 30%.

It'd be nice to do what we did with the RAH (similar to missiles, etc).  This was just a quick fix as long as I was in there doing the RAH stuff.

Ebag333:
It was bugging me so rewrote the function.

http://puu.sh/poAX2/7b189ea3bb.png

Basically the same style function as RAH, missiles, etc. Much more clear to me, and based on feedback I've gotten, more clear to others.

blitzmann:
Thanks!

gui/builtinViewColumns/misc.py
```diff
@@ -231,21 +231,20 @@ def __getData(self, stuff):
...
-            if display == scanResBonus:
-                ttEntries.append("scan resolution")
-            tooltip = "{0} bonus".format(formatList(ttEntries)).capitalize()
+
+            text = "{0}% | {1}% | {2}%".format(
+                formatAmount(scanResBonus, 3, 0, 3),
+                formatAmount(lockRangeBonus, 3, 0, 3),
+                formatAmount(gravBonus, 3, 0, 3),
+            )
+            tooltip = "Applied bonuses:\n{0}% scan resolution| {1}% lock range | {2}% sensor strength".format(
```
Ebag333: Missing space at: scan resolution|

:(

blitzmann: I already fixed it in master =)

PR #644, opened 2016-06-11
Added base column to item window

Ebag333:
Added a new column for the base (original) value.  Renamed column from "value" to "current value" to hopefully avoid confusion.

Also made the window a bit larger, since it's a bit small to fit everything.

Example of an item and a ship:

http://puu.sh/powiW/8837cc074e.png

### DIFF
--- a/gui/itemStats.py
+++ b/gui/itemStats.py
@@ -246,7 +246,7 @@ class ItemDescription ( wx.Panel ):
         if not item.description:
             return
 
-        desc = item.description.replace("\r", "<br>")
+        desc = item.description.replace("\n", "<br>")
         # Strip font tags
         desc = re.sub("<( *)font( *)color( *)=(.*?)>(?P<inside>.*?)<( *)/( *)font( *)>", "\g<inside>", desc)
         # Strip URLs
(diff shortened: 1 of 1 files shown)


gui/itemStats.py
```diff
@@ -246,7 +246,7 @@ def __init__(self, parent, stuff, item):
         if not item.description:
             return
 
-        desc = item.description.replace("\n", "<br>")
+        desc = item.description.replace("\r", "<br>")
```
blitzmann: Revert this please. See https://github.com/pyfa-org/Pyfa/commit/13a15a38c54b96de80788e44146a26119683375e

blitzmann:
I commented on a line that needs to be reverted, unless you specifically changed it for a reason, in which case let me know. =)

Something else that you might want to look at is only having the base value column show up when the user looks at a module which is not fitted. That way we don't have two columns with the same value. IIRC, the `stuff` parameter that is passed into the class is `None` if you're looking at a simple market item

Ebag333:
Updated to only show one column if the stuff parameter is empty.

PR #649, opened 2016-06-12
Updated ECM to be more descriptive

Ebag333:
Gave it the RAH treatment. Now shows min/max values if it's a racial (since they have a strong/weak value), and only one for multi/burst (since it's omni).  Tooltip shows all four broken out.

Tried showing all 4 in the misc window, it gets kinda cluttered.  Showing min/max with all four broken out in the tooltip seems to be the best compromise.


Ebag333:
Example of the selected racial ECM:
http://puu.sh/pqyGm/5aba0dd6b6.png

blitzmann:
Capitalize the sensor types and it's good to go =)

Ebag333:
Done and done.

blitzmann:
Thanks!

PR #650, opened 2016-06-13
Grouped Attributes override

Ebag333:
Users are confused because they find the menu to edit attribute
overrides, change values, then do not see their values reflected.  There
is an option that needs to be enabled under the edit menu to show the
users overrides.

Moved this option from the edit menu to the window menu (next to the
existing attributes override editor), and grouped them together.
Functionality is not changed, merely placement of the overrides toggle.


blitzmann:
Fuck semantics, it works better this way.

PR #669, opened 2016-07-02
Fixed type of weapon

Ebag333:
Used to be applied to Naglfar, but everyone knows that Vehement is the
new Naglfar, and never use projectiles after memorial day (only
hybrids).


Ebag333:
I also tweaked the effect file to better match the other effect files. Because looking at them all to compare and seeing _src_ vs _ship_ was making my OCD twitch...

blitzmann:
Thanks!

PR #683, opened 2016-07-15
Database and Logging Preference Windows

Ebag333:
So far I've added the two panes.  The values seem to work, and will carry through a session, but there's not really anywhere to save them to to keep them long term.

Most of these seem to come from the command line switches (which are probably not the easiest way to handle them for end users).

Shouldn't be hard to tweak it to be handled like the rest of the preferences.  Question is, do we want to continue to keep support for command line switches, or can we ditch that functionality?

Todo:
*Figure out how to save these long term
*Make the database import checkbox work

### DIFF
--- a/gui/builtinPreferenceViews/__init__.py
+++ b/gui/builtinPreferenceViews/__init__.py
@@ -1,4 +1,4 @@
-__all__ = ["pyfaGeneralPreferences","pyfaHTMLExportPreferences","pyfaUpdatePreferences","pyfaNetworkPreferences"]
+__all__ = ["pyfaGeneralPreferences","pyfaHTMLExportPreferences","pyfaUpdatePreferences","pyfaNetworkPreferences","pyfaDatabasePreferences","pyfaLoggingPreferences"]
 
 import wx
 
--- a/gui/builtinPreferenceViews/pyfaDatabasePreferences.py
+++ b/gui/builtinPreferenceViews/pyfaDatabasePreferences.py
@@ -1,0 +1,110 @@
+import wx
+
+from gui.preferenceView import PreferenceView
+from gui.bitmapLoader import BitmapLoader
+
+import gui.mainFrame
+import service
+import config
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+from eos.db.saveddata.loadDefaultDatabaseValues import DefaultDatabaseValues
+
+
+class PFGeneralPref ( PreferenceView):
+    title = "Database"
+
+    def populatePanel( self, panel ):
+        self.mainFrame = gui.mainFrame.MainFrame.getInstance()
+        self.dirtySettings = False
+        #self.openFitsSettings = service.SettingsProvider.getInstance().getSettings("pyfaPrevOpenFits", {"enabled": False, "pyfaOpenFits": []})
+
+        mainSizer = wx.BoxSizer( wx.VERTICAL )
+
+        self.stTitle = wx.StaticText( panel, wx.ID_ANY, self.title, wx.DefaultPosition, wx.DefaultSize, 0 )
+        self.stTitle.Wrap( -1 )
+        self.stTitle.SetFont( wx.Font( 12, 70, 90, 90, False, wx.EmptyString ) )
+        mainSizer.Add( self.stTitle, 0, wx.ALL, 5 )
+
+        self.stSubTitle = wx.StaticText( panel, wx.ID_ANY, u"(Changes Require Restart)", wx.DefaultPosition, wx.DefaultSize, 0 )
+        self.stSubTitle.Wrap( -1 )
+        self.stSubTitle.SetFont( wx.Font( 10, 70, 90, 90, False, wx.EmptyString ) )
+        mainSizer.Add( self.stSubTitle, 0, wx.ALL, 5 )
+
+        self.m_staticline1 = wx.StaticLine( panel, wx.ID_ANY, wx.DefaultPosition, wx.DefaultSize, wx.LI_HORIZONTAL )
+        mainSizer.Add( self.m_staticline1, 0, wx.EXPAND|wx.TOP|wx.BOTTOM, 5 )
+
+        #Save in Root
+        self.cbsaveInRoot = wx.CheckBox( panel, wx.ID_ANY, u"Use Executable Path", wx.DefaultPosition, wx.DefaultSize, 0 )
+        mainSizer.Add( self.cbsaveInRoot, 0, wx.ALL|wx.EXPAND, 5 )
+
+        #Database path
+        self.stSetFitDBPath = wx.StaticText( panel, wx.ID_ANY, u"Saved Fit Database Path:", wx.DefaultPosition, wx.DefaultSize, 0 )
+        self.stSetFitDBPath.Wrap( -1 )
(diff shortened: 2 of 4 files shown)


Ebag333:
@blitzmann wanted to get your feedback on thoughts for storing these long term and moving it out of the command line.

From config.py
```
    global debug
    global pyfaPath
    global savePath
    global saveDB
    global gameDB
    global saveInRoot
```

These are the variables we're dealing with (well, except pyfaPath).

Mostly I want to expose these to more easily allow me to run two installs of Pyfa without using the same DB.  My DB is heavily corrupted, and I suspect that part of that is caused by having it open in two different installs/versions (to more easily compare features as I've been building stuff).

blitzmann:
Can't pull / test this atm, but this is basically a database / user settings path preference pane, correct? 

I've thought about this before, and have wanted to do something like this. However, there's the problem with how to handle user settings. As you've mentioned, there's an issue with storing these long term. The `savePath` determines where all user data should be saved. However, if you change it through preferences, then that path will be saved in the preferences which is stored at wherever it was stored before. You would have to check permissions for the new locations beforehand, and spin up a new user / settings directory. After that, moving the user database, which in itself can be problematic... I wonder if you can have sqlalchemy switch the database path on the fly. All of this should be totally possible, but as I'm adverse to fucking with user data unless needed, I opted to just sit on these types of things.

You note a use-case for developers who have broken vs pristine database. This is the thought behind the `saveInRoot` option (`-s`). Use this, and all info is self-contained with that specific pyfa directory. There is also the `-s` flag, which allows you to specify your own save path. But again, we run into the problem of having save in root hidden as a switch. There was an idea of possible automatically adding this switch when installed via the .exe installer. That would be an option in InnoScript, and I haven't had the time to look into it.

I do like the idea of users having a GUI interface to turn debugging on. 

I've also had an idea of allowing the user to supply a switch that will pop up a list of user databases currently located in the save path directory. Once the user selects one, pyfa uses that one for it's instance. I have a bunch for testing, and in order to use them it's a game of renaming them to what pyfa expects. This would allow me to just copy another one in there and use the switch to determine which one I want to work off of. Could also potentially use it as a way to generate a new database if needed. 

In summary:
* I think debug option absolutely should be exposed
* I think save in root should be exposed as well, but this is probably better suited as an install option in InnoScript. Then again, it automatically defaults to ProgramFiles, and I'm not sure if permissions will allow save in root to work in that. I've never actually tried.
* I am not sure about the idea of having the user pick their own save path. The fact that these setting would be saved in the settings, which is saved in the user path, which is what we're trying to change, could cause a lot of annoyances. I introduce the alternative idea of a pop up when starting pyfa to choose their user database (user can toggle whether or not this pop up happens in settings).
* As for the eve database, this is supposed to be pristine. I don't think we should go through the effort of implementing a way to change this, as only developers would have need for it, and even then on a limited basis. User's aren't going to want an old database, and if they do they would be better served using an older version of pyfa that includes it. For developers, they can easily change the database and commit to the branch that they are working on to always have a database that reflects that branch.

Thoughts?

Ebag333:
>Can't pull / test this atm, but this is basically a database / user settings path preference pane, correct?

Correct.

>However, if you change it through preferences, then that path will be saved in the preferences which is stored at wherever it was stored before. You would have to check permissions for the new locations beforehand, and spin up a new user / settings directory. After that, moving the user database, which in itself can be problematic... I wonder if you can have sqlalchemy switch the database path on the fly. All of this should be totally possible, but as I'm adverse to fucking with user data unless needed, I opted to just sit on these types of things.

All valid points.

My thought was simply to save the setting, and the user would have to restart Pyfa for the new setting to take effect.  Moving the DB wouldn't be for your average user, so I don't really have a problem with making them restart Pyfa.  Also, not moving/migrating the DB at all, but simply treating it as a new setup (if they forget to move theirs manually).

>I think save in root should be exposed as well, but this is probably better suited as an install option in InnoScript. Then again, it automatically defaults to ProgramFiles, and I'm not sure if permissions will allow save in root to work in that. I've never actually tried.

For Windows folks, it's bad practice to store user data in Program Files.  It requires admin rights, you'll get UAC popups, etc.  That being said, if a user wants to override the default, I don't see a reason to stop them.  There are legitimate use cases, like for example I install Pyfa onto my Google Drive, which means the latest version follows me no matter what computer I log onto.  But currently, my data does not.  (Developers with multiple installs is another option).

>I am not sure about the idea of having the user pick their own save path. The fact that these setting would be saved in the settings, which is saved in the user path, which is what we're trying to change, could cause a lot of annoyances. I introduce the alternative idea of a pop up when starting pyfa to choose their user database (user can toggle whether or not this pop up happens in settings).

I love the option to choose a user database.  Even if it was a button in the preferences, it'd be better than jumping through hoops with the whole renaming dance.

I would still say to at least _expose_ it. Right now people don't know where the databases are stored, so simply having the path in an easily place to copy/paste would help users when  you ask them to send them their DB.

>As for the eve database, this is supposed to be pristine. I don't think we should go through the effort of implementing a way to change this, as only developers would have need for it, and even then on a limited basis. User's aren't going to want an old database, and if they do they would be better served using an older version of pyfa that includes it. For developers, they can easily change the database and commit to the branch that they are working on to always have a database that reflects that branch.

Pretty much the same as above.  If you write the code for the popup choice for the user DB, it's trivial to do the same for the EVE DB.  The valid non-dev use case would be switching between SiSi and TQ databases.  We could also more easily beta test updates this way.

But again, exposing the path so that people can find it is a good thing, in my mind.

blitzmann:
> I would still say to at least expose it. Right now people don't know where the databases are stored, so simply having the path in an easily place to copy/paste would help users when you ask them to send them their DB.

Ah, yeah, I forgot to mention: Having a menu option for something like "Open User Data Directory", or maybe put it somewhere in preferences, has been something I've been wanting for a while, just never got around to doing it. So this would expose where their files are located. 

> For Windows folks, it's bad practice to store user data in Program Files. It requires admin rights, you'll get UAC popups, etc. That being said, if a user wants to override the default, I don't see a reason to stop them. There are legitimate use cases, like for example I install Pyfa onto my Google Drive, which means the latest version follows me no matter what computer I log onto. But currently, my data does not. (Developers with multiple installs is another option).

The save in root option is for things like you stated: installing on Google Drive, Sync, Dropbox, flash drive. If we were to go the route of using inno setup to utilize this option (my preference), then we should make sure that the option is only available when the user is trying to install to a user-owned directory. 

Again, I don't know about letting the user change the save path, simply because that brings about complications with settings (and if the user sets it to Program Files and then can't save... ugh, I don't want those issues coming in). BUT, I think giving them a button or menu item the opens their save path is a good compromise, as well as developing a simple GUI window that pop open allowing user to select their user database before SQLAlchemy is initialized. That should cover most use cases where different save paths might be wanted.

As for the eve database... I dunno. Yeah, it would help with switching between tq/sisi data, but then again we don't release just the database for these, we usually issue a pre-release. It is out of the users scope to compile a sisi database for themselves, so this would really only benefit developers and extreme power users. If we get the DB switching working for user databases, I don't really see a reason why we wouldn't let these be selectable as well.

blitzmann:
Also, I think we should keep the current command switches for the sake of compatibility if possible. I don't want peoples set ups to break because the new release removes this functionality.

Ebag333:
This has been folded over into #868 .

No need to keep this open any longer.

blitzmann:
yay, down to 14 PRs

PR #684, opened 2016-07-17
Applied size reduction effect to neut/nos

Ebag333:
Relevant code is in the effects file. Basically grabs the ships rig size, and applies the small/medium/large modifiers as a straight up percentage.

Example, ship on the left with the legacy code, ship on the right with the new code.

http://puu.sh/q4qsc/83975220d9.png

Todo:
*Could figure out how CCP handles unrigged ships (freighters, shuttles, pods).  This is an edge case, so we may simply choose not to cover it.
*Not sure how CCP determines the actual size, they may use a different formula than rig size.
*Need to handle the sig radius reduction for capital sized neuts. Still not sure on formula.
*Need to verify that stacking behavior works with cap batteries.  Assumption here is that the module size reduction applies without a stacking penalty and prior to the cap battery reduction.

### DIFF
--- a/eos/effects/energydestabilizationnew.py
+++ b/eos/effects/energydestabilizationnew.py
@@ -1,10 +1,9 @@
 # Not used by any item
 from eos.types import State
 type = "active", "projected"
-def handler(fit, container, context):
-    if "projected" in context and ((hasattr(container, "state") \
-    and container.state >= State.ACTIVE) or hasattr(container, "amountActive")):
-        multiplier = container.amountActive if hasattr(container, "amountActive") else 1
-        amount = container.getModifiedItemAttr("energyNeutralizerAmount")
-        time = container.getModifiedItemAttr("duration")
-        fit.addDrain(time, amount * multiplier, 0)
+def handler(fit, src, context):
+    if "projected" in context and ((hasattr(src, "state") and src.state >= State.ACTIVE) or hasattr(src, "amountActive")):
+        multiplier = src.amountActive if hasattr(src, "amountActive") else 1
+        amount = src.getModifiedItemAttr("energyNeutralizerAmount")
+        time = src.getModifiedItemAttr("duration")
+        fit.addDrain(src, time, amount * multiplier, 0)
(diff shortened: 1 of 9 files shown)


Ebag333:
>Need to handle the sig radius reduction for capital sized neuts.

This one is tricky, because I know from testing that a capital neut basically does nothing to a frigate.  But if we only apply the size reduction, it still instacaps the frigate (according to Pyfa).

@blitzmann if you want to merge this PR, we can always open a new one specific to the carrier neuts.

Ebag333:
From Reddit:
>Signature Radius works as per the bomb damage formula.
https://www.reddit.com/r/evecapitalfocusgroup/comments/489zj7/feedback_wanted_capital_neuts_nos/

And now looking up the bomb sig radius formula, turns up this:
```
Damage = Base_Damage * MIN(sig/Er,1)
Where
sig = ship's signature
Er = Explosion Radius of bomb
```
https://oldforums.eveonline.com/?a=topic&threadID=1505798

No idea if this is the correct formula or not but it looks legit.

Ebag333:
The relevant bit of code for the sig reduction:
```
        energyNeutralizerSignatureResolution = module.getModifiedItemAttr("energyNeutralizerSignatureResolution")
        signatureRadius = fit.ship.getModifiedItemAttr("signatureRadius")

        #Signature reduction, uses the bomb formula as per CCP Larrikin
        if energyNeutralizerSignatureResolution:
            sigRatio = signatureRadius/energyNeutralizerSignatureResolution

            sigReductionList = [1, sigRatio]
            amount = amount*min(sigReductionList)
```

So a couple things on this.

1) I'm not sure that the is the actual formula used. (See post above)
2) I'm not sure how this stacks.  Testing in game seems to show it does both (for example there is virtually no effect against frigates )

blitzmann:
This all looks pretty good to me. I will wait until I can sit down and test things myself on SISI for expected values before merging. Great work @Ebag333!

Ebag333:
Confirmed by `sources`

Issue #694, opened 2016-08-09
Overrides window does not work


lunedis:
I open the overrides window, type into the search bar, press [Enter], press the search icon, but nothing happens.

I don't have a git version I can start to see the log atm.

Ebag333:
It's happening in Master.

So far I've traced it down to where `scheduleSearch` hands off the search to `sMkt.searchItems`, but it never seems to come back.

The market service has changed quite a bit for fighters and a bunch of other things.

It was working though when I submitted the change to move the override toggle on the menu....

Ebag333:
Okay, it's the filter being off that's the problem. Turn on the filter (default), and it works.

Hmm, wonder if it's something in the database....

blitzmann:
Fixed 5a5ab84

Issue #673, opened 2016-07-05
Show market prices on the compare window


Ebag333:
As an enhancement, would be nice to show the market prices for the items on the item compare tab. (As a new column on far right.)

This would let you quickly see which ones are even worth looking at, instead of having to load each one onto the fit individually.

blitzmann:
![capture](https://cloud.githubusercontent.com/assets/3904767/16640785/04e578d0-43c9-11e6-9039-b6f0973bcd58.PNG)

PR #697, opened 2016-08-11
Attribute Editor Broken

Ebag333:
When the filter is disabled, the search fails to return any values.

See #694


Ebag333:
This is a quick and dirty fix and not a final one. But merging this would at least allow it to work, or we can keep open until we figure out the root cause.

Ebag333:
Okay, the problem lies with #673.

At a _guess_?  There's something in the database we can't get a price for (there's actually a lot of somethings).  So since it's trying to get a price, it bombs and returns nothing.

Having it filter works because all the market items are valid to get prices for.

blitzmann:
Fix already pushed :)

PR #700, opened 2016-08-19
fixed exception that froze interface

PageArkanis:
```
  File ".../Pyfa/gui/builtinViews/fittingView.py", line 463, in fitChanged
    self.refresh(self.mods)
  File ".../Pyfa/gui/builtinViews/fittingView.py", line 590, in refresh
    self.SetItemBackgroundColour(i, self.slotColour(mod.slot))
  File ".../Pyfa/gui/builtinViews/fittingView.py", line 560, in slotColour
    return self.slotColourMap[slot] or self.GetBackgroundColour()
KeyError: 8
```

### DIFF
--- a/gui/builtinViews/fittingView.py
+++ b/gui/builtinViews/fittingView.py
@@ -557,7 +557,7 @@ class FittingView(d.Display):
                      5: ''}
 
     def slotColour(self, slot):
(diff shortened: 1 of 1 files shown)


blitzmann:
Nice catch, thanks!

PR #701, opened 2016-08-19
Update README.md with current brew instructions

shagie:
```
Error: No available formula with the name "pyfa" 
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
This formula was found in a tap:
Caskroom/cask/pyfa
To install it, run:
  brew install Caskroom/cask/pyfa
```


blitzmann:
Thanks!

PR #702, opened 2016-08-23
More properly behaviour for fitting implant sets

archonoff:
Hi
In my opinion it's more reasonable behaviour for this function ("add implant set")

Old behaviour:
It actually ADDS implants from set into current fit. If there were other implants in fit and also empty slots in set, the old implants have been remained in fit. But I use this feature e.g. for jump clones and I need change all previously fitted implants.

New behaviour:
All fitted implants are removed from fit before adding new from set.

### DIFF
--- a/gui/builtinContextMenus/implantSets.py
+++ b/gui/builtinContextMenus/implantSets.py
@@ -69,6 +69,11 @@ class ImplantSets(ContextMenu):
         else:
             sFit = service.Fit.getInstance()
             fitID = self.mainFrame.getActiveFit()
+
+            # at first remove all implants
+            sFit.removeAllImplants(fitID)
+
+            # and add new implants from set
             for implant in set.implants:
                 sFit.addImplant(fitID, implant.item.ID, recalc=implant == set.implants[-1])
 
--- a/gui/builtinContextMenus/implantSets.py
+++ b/gui/builtinContextMenus/implantSets.py
@@ -12,7 +12,7 @@ class ImplantSets(ContextMenu):
         return srcContext in ("implantView", "implantEditor")
 
     def getText(self, itmContext, selection):
-        return "Add Implant Set"
+        return "Fit Implant Set"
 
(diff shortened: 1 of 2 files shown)


Kyria:
Can't we just add a "prompt popup" asking whether we want to just fill empty slots or replace all slots ? 
We "need" your behavior, but i think we also need the old one, as not everyone might want to replace all slots each time, for example the omega slot, since it's often used for fitting implants, or if they are using 2 sets or any other good|bad fitting reason :)

archonoff:
@Kyria 
> not everyone might want to replace all slots each time, for example the omega slot
For this purposes I just create two implants sets with only one different implant. I need make this action only one time. And further I can change implant sets in two clicks.
And I deffinitly don't wont see a "promt popup" every time I change implant set.

blitzmann:
I dunno, I'm also in favor of a prompt, or maybe a separate submenu with two options: "Replace All" and "Add / Overwrite" (or however it should be worded), only shown if fit currently has it's own fittings. Food for thought. 

I haven't had the time to take a look at this PR, but it seems pretty straightforward. :)

Ebag333:
>But I use this feature e.g. for jump clones and I need change all previously fitted implants.

Why wouldn't you use chars for JC like functionality?  You can duplicate chars quickly/easily (even keeps the API stuff), and then easily add implants to that.

I think the bigger issue is that there's two different ways to accomplish the same thing, and there's a _lot_ of overlap between them.  I think a more unified solution would be best, but then you're getting into a much bigger rewrite.

I think a better solution all around would be to allow for adhoc implants (like today with the dragging and dropping), move the implant sets to replace the char implants, and make those more fixed sets of implants.  Basically ditch the char implants entirely, and make implant sets more fixed.

DarkFenX:
With ability to multi-select implants i do not think it is an issue. Ctrl-A, delete, add implant set. I think it's better than introducing additional behavior controls.

Issue #726, opened 2016-09-06
Feature Request:  Export fit as a multi-buy shopping list 


mkemnetz:
It would be awesome if instead of copying a fit to the clipboard as an eft block, you could also export the fit as a simple text shopping list to be pasted into the multibuy window.Basically you would go from this:

```
[Bhaalgorn, EFT Block]

1600mm Steel Plates II
1600mm Steel Plates II
Damage Control II
Imperial Navy Energized Adaptive Nano Membrane
Corpum C-Type Energized Explosive Membrane
Corpum B-Type Energized Kinetic Membrane
Corelum B-Type Energized Thermal Membrane

Sensor Booster II, ECCM Script
Sensor Booster II, ECCM Script
100MN Afterburner II
Ta3 Compact Ship Scanner
Large Cap Battery II

Dark Blood Heavy Energy Nosferatu
Dark Blood Heavy Energy Nosferatu
Heavy Energy Neutralizer II
Heavy Energy Neutralizer II
Heavy Energy Neutralizer II
Heavy Energy Neutralizer II
Heavy Energy Neutralizer II

Large Anti-EM Pump I
Large Trimark Armor Pump I
Large Trimark Armor Pump I


Valkyrie II x5
Vespa EC-600 x5
Infiltrator II x5
```

To this:

```
Bhaalgorn
1600mm Steel Plates II
1600mm Steel Plates II
Damage Control II
Imperial Navy Energized Adaptive Nano Membrane
Corpum C-Type Energized Explosive Membrane
Corpum B-Type Energized Kinetic Membrane
Corelum B-Type Energized Thermal Membrane
Sensor Booster II
ECCM Script
Sensor Booster II
ECCM Script
100MN Afterburner II
Ta3 Compact Ship Scanner
Large Cap Battery II
Dark Blood Heavy Energy Nosferatu
Dark Blood Heavy Energy Nosferatu
Heavy Energy Neutralizer II
Heavy Energy Neutralizer II
Heavy Energy Neutralizer II
Heavy Energy Neutralizer II
Heavy Energy Neutralizer II
Large Anti-EM Pump I
Large Trimark Armor Pump I
Large Trimark Armor Pump I
Valkyrie II x5
Vespa EC-600 x5
Infiltrator II x5
```

To go from an EFT block to a shopping lift the brackets and name of the fit need to be removed.  Then ammo in modules need to be handled as multibuy doesn't understand commas.  This would make buying a fit with cargo a lot easier.

PR #733, opened 2016-09-14
Multibuy format export feature, adresses #726

petosorus:
Hey, it's my first time working on Pyfa but I think I got what needed to be done for this.
I compared with the example in #726 and it matches.

Not sure about where to put the Multibuy option in the list so I just put it at the bottom, written MultiBuy (not sure about typography either).

### DIFF
--- a/gui/copySelectDialog.py
+++ b/gui/copySelectDialog.py
@@ -26,17 +26,19 @@ class CopySelectDialog(wx.Dialog):
     copyFormatXml = 2
     copyFormatDna = 3
     copyFormatCrest = 4
+    copyFormatMultiBuy = 5
 
     def __init__(self, parent):
         wx.Dialog.__init__(self, parent, id = wx.ID_ANY, title = u"Select a format", size = (-1,-1), style = wx.DEFAULT_DIALOG_STYLE)
         mainSizer = wx.BoxSizer(wx.VERTICAL)
 
-        copyFormats = [u"EFT", u"EFT (Implants)", u"XML", u"DNA", u"CREST"]
+        copyFormats = [u"EFT", u"EFT (Implants)", u"XML", u"DNA", u"CREST", u"MultiBuy"]
         copyFormatTooltips = {CopySelectDialog.copyFormatEft: u"EFT text format",
                               CopySelectDialog.copyFormatEftImps: u"EFT text format",
                               CopySelectDialog.copyFormatXml: u"EVE native XML format",
                               CopySelectDialog.copyFormatDna: u"A one-line text format",
-                              CopySelectDialog.copyFormatCrest: u"A JSON format used for EVE CREST"}
+                              CopySelectDialog.copyFormatCrest: u"A JSON format used for EVE CREST",
+                              CopySelectDialog.copyFormatMultiBuy: u"MultiBuy text format"}
         selector = wx.RadioBox(self, wx.ID_ANY, label = u"Copy to the clipboard using:", choices = copyFormats, style = wx.RA_SPECIFY_ROWS)
         selector.Bind(wx.EVT_RADIOBOX, self.Selected)
         for format, tooltip in copyFormatTooltips.iteritems():
--- a/gui/mainFrame.py
+++ b/gui/mainFrame.py
@@ -678,6 +678,10 @@ class MainFrame(wx.Frame):
         sFit = service.Fit.getInstance()
         toClipboard(sFit.exportXml(None, self.getActiveFit()))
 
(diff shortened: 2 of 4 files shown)


blitzmann:
(☞ﾟ∀ﾟ)☞ You're awesome.

I'll take a look later and test, but looking at the code I don't see any issues.

Ebag333:
Great start.

I tested it a bit, and while it exports the fit just fine, it doesn't export the following:
- Implants (Take a look at EFT Implants for some code to steal)
- Boosters
- Fighters (Don't feel bad about this, the EFT export doesn't handle fighters either....)

It seems to export drones and cargo okay. 

The code looks good, not spotting any issues there.

petosorus:
I added Implants, Boosters (they are not covered by implants) and Fighters.

I wonder if charges need a quantity. For now we have
```
Light Electron Blaster II
Void S
```
but may need
```
Light Electron Blaster II
Void S x120
```

I don't have Eve to test right now, but should have it around 12 Eve time.

petosorus:
On a side note, should EFT export add Fighters and Boosters that are not in cargo?

Ebag333:
>I wonder if charges need a quantity. 

They should. The charges out of cargo have the amount, but loaded charges just has a single one. It'd be nice if they reflected the amount loaded. 

Listing them multiple times is OK too, EVE will combine them automatically. 

>On a side note, should EFT export add Fighters and Boosters that are not in cargo?

Well, while you are in there, might as well. 

I'm absolutely amazed that no one has reported on this yet.

petosorus:
>>I wonder if charges need a quantity.

>They should. The charges out of cargo have the amount, but loaded charges just has a single one. It'd be nice if they reflected the amount loaded. 

I just did this now that I am online, seems to work.

>>On a side note, should EFT export add Fighters and Boosters that are not in cargo?

>Well, while you are in there, might as well.

>I'm absolutely amazed that no one has reported on this yet. 

I'm amazed too about the Fighters. I wasn't sure of the Boosters export legitimacy because it's under the Booster tab, like projected effects, and not in cargo. 
Should this be done on a new branch or on the same?

Ebag333:
>I'm amazed too about the Fighters. I wasn't sure of the Boosters export legitimacy because it's under the Booster tab, like projected effects, and not in cargo. 
>Should this be done on a new branch or on the same?

I did open an issue for fighters that you can reference. 

Umm. Boosters (and implants) are complicated and ugly. The current EFT export with implants gets the implant section rejected. If you put them under cargo, they will import but are invisible. 

Boosters are technically just implants, so I'm rather tempted just to say export them under the implant section for EFT. Better solution is probably to export them as cargo for the standard export. 

It's all exports, so the same branch should be fine.

blitzmann:
I should mention that whatever we do here with the boosters, we need to ensure it actually imports correctly into both EVE and EFT. :)

Looking good so far though, thanks for the help @petosorus!

Ebag333:
@blitzmann 
>I should mention that whatever we do here with the boosters, we need to ensure it actually imports correctly into both EVE and EFT. :)

So does it count as correct if it rejects the section, or imports it but it's invisible? 😆 

>Umm. Boosters (and implants) are complicated and ugly. The current EFT export with implants gets the implant section rejected. If you put them under cargo, they will import but are invisible.

On the subject of boosters vs implants, why do we handle them differently when in game boosters simply are implants that expire?

blitzmann:
>On the subject of boosters vs implants, why do we handle them differently when in game boosters simply are implants that expire?

Because that's how it's always been. I don't know if there is a real reason. There's been no reason to change it (and indeed, if we ever get around to adding booster side effects, they should stay separate)

Ebag333:
I don't want to hijack this thread further, but as food for though if we treated them as implants then we could eliminate a good chunk of code that's essentially just duplicated.

(Potentially also eliminate a tab from the gui, since there's no real reason we couldn't do both in a single tab.)

blitzmann:
I'd rather not divert efforts into merging implants and boosters. The code is already there, and works. If we ever implement booster side effects, then they will have to be separate, both on the back end and the GUI, to support the effects attached to them.

petosorus:
I added fighters to the EFT exports, according to #736.

Concerning boosters in EFT, I decided to not export them with the base EFT export, as we don't export implants. 
I included them in the EFT Implants export, but like the implants, they are not recognized when the fit is imported in game. I left them in the export for consistency.

blitzmann:
excellent. I'll be merging this with the next major release (have a planned point release coming up if I ever get off my ass and do it). Thanks a bunch!

PR #735, opened 2016-09-15
Enable job for processing fit calculations

Ebag333:
This is just a proof of concept....

There are numerous scenarios where the fitting job gets recalculated multiple times.

Examples:
On first fit load (3 times)
If a user clicks too quickly (turning on/off multiple modules, adding modules, etc)

This splits the recalc function off into a scheduled job.  A timer is created for 1 second from an event to run the job.  If the user is impatient and makes further changes, the existing job is cancelled and a new job/timer is created.

Both have a short delay (1s) to prevent a race condition from creating multiple jobs that run at the same time.

This _mostly_ works.  The GUI does not get updated always properly, sometimes does not show the correct values until the next job is ran.

Perhaps @blitzmann can chime in on a way to force a full gui refresh....

### DIFF
--- a/service/fit.py
+++ b/service/fit.py
@@ -22,7 +22,6 @@ import copy
 import threading
 import logging
 import wx
-import gui
 import sched, time
 from codecs import open
 
--- a/service/fit.py
+++ b/service/fit.py
@@ -19,6 +19,7 @@
 
 import locale
 import copy
+import config
 import threading
 import logging
 import wx
@@ -44,7 +45,8 @@ from service.port import Port
 logger = logging.getLogger(__name__)
 
 schedule = sched.scheduler(time.time, time.sleep)
-recalcJob = []
+recalcJobID = []
+
 
 class FitBackupThread(threading.Thread):
     def __init__(self, path, callback):
@@ -98,8 +100,8 @@ class Fit(object):
         self.character = Character.getInstance().all5()
         self.booster = False
         self.dirtyFitIDs = set()
-        RecalcTimer = RecalcJob()
-        RecalcTimer.start()
+        self.RecalcTimer = RecalcJob()
+        self.RecalcTimer.start()
 
         serviceFittingDefaultOptions = {
             "useGlobalCharacter": False,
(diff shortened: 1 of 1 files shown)


service/fit.py
```diff
@@ -1125,8 +1128,41 @@ def refreshFit(self, fitID):
         self.recalc(fit)
 
     def recalc(self, fit, withBoosters=True):
+        global recalcJob
+
+        if not schedule.empty():
+            if recalcJob in schedule.queue:
+                logger.debug("Job already scheduled. Cancelling existing job")
+                schedule.cancel(recalcJob)
+                pass
```
Ebag333: I know I don't need the 
```
     pass
else:
     pass
```
here.  This is just simply for some flow control, and I was experimenting with different ways of handling creating the job.

blitzmann:
Wow, funny, I was actually just thinking about doing something like this while working on a similar issue at work where we had to a timeout to pause the execution and cancel it if it was requested again.

Haven't taken a look at the code yet. Some things: does this force a 1 second delay between all fit recalculations? So, turn module off, wait one second? Turn module on, wait another second? This idea would work well for those instances where the fit is recalculated many times in quick succession, but for every day use where users are expecting a near instant result, dunno if it's what we want? Perhaps delegate this to a preference, or only do it for specific events (fit loading, if we detect user has made more than 1 modification in 1 second, etc)

Also, as I've stated before, there was a reason that fits do multiple calculations the first time they load. I don't quite remember it, and it may very well simply be legacy code, but I want to say that one calc depended on another. Although, thinking about it now, that can't be it since a recalc clears previous values. So I dunno.

Interesting concept anyway. Will take a look when I get a chance.

Ebag333:
Basically when you change something it does two things:
1. Creates a job to run the recalc 1s from now (via scheduler).
2. Sets a job (using threading, so it's not holding up the gui) for 1 down the road.  After the timer goes off, it executes all jobs in the scheduler queue.

If anything triggers a recalc, the existing scheduler job is purged and a new one created.  A new threading job is created for 1s down the road (since from what I've read the answer to "can you delete a scheduled threading job" is "don't", I didn't try and purge the existing one).

So if you make a bunch of changes rapidly, the update will most likely fire off when the first threading job executes (and since we have multiple jobs, most likely at most 1s after you stop making changes).  If you're slow enough (and have perfect timing), then theoretically you can keep it from executing forever, but in practice it's about a second before it updates.

I tried setting lower values and found that it'd kick off multiple update jobs during "normal" operations (like when a fit is first loaded, and it's refreshed 3 times).  1s for the threading job and 1s for the scheduler job seemed to be about as low as we could get without recalculating the fit faster than the code could actually keep up (in other words, it takes longer than 3 seconds for a fit to load...)

The general "feel" doesn't feel any more sluggish than Pyfa does normally.  It's a bit hard to tell because the GUI isn't refreshing properly, but I _suspect_ that's mostly a GUI refresh issue.

If we moved stuff down into the `recalc` function, we could also cut out a _lot_ of code from `fit.py`.  Most of the functions have `eos.db.commit()` in them, for example, and there's probably other stuff we could move into a centralized location.  So that might help clean up fit.py a bit.

blitzmann:
> The general "feel" doesn't feel any more sluggish than Pyfa does normally. 

Honestly this is all I care about. pyfa already has a (credible) reputation of being slow as balls, so we shouldn't purposefully make it any slower. I'll have to test this out. I am not sure what the GUI issue could be... IIRC whenever a fit is recalc'd a wx event is fired (`FitChanged`), which is then caught by the various GUI elements to update. If that event is being fired immediately, but the calc doesn't happen for 1 second, that could be the issue.

Ebag333:
>IIRC whenever a fit is recalc'd a wx event is fired (FitChanged), which is then caught by the various GUI elements to update. If that event is being fired immediately, but the calc doesn't happen for 1 second, that could be the issue.

Great. wxPython. You know how much I love that.

>pyfa already has a (credible) reputation of being slow as balls, so we shouldn't purposefully make it any slower.

A _huge_ amount of the slow down comes because a recalc is triggered with every bloody action you do.  Sometimes one action will trigger two or more recalcs.

As an example, if it takes .5 seconds to recalc (and that's probably being generous) and I add 8 guns to a fit, that's a 4 second delay.  If we broke out the recalc (and gui refresh), then I can add all 8 guns fairly quickly (there's still a delay as it has to add them to the fitting window, but it's much shorter), and then there's a 1 second delay for the fit stats to update.  Call it 2 seconds (it's probably actually even less as it doesn't take a full second to add all 8 guns).

I haven't benchmarked it, but I'd hazard a guess that in that scenario it's more like 8-10 seconds  for the current branch, and 2-3 for this branch.

We simply keep adding more and more things that will trigger a refresh, so this problem is going to keep getting worse rather than better (well, okay, when we gut the fleet boosters it'll get better).  Even though this hard sets a delay in, at least it's a quantifiable and known delay (and one that we can largely control).

Ebag333:
Replaced by #848

PR #744, opened 2016-09-19
Fix three...I mean...four XL Rig effect files

Ebag333:
```
    fit.modules.filteredItemIncrease(lambda mod: mod.item.group.name == "Structure Doomsday Weapon", "lightningWeaponTargetAmount", src.getModifiedItemAttr("structureRigDoomsdayTargetAmountBonus"), stackingPenalties=True)
    # This is all wrong.  We need to modify the module attached to the ship (citadel) and add the bonus from the rig.
    # So Rig -> Ship -> Module
```

I'm not sure how to go about this, so disabled for now....

The problem is that the ship doesn't have the attribute, and passing it through this way doesn't work.  For some reason, `fit.modules` doesn't seem to have any of the normal classes/functions attached to it. 

The other two effect files are fixed, however, and this one is disabled until we can figure it out.

### DIFF
--- a/eos/effects/structurerigdoomsdaydamageloss.py
+++ b/eos/effects/structurerigdoomsdaydamageloss.py
@@ -1,5 +1,6 @@
 type = "passive"
 def handler(fit, src, context):
-    fit.modules.filteredItemMultiply(lambda mod: mod.item.group.name == "Structure Doomsday Weapon",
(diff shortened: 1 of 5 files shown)


Ebag333:
@blitzmann I think this is ready to merge.  Can you take a look? :)

I know a couple of these were implemented using loops, but I don't know a clean(er) way to do it otherwise.

blitzmann:
fixed 44fb4a5, not merged. Thanks!

PR #756, opened 2016-10-01
Fixed a few fighter/carrier bugs:

MrNukealizer:
* Omnidirectional Tracking Link overheat applies to explosion radius/velocity bonuses.
* Networked Sensor Array affects fighters' sensor strengths.
* Long range heavy fighters no longer appear to require Bomb Deployment.
* Carriers and supercarriers are no longer capped at 300km lock range.
* Fixed superiority fighters' main attack and the ammo count for limited abilities.
* Added the Evasive Maneuvers ability to superiority fighters.
* Made fighter DPS calculations more accurate when factoring reload time. It now realizes that all abilities stop while reloading, instead of just the ones with charges.

### DIFF
--- a/eos/effects/fighterabilityevasivemaneuvers.py
+++ b/eos/effects/fighterabilityevasivemaneuvers.py
@@ -1,0 +1,24 @@
+# Not used by any item
+"""
+Since fighter abilities do not have any sort of item entity in the EVE database, we must derive the abilities from the
+effects, and thus this effect file contains some custom information useful only to fighters.
+"""
+# User-friendly name for the ability
+displayName = "Evasive Maneuvers"
+
+prefix = "fighterAbilityEvasiveManeuvers"
+
+# Is ability applied to the fighter squad as a whole, or per fighter?
+grouped = True
+
+type = "active"
+runTime = "late"
+def handler(fit, module, context):
+    module.boostItemAttr("maxVelocity", module.getModifiedItemAttr("fighterAbilityEvasiveManeuversSpeedBonus"))
+    module.boostItemAttr("signatureRadius", module.getModifiedItemAttr("fighterAbilityEvasiveManeuversSignatureRadiusBonus"), stackingPenalties = True)
+    
+	# These may not have stacking penalties, but there's nothing else that affects the attributes yet to check.
+    module.multiplyItemAttr("shieldEmDamageResonance", module.getModifiedItemAttr("fighterAbilityEvasiveManeuversEmResonance"), stackingPenalties = True)
+    module.multiplyItemAttr("shieldThermalDamageResonance", module.getModifiedItemAttr("fighterAbilityEvasiveManeuversThermResonance"), stackingPenalties = True)
+    module.multiplyItemAttr("shieldKineticDamageResonance", module.getModifiedItemAttr("fighterAbilityEvasiveManeuversKinResonance"), stackingPenalties = True)
+    module.multiplyItemAttr("shieldExplosiveDamageResonance", module.getModifiedItemAttr("fighterAbilityEvasiveManeuversExpResonance"), stackingPenalties = True)
--- a/eos/effects/modulebonusnetworkedsensorarray.py
+++ b/eos/effects/modulebonusnetworkedsensorarray.py
@@ -7,15 +7,10 @@ def handler(fit, src, context):
     fit.ship.boostItemAttr("scanResolution", src.getModifiedItemAttr("scanResolutionBonus"), stackingPenalties=True)
 
     for scanType in ('Magnetometric', 'Ladar', 'Gravimetric', 'Radar'):
-        fit.ship.boostItemAttr("scan{}Strength".format(scanType),
-                               src.getModifiedItemAttr("scan{}StrengthPercent".format(scanType)),
-                               stackingPenalties=True)
-                               
-    for fighter in fit.fighters:
-        for scanType in ('Magnetometric', 'Ladar', 'Gravimetric', 'Radar'):
-            fighter.boostItemAttr("scan{}Strength".format(scanType),
-                               src.getModifiedItemAttr("scan{}StrengthPercent".format(scanType)),
-                               stackingPenalties=True)
+        attr = "scan{}Strength".format(scanType)
+        bonus = src.getModifiedItemAttr("scan{}StrengthPercent".format(scanType))
+        fit.ship.boostItemAttr(attr, bonus, stackingPenalties=True)
+        fit.fighters.filteredItemBoost(lambda mod: mod.item.requiresSkill("Fighters"), attr, bonus, stackingPenalties=True)
 
     # EW cap need increase
     groups = [
--- a/eos/effects/modulebonusnetworkedsensorarray.py
+++ b/eos/effects/modulebonusnetworkedsensorarray.py
@@ -4,11 +4,16 @@
 # Module: Networked Sensor Array
 type = "active"
 def handler(fit, src, context):
-    fit.ship.multiplyItemAttr("maxTargetRange", src.getModifiedItemAttr("maxTargetRangeMultiplier"), stackingPenalties=True, penaltyGroup="postMul")
     fit.ship.boostItemAttr("scanResolution", src.getModifiedItemAttr("scanResolutionBonus"), stackingPenalties=True)
 
     for scanType in ('Magnetometric', 'Ladar', 'Gravimetric', 'Radar'):
         fit.ship.boostItemAttr("scan{}Strength".format(scanType),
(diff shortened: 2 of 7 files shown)


eos/effects/modulebonusnetworkedsensorarray.py
```diff
@@ -4,13 +4,18 @@
...
     for scanType in ('Magnetometric', 'Ladar', 'Gravimetric', 'Radar'):
         fit.ship.boostItemAttr("scan{}Strength".format(scanType),
                                src.getModifiedItemAttr("scan{}StrengthPercent".format(scanType)),
                                stackingPenalties=True)
+                               
+    for fighter in fit.fighters:
+        for scanType in ('Magnetometric', 'Ladar', 'Gravimetric', 'Radar'):
+            fighter.boostItemAttr("scan{}Strength".format(scanType),
+                               src.getModifiedItemAttr("scan{}StrengthPercent".format(scanType)),
+                               stackingPenalties=True)
```
blitzmann: Instead of looping over the fighters themselves, try something like this:

```
for scanType in ('Magnetometric', 'Ladar', 'Gravimetric', 'Radar'):         
    fit.fighters.filteredItemBoost(lambda mod: mod.item.requiresSkill("Fighters"), "scan{}Strength".format(scanType), getModifiedItemAttr("scan{}StrengthPercent".format(scanType)), True)
```

(untested)

This basically accomplishes the same thing, but it's the construct uses in nearly all other effect files.

eos/saveddata/ship.py
```diff
@@ -52,6 +52,9 @@ def __init__(self, item, parent=None):
         self.__itemModifiedAttributes.original = dict(self.item.attributes)
         self.__itemModifiedAttributes.original.update(self.EXTRA_ATTRIBUTES)
         self.__itemModifiedAttributes.overrides = self.item.overrides
+        
+        if "maximumRangeCap" in self.__itemModifiedAttributes.original:
+            cappingAttrKeyCache["maxTargetRange"] = "maximumRangeCap"
```
blitzmann: are you sure you want to set this to a string and not the value? I haven't tested these yet so I'm not sure, just double checking. :)

eos/effects/modulebonusnetworkedsensorarray.py
MrNukealizer: Done.

eos/saveddata/ship.py
MrNukealizer: It needs to be either the name or ID of the attribute that holds the new cap value.

eos/effects/modulebonusnetworkedsensorarray.py
blitzmann: Ah, that makes more sense than what I suggested! Great work :)

eos/saveddata/ship.py
blitzmann: Awesome, thanks for clarifying. Just wasn't sure of the mechanics behind it :)

blitzmann:
I'll be pulling this down and actually testing it soon, but everything look great code-wise. :)

eos/saveddata/ship.py
blitzmann: Also, was the range cap not working to begin with? I thought it was. What was the reason for this change, specifically?

MrNukealizer: The range cap works fine, it just doesn't currently have any exception for carriers and supers. A few months ago something was broken and you changed the capping system to fix it, but the current code never changes the cap from the default.

eos/effects/modulebonusnetworkedsensorarray.py
Ebag333: >This basically accomplishes the same thing, but it's the construct uses in nearly all other effect files.

Oh the second half of this statement is _such_ a lie. :D

Still, it's better to use it.

eos/effects/modulebonusomnidirectionaltrackinglinkoverload.py
```diff
@@ -4,6 +4,9 @@
...
 def handler(fit, module, context):
-    module.boostItemAttr("maxRangeBonus", module.getModifiedItemAttr("overloadTrackingModuleStrengthBonus"))
-    module.boostItemAttr("falloffBonus", module.getModifiedItemAttr("overloadTrackingModuleStrengthBonus"))
-    module.boostItemAttr("trackingSpeedBonus", module.getModifiedItemAttr("overloadTrackingModuleStrengthBonus"))
+    overloadBonus = module.getModifiedItemAttr("overloadTrackingModuleStrengthBonus")
+    module.boostItemAttr("maxRangeBonus", overloadBonus)
+    module.boostItemAttr("falloffBonus", overloadBonus)
+    module.boostItemAttr("trackingSpeedBonus", overloadBonus)
+    module.boostItemAttr("aoeCloudSizeBonus", overloadBonus)
+    module.boostItemAttr("aoeVelocityBonus", overloadBonus)
```
Ebag333: Where are you getting
```
aoeCloudSizeBonus
aoeVelocityBonus
```
from?

The info says:
```
Improves the optimal range, falloff range, explosion radius and explosion velocity of all fighter weapons.
```

Looking at a Templar, I would think that the stats needed to be modified would be:

```
fighterAbilityAttackMissileExplosionRadius
fighterAbilityAttackMissileExplosionVelocity
fighterAbilityMissilesExplosionRadius
fighterAbilityMissilesExplosionVelocity
```
The other fighters might have different stats to modify.

eos/effects/fighterabilityevasivemaneuvers.py
```diff
@@ -0,0 +1,24 @@
+# Not used by any item
+"""
+Since fighter abilities do not have any sort of item entity in the EVE database, we must derive the abilities from the
+effects, and thus this effect file contains some custom information useful only to fighters.
+"""
+# User-friendly name for the ability
+displayName = "Evasive Maneuvers"
```
Ebag333: Why do we need this? It doesn't seem to be used anywhere.

Ebag333: Why do we need this? Doesn't seem to be used anywhere.

eos/effects/fighterabilityevasivemaneuvers.py
```diff
@@ -0,0 +1,24 @@
...
+
+prefix = "fighterAbilityEvasiveManeuvers"
```
Ebag333: (Same question as above.)

eos/effects/fighterabilityevasivemaneuvers.py
```diff
@@ -0,0 +1,24 @@
...
+
+# Is ability applied to the fighter squad as a whole, or per fighter?
+grouped = True
+
+type = "active"
+runTime = "late"
+def handler(fit, module, context):
```
Ebag333: Minor quibble, but it should probably be:
`+def handler(fit, container, context):`

Module implies it's actually a module, when it's not, it's a fighter.

I _would_ say use `fighter` instead, but unfortunately we can have effects that cross boundaries, so might as well standardize on something fairly generic.

It's not a major issue, lots of other effect files don't do this, but I've started doing this when creating new effect files (@blitzmann and I had a discussion about it after I submitted code doing the same thing :) ).

MrNukealizer: I basically just copied eos/effects/fighterabilitymicrowarpdrive.py and tweaked it, so I'm not an expert on this stuff.

eos/effects/fighterabilityevasivemaneuvers.py
Ebag333: Fair enough

eos/effects/fighterabilityevasivemaneuvers.py
Ebag333: Same as above

eos/effects/fighterabilityevasivemaneuvers.py
blitzmann: What he said. It's my hacky way of implementing it before they added this data to the cache:

https://gist.github.com/blitzmann/935c2d5b178d62205ba88a30463e8eda

Which I don't believe is even in the SDE or on CREST. v0v

eos/effects/fighterabilityevasivemaneuvers.py
blitzmann: Yeah, I've been calling it `src` as it's whatever item the effect is coming from. I'm not worried about that though.

eos/effects/fighterabilityevasivemaneuvers.py
Ebag333: Time to go beat CCP until they add it then. :)

blitzmann:
Thanks @MrNukealizer, this is a big help!

Issue #746, opened 2016-09-21
Feature Request : HTTP Proxy Username/Password


kelyen:
Hello,

Can we have the possibility to add a username and password in : 

Pyfa > Window > Preferences > Network > Manually Add Proxy

Thank you,

blitzmann:
Hey there,

This is a duplicate of #26, however I'm going to close that one and keep this one open simply because that one is buried in years of issues and this one might get some visibility. 

I personally don't have the knowledge to handle this feature. I've never used proxies, don't know how to set one up, and trying to get it to work just seems... It would take a bit of research on my part, and that's time I don't currently have :(

Maybe someone has the knowledge behind how proxy stuff works and wants to tackle this? If so, leave a comment!

minlexx:
I think this feature can be done. I always wanted to do some hacking with Pyfa, maybe I should start with something simple like this...

blitzmann:
@minlexx excellent! It would be a great help. This is where the proxy stuff resides:

https://github.com/pyfa-org/Pyfa/blob/master/service/network.py#L76

webera:
Shouldn't it be already possible to use the current implementation by specifiying the address like that:

http://proxyusername:proxypassword@proxyaddress

so like

http://squid:squid@127.0.0.1

?

minlexx:
@webera Yes, I just tested (took too much time for me to configure squid3 to require a username/password, but now I can reliably test with proxy). It works for me this way, but this is not obvious for users ;) We could still add explicit input fields for proxy username/password in GUI (somewhere below Port: input section), need @blitzmann opinion about this.

![screen](http://puu.sh/rnrG1/bddc686255.png)

P.S. I tested with EVE API: Character Editor ("Get Characters", "Fetch Skills" buttons). Without username:pass there was TCP_DENIED/407 error - "Proxy authorization required", with username:pass TCP_MISS indicates successful connect (cache miss) in squid proxy server log.

blitzmann:
> took too much time for me to configure squid3 to require a username/password

This is exactly why I gave up on it way back when I looked at it.

I am surprised it is as simple as that though. I agree, an explicit username and password field would be beneficial, and we can concatenate the information in the network service. :)

kelyen:
Thank you very much for the syntax. In fact, I didn't know it before. It's working like a charm.

PR #759, opened 2016-10-02
Add login/password for proxy authorization in network settings

minlexx:
Related to issue #746 HTTP Proxy Username/Password
Briefly, changed 3 files:
- service/settings.py: add fields to store proxy login/pass in NetworkSettings class;
- service/network.py: use proxy login/pass info from NetworkSettings and prepend them to proxy address in form of login:password@host:port
- gui/builtinPreferenceViews/pyfaNetworkPreferences.py: GUI input fields to explicitly configure proxy login/pass

Now it looks like this:
 screenshot

### DIFF
--- a/gui/builtinPreferenceViews/pyfaNetworkPreferences.py
+++ b/gui/builtinPreferenceViews/pyfaNetworkPreferences.py
@@ -115,10 +115,10 @@ class PFNetworkPref ( PreferenceView):
         mainSizer.Add( fgAddrSizer, 0, wx.EXPAND, 5)
 
         # proxy auth information: login and pass
-        self.stPSetLogin = wx.StaticText(panel, wx.ID_ANY, u"Proxy login:", wx.DefaultPosition, wx.DefaultSize, 0)
+        self.stPSetLogin = wx.StaticText(panel, wx.ID_ANY, u"Username:", wx.DefaultPosition, wx.DefaultSize, 0)
         self.stPSetLogin.Wrap(-1)
         self.editProxySettingsLogin = wx.TextCtrl(panel, wx.ID_ANY, self.nAuth[0], wx.DefaultPosition, wx.DefaultSize, 0)
-        self.stPSetPassword = wx.StaticText(panel, wx.ID_ANY, u"pass:", wx.DefaultPosition, wx.DefaultSize, 0)
+        self.stPSetPassword = wx.StaticText(panel, wx.ID_ANY, u"Password:", wx.DefaultPosition, wx.DefaultSize, 0)
         self.stPSetPassword.Wrap(-1)
         self.editProxySettingsPassword = wx.TextCtrl(panel, wx.ID_ANY, self.nAuth[1], wx.DefaultPosition,
                                                      wx.DefaultSize, wx.TE_PASSWORD)
--- a/gui/builtinPreferenceViews/pyfaNetworkPreferences.py
+++ b/gui/builtinPreferenceViews/pyfaNetworkPreferences.py
@@ -70,6 +70,8 @@ class PFNetworkPref ( PreferenceView):
         self.nPort = self.settings.getPort()
         self.nType = self.settings.getType()
         self.nAuth = self.settings.getProxyAuthDetails()  # tuple of (login, password)
+        if self.nAuth is None:
+            self.nAuth = ("", "")  # we don't want None here, it should be a tuple
 
         ptypeSizer = wx.BoxSizer( wx.HORIZONTAL )
 
--- a/gui/builtinPreferenceViews/pyfaNetworkPreferences.py
+++ b/gui/builtinPreferenceViews/pyfaNetworkPreferences.py
@@ -69,6 +69,7 @@ class PFNetworkPref ( PreferenceView):
         self.nAddr = self.settings.getAddress()
         self.nPort = self.settings.getPort()
         self.nType = self.settings.getType()
+        self.nAuth = self.settings.getProxyAuthDetails()  # tuple of (login, password)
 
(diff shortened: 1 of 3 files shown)


service/network.py
```diff
@@ -71,13 +72,29 @@ def request(self, url, type, data=None):
...
 
         proxy = NetworkSettings.getInstance().getProxySettings()
         if proxy is not None:
-            proxy = urllib2.ProxyHandler({'https': "{0}:{1}".format(*proxy)})
-            opener = urllib2.build_opener(proxy)
+            # proxy is tuple of (host, port):  (u'192.168.20.1', 3128)
+            # build default proxy handler
+            proxy_handler = urllib2.ProxyHandler({'https': "{0}:{1}".format(proxy[0], proxy[1])})
+            proxy_auth = NetworkSettings.getInstance().getProxyAuthDetails()
+            if proxy_auth is not None:
```
blitzmann: Looking at `getProxyAuthDetails()`, this will never return `None`, instead it will always return a tuple of two strings. So why are we checking if it's `None` here? Doesn't this make a bit more sense:

```
proxy_auth = NetworkSettings.getInstance().getProxyAuthDetails()
if proxy_auth is not None:
    proxy_handler = urllib2.ProxyHandler({'https':  '{0}:{1}@{2}:{3}'.format(proxy_auth[0], proxy_auth[1], proxy[0], proxy[1])})
else:
    proxy_handler = urllib2.ProxyHandler({'https': "{0}:{1}".format(proxy[0], proxy[1])})
```

And then just make the `getProxyAuthDetails()` return None if it doesn't have both username and password?

gui/builtinPreferenceViews/pyfaNetworkPreferences.py
```diff
@@ -111,6 +112,21 @@ def populatePanel( self, panel ):
 
         mainSizer.Add( fgAddrSizer, 0, wx.EXPAND, 5)
 
+        # proxy auth information: login and pass
+        self.stPSetLogin = wx.StaticText(panel, wx.ID_ANY, u"Proxy login:", wx.DefaultPosition, wx.DefaultSize, 0)
+        self.stPSetLogin.Wrap(-1)
+        self.editProxySettingsLogin = wx.TextCtrl(panel, wx.ID_ANY, self.nAuth[0], wx.DefaultPosition, wx.DefaultSize, 0)
+        self.stPSetPassword = wx.StaticText(panel, wx.ID_ANY, u"pass:", wx.DefaultPosition, wx.DefaultSize, 0)
```
blitzmann: This is nitpickey, but can we change the labels to simply say "Username" and "Password"? :D

minlexx:
@blitzmann yeah, handling of return value from getProxyAuthDetails() in network service was my mistake...

blitzmann:
Just an update: Everything looks fine to me, I just haven't had time to verify functionality (at least make sure things don't break, as I don't have a proxy to text with). I assure you it'll be merged before the next major release :D

service/settings.py
```diff
@@ -198,13 +209,33 @@ def autodetect(self):
...
+        if self.getMode() == self.PROXY_MODE_MANUAL:
             return (self.getAddress(), int(self.getPort()))
 
+    def getProxyAuthDetails(self):
+        if self.getMode() == self.PROXY_MODE_NONE:
+            return None
+        if (self.serviceNetworkSettings["login"] is None) or (self.serviceNetworkSettings["password"] is None):
+            return None
+        # in all other cases, return tuple of (login, password)
+        return (self.serviceNetworkSettings["login"], self.serviceNetworkSettings["password"])
```
Ebag333: Why are we returning a tuple here?  My first thought was to return a dict, but then looking at it again, why wouldn't we just:
`return self.serviceNetworkSettings`

If we want to return multiple values separately for some reason, you're better off using a dict so that the returned values actually have something meaningful to reference them by, and you're not having to do `returnValue[1]` type stuff.

`return {'login': self.serviceNetworkSettings["login"], 'password': self.serviceNetworkSettings["password"]}`

blitzmann: `tuple` is technically semantically correct, as username and password go hand in hand, and it also matches how hostname and port are utilized as well. Although I personally don't care either way. Returning `self.serviceNetworkSettings` works just as well. It's authors discretion :)

minlexx: I used tuple because: 
1) I saw similar code was for proxy address and port return https://github.com/pyfa-org/Pyfa/blob/master/service/settings.py#L206 ```return (self.getAddress(), int(self.getPort()))```
2) I left a comment near the code that is using the result
```self.nAuth = self.settings.getProxyAuthDetails()  # tuple of (login, password)```

Agree with @Ebag333 that dict with keys makes code a bit more readable, will do whatever you like.

Ebag333: `self.serviceNetworkSettings` is already a dict so that should work if you wanted to just return that.

As long as you've commented the code it's not a major issue (like I said, it's a minor quibble).  I prefer the dict as it's easier to follow when tracing the code, but it's up to you.

>I saw similar code

That would be from someone named `HomeWorld` back in 2011.

A _lot_ of Pyfa is *very* old, and some of the services are some of the oldest code around.  Most of `settings.py` is from the 2010-2012 days, and I wouldn't exactly adhere myself to those designs.

Not that it's necessarily bad code (it works fine, and like I said earlier, this is a minor quibble), but a lot has changed in the last 6 years.

blitzmann: > Agree with @Ebag333 that dict with keys makes code a bit more readable, will do whatever you like.

As stated, I really don't care either way. `tuple` matches what's already there and makes semantic sense, dictionary is more readable and intuitive.

Issue #762, opened 2016-10-07
Mining Drones


Ebag333:
Couple issues with Mining drones
1. The mining drone rigs have a stacking penalty. They shouldn't.
2. Drones have an innate bonus on them of 150% (for T2 mining drones). This isn't calculated.

> Neugeniko (Jita Miners Indices)
> I took 1 drone and it mined 90m3 of veldspar in 1 cycle of 60 sec
> 90.0 m3 or 900 units
> Fit is orca with 2x Large Drone Mining Augmentor II, 1x Large Drone Mining Augmentor I
> T2 mining drone

Once the stacking penalty is removed off the rigs, T2 mining drones in Pyfa show as 60 m3 per cycle.  They should be 90, which seems to be where that 150% bonus comes into play.

blitzmann:
Both items should be fixed

blitzmann:
Reopening - I like to close these things once the stable release is out :)

PR #765, opened 2016-10-08
Mining Drones: Apply Mining Effect

Ebag333:
This effect is only used by mining drones, and it's not been
implemented.

See #762 for more info

### DIFF
--- a/eos/effects/mining.py
+++ b/eos/effects/mining.py
@@ -7,9 +7,6 @@ type = "passive"
 def handler(fit, container, context):
     miningDroneAmountPercent = container.getModifiedItemAttr("miningDroneAmountPercent")
     if (miningDroneAmountPercent is None) or (miningDroneAmountPercent == 0):
-        miningDroneAmountPercent = 1
+        pass
     else:
-        miningDroneAmountPercent = miningDroneAmountPercent/100
-
-    fit.drones.filteredItemMultiply(lambda drone: drone.item.group.name == "Mining Drone",
-                                 "miningAmount", miningDroneAmountPercent)
+        container.multiplyItemAttr("miningAmount", miningDroneAmountPercent/100)
--- a/eos/effects/mining.py
+++ b/eos/effects/mining.py
@@ -1,0 +1,15 @@
+# miningDroneOperationMiningAmountBonusPostPercentMiningDroneAmountPercentChar
+#
+# Used by:
+# Mining Drones
+
+type = "passive"
+def handler(fit, container, context):
+    miningDroneAmountPercent = container.getModifiedItemAttr("miningDroneAmountPercent")
+    if (miningDroneAmountPercent is None) or (miningDroneAmountPercent == 0):
+        miningDroneAmountPercent = 1
+    else:
+        miningDroneAmountPercent = miningDroneAmountPercent/100
+
+    fit.drones.filteredItemMultiply(lambda drone: drone.item.group.name == "Mining Drone",
+                                 "miningAmount", miningDroneAmountPercent)


Ebag333:
Effect 17 is only used by mining drones.

```
SELECT * FROM dgmtypeeffects DTE 
LEFT JOIN invtypes IT on DTE.typeID = IT.typeID
WHERE DTE.effectID = '17'
```

```
"1202"	"17"	"1202"	"Civilian Mining Drone"	"Civilian Mining Drone"
"3218"	"17"	"3218"	"Harvester Mining Drone"	"The Harvester is the most efficient mining drone money can buy."
"41030"	"17"	"41030"	"Excavator Drone"	
"10246"	"17"	"10246"	"Mining Drone I"	"Mining Drone"
"10250"	"17"	"10250"	"Mining Drone II"	"Mining Drone"
```

Ebag333:
So moment of honesty here.

If you look at the mining drone out of the market, this stat doesn't exist in the attribute list.

Applied to the Orca, it does.

I have no idea where it comes from.  It's a brand new stat, but I didn't find it searching through the code base (I might have just missed it, heaven knows we have enough files with "mining" and "drone" in them).

It matches the value that we need to have exactly, so it makes sense that we need to apply it.  I just can't figure out where it comes from.

Ebag333:
Verified with FuzzySteve, a Skiff with T1 mining drone and V skills should pull in 37.5 m3 per cycle.  

So this is verified against both the Orca and a Skiff.

blitzmann:
I don't believe this is the proper fix. `mining` hasn't been an implemented effect in forever, it it wierd to me that we just started getting concerns about it...

Not only that, but this patch increases the mining amount exponentially for every drone fit and active (total mining for the fit):

1. 1.42m2/s
2. 4.25m2/s
3. 9.56m2/s
4. 19.1m2/s
5. 35.9m2/s

Ebag333:
>Not only that, but this patch increases the mining amount exponentially for every drone fit and active

Yup, by 50%. More if you're using rigs (since those were stacking penalized and shouldn't be).

I ain't going to lie, mining lasers and drones are weird, implemented weird, have weird stats, and just all around annoying.

>`mining` hasn't been an implemented effect in forever

Using `mining` is a bit hackish because we need an effect that we can take the `miningDroneAmountPercent` stat and apply it to drones.  There's not really any other good spots for that, and since it wasn't being used and only applies to drones it works.

Does CCP apply `miningDroneAmountPercent` under the `mining` effect? vOv

>it it wierd to me that we just started getting concerns about it...

Yes, because mining drones were _so_ popular with the hardcore min/maxers who do tons of theorycrafting.  I mean, you _never_ see anything _but_ mining drones being used! [/sarcasm]

In all seriousness, hardly anyone used mining drones seriously, they just left combat drones out to better AFK mine and protect against rats.  That's changing drastically with this patch, so it's no surprise to me that people are starting to pay attention to what the numbers _actually_ are.

I confirmed the numbers against an Orca and Skiff, _with_ `miningDroneAmountPercent` applied (which we do not do today) the numbers match.  _Without_ it applied, it's 50% too low for both ships.

If you want to log onto SiSi and try out mining ships and compare numbers, feel free.  I have _no_ mining skills at all, so it's not much of an option for me, but from everything that I have found the numbers are indeed correct with this patch applied.

blitzmann:
So, here's some problems:

```
fit.drones.filteredItemMultiply(lambda drone: drone.item.group.name == "Mining Drone",
                                 "miningAmount", miningDroneAmountPercent)
```

You don't want the `filteredItemMultiply` there as that is giving each drone an increase. it should be something like 

```
container.multiplyItemAttr("miningAmount", miningDroneAmountPercent)
```

(off topic, but you don't want `Multiply` period. It's a percentage boost, so use the `boostItemAttr` method)

The problem with that is the way the drones are implemented. They have an `amountActive` property, and we do a loop based on this and run the effect x number of times, where x is the amount of drones in that stack. So it will still run 5 times on the drone stack when we only want to run it once.

I have an idea of how to fix this. it's a hack, but wtf isn't these days

Ebag333:
>The problem with that is the way the drones are implemented. They have an amountActive property, and we do a loop based on this and run the effect x number of times, where x is the amount of drones in that stack. So it will still run 5 times on the drone stack when we only want to run it once.

Whoops.  Well, it works perfectly fine with 1 drone!

>.>

<.<

Ebag333:
>off topic, but you don't want Multiply period. It's a percentage boost, so use the boostItemAttr method

I think you do and here's why.

The stat is actually 150, not 1.5 and not 50.  So if you use `boostItemAttr`, it goes to a 250% boost.

I was getting around that by dividing by 100, so it ends up as 1.5, then `Multiply` works just fine.

Could also subtract 100 and use `boostItemAttr`, but either way it's `maths`.

Ebag333:
Pushed a fix, numbers are right now when you have more than one drone (don't get 1.5x multiplier for each drone).

Maths are hard.

https://puu.sh/rLFpb/25685e4626.png

blitzmann:
>The stat is actually 150, not 1.5 and not 50. So if you use boostItemAttr, it goes to a 250% boost.

Fair enough, didn't think about that. But you're still going to get additional runs of the effect if they are stacked. I will merge this into a local branch and make my changes and test, then push

blitzmann:
Should be fixed. If there are any issues with it, please report in the issue :) Thanks for the help!

Issue #714, opened 2016-08-31
Few bugs when a fit projected by itself, and remote sensor booster ECCM bug


sephasis:
![self projected proteus](https://cloud.githubusercontent.com/assets/21352052/18114146/e2f66450-6f6f-11e6-865d-95442edd3a24.PNG)

when a fit projected by itself, DPS and Recharge Rate shows weird.

This proteus' DPS is 605 and armor repairing rate must be 188/s.
(it's like 50% remote repairing effective bonus of Adaptive Augmenter is not applied.)

---

![remote sensor booster eccm](https://cloud.githubusercontent.com/assets/21352052/18114211/6a2530dc-6f70-11e6-8cf5-18a61eeede2b.PNG)

This Osprey is jammed by a Griffin with 5 Gravimetric ECM but its sensor strength is being boosted by another osprey with 4 remote sensor boosters (ECCM Script).
so sensor strength is 183 but pyfa says jamming chance is 97.5%.
it's same chance with normal Osprey without any boosting of remote sensor boosters.

blitzmann:
I'm not sure I understand the first bit with the prteus. Can you lay out exactly what values you expect to see vs the values that pyfa is showing?

As for the second issue with the sensor strength, I'll look into it. :)

sephasis:
I cannot reenact about DPS bug in this home computerm i'll check about that later but i could reenact remote repair bug so i'll show it.

![image](https://cloud.githubusercontent.com/assets/21352052/18224650/236d8720-7218-11e6-9e30-6abeea28709b.png)

proteus's remote repairing rate is +188/s. It was about +125.3/s but 'Proteus Defensive - Adaptive Augmenter' subsystem gives 50% bonus.

![image](https://cloud.githubusercontent.com/assets/21352052/18224662/70f1fc24-7218-11e6-9088-17a76d8d296a.png)

But when i set something repaired by proteus, pyfa says the rate is +125.3/s (raw HP).

![image](https://cloud.githubusercontent.com/assets/21352052/18224677/cd2227e4-7218-11e6-95f6-4f93310f12ed.png)

whatever proteus repairs...

![image](https://cloud.githubusercontent.com/assets/21352052/18224683/19e7596e-7219-11e6-9ac0-5d9773daf3fd.png)

another T3 cruisers are same. but another ships with logistics bonus like T1 frigs or cruisers, or T2 logistic frigates, etc, don't have any bug like this.

Ebag333:
[...]

PR #770, opened 2016-10-16
Subsystems calculate after projection calcs

Ebag333:
Moved the 4 remote rep subsystem effect bonuses to calculate early in the engine.  This correctly calculates these bonuses for effects, previously these effect bonuses wouldn't be calculated until after projection calcs ran.


Ebag333:
See #714 for more information and screenshots.

blitzmann:
I'll be real, I didn't really test these changes, but it makes sense. :D

PR #771, opened 2016-10-16
Applied Mining Crystal Bonus to modules

Ebag333:
Previously the mining crystal bonus was not applied to modules, so adding a crystal didn't change the amount mined at all.  This now applies.  Left the old application in place, so it still applies to a nonsense `specialtyMiningAmount` stat if you want to see that seperated out for some weird reason.

### DIFF
--- a/eos/effects/mininginfomultiplier.py
+++ b/eos/effects/mininginfomultiplier.py
@@ -6,3 +6,5 @@
 type = "passive"
 def handler(fit, module, context):
     module.multiplyItemAttr("specialtyMiningAmount", module.getModifiedChargeAttr("specialisationAsteroidYieldMultiplier"))
+    fit.modules.filteredItemMultiply(lambda mod: mod.item.requiresSkill("Mining"),
+                              "miningAmount", module.getModifiedChargeAttr("specialisationAsteroidYieldMultiplier"))
--- a/eos/effects/mininginfomultiplier.py
+++ b/eos/effects/mininginfomultiplier.py
@@ -5,6 +5,5 @@
 # Charges named like: Mining Crystal (32 of 32)
 type = "passive"
 def handler(fit, module, context):
-    module.multiplyItemAttr("specialtyMiningAmount", module.getModifiedChargeAttr("specialisationAsteroidYieldMultiplier"))
-    fit.modules.filteredItemMultiply(lambda mod: mod.item.requiresSkill("Mining"),
-                              "miningAmount", module.getModifiedChargeAttr("specialisationAsteroidYieldMultiplier"))
+    #module.multiplyItemAttr("specialtyMiningAmount", module.getModifiedChargeAttr("specialisationAsteroidYieldMultiplier"))
+    module.multiplyItemAttr("miningAmount", module.getModifiedChargeAttr("specialisationAsteroidYieldMultiplier"))


Ebag333:
Don't merge this one yet. Need to validate numbers.

blitzmann:
Made a change request to remind me that this needs to be validated. Let me know when you do so. Thanks!

Ebag333:
Using this fit:

```
[Skiff, Jenn Antollare's Skiff]
Mining Laser Upgrade II
Damage Control II
Mining Laser Upgrade II
Adaptive Invulnerability Field II
Adaptive Invulnerability Field II
EM Ward Amplifier II
Medium Shield Extender II
Survey Scanner II
Modulated Strip Miner II
Modulated Strip Miner II
Medium Core Defense Field Extender I
Medium Core Defense Field Extender I
```

With
```
[Orca, Orca]
Reinforced Bulkheads II
Damage Control II
ML-3 Scoped Survey Scanner
50MN Cold-Gas Enduring Microwarpdrive
Mining Foreman Link - Laser Optimization II
Small Tractor Beam II
Mining Foreman Link - Mining Laser Field Enhancement II
```

Here are some stats.

With links running:
>my booster only has industrial command ship 4
>(and the implant)
>From the tooltip, 1681 m3 with a t2 crystal, 961 without a crystal.
>With no crystal, 3203 pyro. 960.90m3
>with a T2 crystal, of the right sort, 5605 pyro. 1681.50m3

WIthout links running:
> tooltip says 1462m3 per cycle.
> mined quantity agrees
> that's with a crystal

Ebag333:
Okay, tweaked effects and now it works.

_*KINDA*_.

CCP basically screwed us over on this one.  So here's the way it works.  Mining lasers have two stats, one known as `miningAmount` and one known as `specialtyMiningAmount`.  We know and use the `miningAmount` stat, though at some point someone wrote an effect for `specialtyMiningAmount`.

Okay, so here's what they are.
`miningAmount` is your standard mining amount, so if you don't have crystals loaded, or are using a crystal that doesn't match the rock you're mining.
`specialtyMiningAmount` is what you'll mine if you're using crystals _and_ you're mining the correct rock.

Got that?

The _problem_ is that Pyfa _only_ uses `miningAmount`.  If we had our own mining pane, we could simply expose `specialtyMiningAmount` to show what you mine with the right crystal and right rock, but we don't have the room for that.

So, we have a few choices.

1. Overwrite `miningAmount` with the boosted `specialtyMiningAmount` (it doesn't literally overwrite it, we're just switching which one gets the crystal bonus).  This fixes the GUI, but the attributes will be off and it's really hacky.  I've already tested code that will do this, so this option is trivial.
2. Expose the mining pane on it's own and add `specialtyMiningAmount` to it.  This is probably the best option, but it'll eat up more GUI space.  *BUT WAIT!* There's more!  _Someone_ with great foresight already implemented code to allow us to selectively hide/show panes.  So it would be trivial to add that code in, so if users hate seeing that pane they can hide it.
3. Say fuck it and leave it as is.  The GUI will be wrong, but attributes will be right and users can drill down to look, not that they ever will.

blitzmann:
@Ebag333 just to confirm, is the `miningAmount` and `specialtyMiningAmount` attributes being calculated correctly, and the problem just stems from how to convey that information to the user?

Ebag333:
@blitzmann I'm 95% sure that they are yes. So yeah, how we show that is the issue at hand.

blitzmann:
Were you ever able to get confirmation form a miner about expected / actual values?

Ebag333:
Some yes, and I have access to more. ☺

I wanted to get the display decisions locked down before spending more time validating results.

blitzmann:
Try to get the results validated and confirmed. No point displaying something that is incorrect, and even if the display is not figured out they can still see the correct value in the attributes. :)

As for the display, I honestly haven't really had time to think too much about it. there's a few ifferent way to do it, but I'll bugger with it when I have time later

Ebag333:
All results I've tested have been okay so far. 

Make up your mind. 😉

blitzmann:
Merged this, I'll take a look at the gui later. I think I have an idea for a simple solution for it

Ebag333:
Wait. What? This wasn't ready to be merged....

PR #774, opened 2016-10-19
Eos Code Cleanup

Ebag333:
_LOTS_ of cleanup.

Actual code changes are _very_ minimal, only what is strictly necessary to handle validation/inspection issues.

After code cleanup, inspection for `Eos` shows:
(this excludes `eos.effects` as that is a bit of a mess)
342 warnings
1031 weak warnings

Pre-cleanup (again no `eos.effects`):
424 warnings
2038 weak warnings

.
For `eos.effects` post: 
0 warnings
2001 weak warnings

For `eos.effects` pre:
6 warnings
5592 weak warnings

Unfortunately a lot of the warnings can't be easily cleaned up.  There's a slew of shadowing warnings because of variable names we use, and virtually every effect file contains weak warnings about unused local variables because `context` is rarely used.

Some of the other warnings will be cleared up in another PR I will be submitting shortly (regarding requirements).


Ebag333:
So this is interesting.

Wanted to see if this made any different, so launched Pyfa 3 times in a row, opening the same fit (my fairly complex Vengeance that seems to take longer than most to calculate).  Since on initial launch the fit calcs 3 times, that's a total of 9 fit recalcs.

Then repeated the same exact thing on `Pyfa-org/master`.

Average fit calc time for branch: 424.11 ms
Average fit calc time for master: 542.89 ms

There are a number of minor changes that would explain some speed gain (like simplifying if statements, etc), but I'm more than a bit surprised that it's that big of a difference.  Some of it might just be the normal variance of the fitting engine (even doing the exact same thing with the exact same code won't give you the exact same times), but that seems to be a fairly big jump. vOv

blitzmann:
fuck me, 2k+ files changed. 

As a general rule, for a project like pyfa which wasn't' designed with pep8 / code compliance in mind, I don't really care about style, and changing these simply for the purposes of pep8 compliance generally annoys me as it fucks with the git blame history. I wish you would have talked to me first before starting on this endeavor.

That being said, I'm not against the changes. There's a few things that I would do differently (for example - there's a few long lines that have been broken up that I would prefer not to have been), but overall on first glance this seems like a benign PR.

Just know it will take me a while to review because yeah, 2k files  :D

Ebag333:
>fuck me, 2k+ files changed.

In for a penny, in for a pound.

>As a general rule, for a project like pyfa which wasn't' designed with pep8 / code compliance in mind, I don't really care about style, and changing these simply for the purposes of pep8 compliance generally annoys me as it fucks with the git blame history. I wish you would have talked to me first before starting on this endeavor.

Most of the PEP8 stuff I ignored, to be honest. I did fix a lot of them when they were pretty straight forward, but I had actually already turned off a bunch of the pure style validadtions becacuse  it went absolutely insane with Pyfa.

If you want me to go turn it back on and show you what the size of the change is _then_..... :)

Yeah, I know it messes with git blame, but while it touches 2k files you'll see that most files not much change.  I went through every single change and reviewed the diffs to make sure nothing obvious changed, and repeatedly ran through a series of tests as I went along to make sure that Pyfa didn't break (especially when I was touching stuff I wasn't as familiar with, like fighters).

So overall, I don't think it'll screw with git blame _too_ much.

>That being said, I'm not against the changes. There's a few things that I would do differently (for example - there's a few long lines that have been broken up that I would prefer not to have been), but overall on first glance this seems like a benign PR.

Yeah, I actually undid a lot of those because it annoys me.  Makes perfect sense to do a dict/tuple/etc over multiple lines, not so much with just a series of a validation or something where there is no clean breaking point.

I thought it was going to be pretty benign as well, but it actually _seems_ to have made a difference in speed.  I'd be curious to see if you see similar effects.

>Just know it will take me a while to review because yeah, 2k files :D

What, you had something better to do this weekend?

Ebag333:
oh, and @blitzmann it may be easier for you to go commit by commit rather than file by file.

I tried to only do one type of thing to the files, so each commit is a sort of mini-PR.

Also, I did want you to look over one commit in particular:
https://github.com/pyfa-org/Pyfa/pull/774/commits/4036ac8cd22bde8f288b85179c2aa34995572ae5

This one I started working on, with the goal of moving `projectionInfo` out of `if targetFit:` so that pyCharm understands that it is indeed created before it's referenced, but you can see where I stopped.

It works fine as is, but I feel like we should handle it a bit better.

eos/db/saveddata/fit.py
```diff
@@ -106,7 +106,7 @@ def __repr__(self):
                collection_class=HandledProjectedModList,
                cascade='all, delete, delete-orphan',
                single_parent=True,
-               primaryjoin=and_(modules_table.c.fitID == fits_table.c.ID, modules_table.c.projected == True)),
+               primaryjoin=and_(modules_table.c.fitID == fits_table.c.ID, modules_table.c.projected is True)),
```
blitzmann: Has this been tested? A previous commit reverted something like this because sql alchemy doesn't support it.

eos/capSim.py
```diff
@@ -135,13 +132,13 @@ def run(self):
...
-        cap_lowest_pre = capCapacity    # lowest cap value before activations
-        cap = capCapacity                         # current cap value
-        t_wrap = self.period                    # point in time of next period
+        cap_wrap = capCapacity  # cap value at last period
+        cap_lowest = capCapacity  # lowest cap value encountered
+        cap_lowest_pre = capCapacity  # lowest cap value before activations
+        cap = capCapacity  # current cap value
+        t_wrap = self.period  # point in time of next period
 
-        t_now = t_last = 0
```
Ebag333: So we had:

```
        t_now = t_last = 0
        t_max = self.t_max

        while 1:
            activation = pop(state)
            t_now, duration, capNeed, shot, clipSize = activation
            if t_now >= t_max:
                break
```

And it got changed to:

```
        t_last = 0
        t_max = self.t_max

        while 1:
            activation = pop(state)
            t_now, duration, capNeed, shot, clipSize = activation
            if t_now >= t_max:
                break
```

t_now is assigned when we take activation and break it down.  The creation/assignment of it when we assign `t_last` wasn't doing anything, pyCharm complained that it's a variable that's assigned but never used, and I did some quick validation.  Basically we're assigning a 0 to the var, then just blowing that assignment away when we do `t_now, duration, capNeed, shot, clipSize = activation`.

Doesn't _hurt_ to have, but not necessary.  We don't do assignments for `duration`, `capNeed`, `shot`, or `clipSize` either.

eos/db/saveddata/fit.py
blitzmann: @Ebag333 let me know if this has been tested

eos/saveddata/fit.py
```diff
@@ -461,11 +461,16 @@ def calculateModifiedAttributes(self, targetFit=None, withBoosters=False, dirtyS
         timer = Timer(u'Fit: {}, {}'.format(self.ID, self.name), logger)
         logger.debug("Starting fit calculation on: %r, withBoosters: %s", self, withBoosters)
 
+        try:
+            projectionInfo = self.getProjectionInfo(targetFit.ID)
+        except:
+            pass
+        else:
+            logger.debug("ProjectionInfo: %s", projectionInfo)
+
```
blitzmann: @Ebag333 please elaborate

Ebag333: @blitzmann Hmm, for some reason these are hidden and say "outdated".  Not sure why.  And since you didn't ping me on it, I didn't get a notification that it had been updated.  vOv

Anyway.  As I said previously

>Also, I did want you to look over one commit in particular:
>4036ac8
>
>This one I started working on, with the goal of moving projectionInfo out of if targetFit: so that pyCharm understands that it is indeed created before it's referenced, but you can see where I stopped.
>
>It works fine as is, but I feel like we should handle it a bit better.

eos/db/saveddata/fit.py
Ebag333: @blitzmann good catch. :)

Yes, it was tested.  Basically each time I modified files I tested what that commit was working on.

In this case, it _did_ fail because, yes, SQL Alchemy doesn't support it.  If you look down the commit list you'll notice I reverted it back to == for exactly the reason you mention:
https://github.com/pyfa-org/Pyfa/pull/774/commits/7cf4134a7347e1412687eeaec2a6de97da23cc5f

Ebag333:
@blitzmann updated with responses.

eos/db/saveddata/fit.py
blitzmann: @Ebag333 I must have gotten the order of the commits confused then. I thought you had fixed these, and then did this commit which reverted the fix haha

No worries with this then :)

eos/saveddata/fit.py
blitzmann: @Ebag333  Ah, I see what you're talking about now.

In this situation I would prefer to not do a large try/catch block. I would prefer to leave this bit alone, since logically it will always be assigned before being used. I agree, there's probably a better way to do this, but it would require some restructuring of the logic, which I'm not ireally interested in doing atm.

Or, you could probably set `projectionInfo = None` outside the `if` statement, and put the debug log back in the if block. That assigns the variables and prevents pyCharm from yelling whatever obscenities it yells with this kind of 'error'. I would be okay with that, and it looks cleaner. :)

Otherwise, everything else looks good, and I plan to merge this once this is solved.

Ebag333: >Or, you could probably set projectionInfo = None outside the if statement, and put the debug log back in the if block. That assigns the variables and prevents pyCharm from yelling whatever obscenities it yells with this kind of 'error'. I would be okay with that, and it looks cleaner. :)

Brilliant! 

Except I already tried this, and then pyCharm complains that it's assigned and not used, because it knows it's a bullshit assignment. ☺

I'll revert it for now, we can look at it later. It's not a major thing.

blitzmann: > pyCharm complains that it's assigned and not used, because it knows it's a bullshit assignment. ☺

=/

yeah, revert it for now. Also, I'm not sure which pyCharm warning you're getting since I turned most warnings off as it pertains to things like this. >_>

Ebag333:
@blitzmann reverted that one change so I _think_ that covers everything you've asked about so far.

>Average fit calc time for branch: 424.11 ms
>Average fit calc time for master: 542.89 ms

Did you see if there's any difference for you in how it runs speed wise?

blitzmann:
FYI, I `reset --hard` this merge and force pushed the clean master branch to github. When the issues have been fixed, please re-submit pull request. :)

PR #775, opened 2016-10-19
Setup more standard setup.py

Ebag333:
The old setup was very...special.

This is moving back toward a more standardized setup, and one with proper requirements in it to make it easier to setup environments.

Need to confirm with @blitzmann  that this doesn't break the build process.

### DIFF
--- a/requirements-build.txt
+++ b/requirements-build.txt
@@ -1,0 +1,1 @@
+setuptools==28.6.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,9 @@
 requests==2.11.1
 sqlalchemy==1.0.14
+python-dateutil=2.5.3
+matplotlib=1.5.3
+requests=2.11.1
+urllib3=1.18
 
-# Need to investigate these
+# Need to install:
 # wx
-# dateutil
-# urllib3
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,7 @@
+requests==2.11.1
+sqlalchemy==1.0.14
+
+# Need to investigate these
+# wx
+# dateutil
+# urllib3
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,11 @@
 requests==2.11.1
 sqlalchemy==1.0.14
-python-dateutil=2.5.3
-matplotlib=1.5.3
-requests=2.11.1
-urllib3=1.18
+python-dateutil==2.5.3
+matplotlib==1.5.3
+requests==2.11.1
+urllib3==1.18
+# wxPython>=2.9.11
+# wxPython-Common==2.6.3.3
 
 # Need to install:
 # wx
--- a/setup-old.py
+++ b/setup-old.py
@@ -1,0 +1,85 @@
+"""
+Distribution builder for pyfa.
+
+Windows executable: python setup.py build
+Windows executable + installer: python setup.py bdist_msi
+"""
+import requests.certs
+
+# The modules that contain the bulk of teh source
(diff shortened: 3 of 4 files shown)


blitzmann:
My question about this PR is: why? What are the benefits (besides the requirements file, which I'm okay with)?

I'm not going to merge changes to setup.py without there being a good reason. It was a bitch to get it to set up, and honestly is still a bitch to work with. This would break the build process, as now our `dist.py` script doesn't have the information it needs (which is imported via the setup script during release building). I don't think we ever discussed the current build process, have we? It seems you don't really understand how we use the current setup file (which is understandable because I almost definitely don't use it correctly lol).

pyfa utilizes it in two fashions:
1) Building binaries for Windows and OS X -- this only happens if required (see: upgrade to wxWidgets 3.0). Due to the pain of building the binaries, mostly because OS X is required to build the OS X binary, pyfa does not build them for each and every release. Instead, we generate the binaries when needed, strip out the source code files from the package that is generated, and then store the binaries in the `pyfa-skel` repository. The actual build process for release is done via the `scripts/dist.py` script, and it basically pulls the latest changes to `pyfa` repo and `pyfa-skel` repo, merges them, and zips up the result into a release file. It's a poor mans build process, but one that works fairly well. I can go into more detail about this if needed. :)
2) `dist.py` uses values in `setup.py` to select which directories are copied over to the final release packages.

The new setup.py proposed doesn't cover any other those cases, and I'm just curious as to what the new setup file does accomplish.

blitzmann:
I should stress that the things that have been added (like requirements in `setup()`) I can get behind. However, completely obliterating the file is a no go ;)

Ebag333:
>The new setup.py proposed doesn't cover any other those cases, and I'm just curious as to what the new setup file does accomplish.

The goal was to be able to get to a point where you build it like a "normal" Python package.  Or at least as normal as we can get.

I freely admit that I don't know nearly enough about python package building, but there's a number of things you mentioned (like handling building Mac vs Windows) that a standard `setup.py` _should_ be able to handle in a straight forward manner.

This isn't anywhere near complete, and it might all be screwed anyway because of wxPython.

>I should stress that the things that have been added (like requirements in setup()) I can get behind. However, completely obliterating the file is a no go ;)

I actually renamed the original to setup-old.py. :)

blitzmann:
> The goal was to be able to get to a point where you build it like a "normal" Python package. Or at least as normal as we can get.

Trust me, this _is_ as normal as I could get it. Unfortunately, in order to build pyfa binaries from source, various things have to be in alignment and you need the divine grace of the gods to do it. Granted, I am not versed much in the ways of build processes either - I basically hacked at this until it worked and then stopped for fear of it breaking again. 

Out of sheer curiosity, I just tested building a Windows binary using `setup.py build`. Shockingly, it worked, with everything seemingly working fine, so I'm not sure why the changes? (no comment on OS X builds - i fucking despise doing them. It may or may not currently work)

>I actually renamed the original to setup-old.py. :)

I saw that. What I'm saying is that it still needs to be in set up.

Ebag333:
>I saw that. What I'm saying is that it still needs to be in set up.

Mac/other you need to have a special build section with their instructions.  So yeah, that needs to be added, but the way I see to do it in the documentation isn't quite how it's been done.

Plus I don't have a Mac.  But I do have access to a few people who do......

blitzmann:
@Ebag333, I'm going to close this pull request as I feel it's unnecessary at this time to fuck with setup.py.  I don't see the value of it. However, if you want to submit a PR for the requirements.txt and include that in the `setup()` function, feel free :)

PR #783, opened 2016-10-23
Implement new method for exporting values

Ebag333:
Original method would only export values that were attached directly to the ship.  New method starts from the modified dict, which will export values without attribute IDs or original values.

### DIFF
--- a/gui/itemStats.py
+++ b/gui/itemStats.py
@@ -374,21 +374,40 @@ class ItemParams (wx.Panel):
                 ]
             )
 
-            for attributeName in self.attrInfo:
-                attribute = self.attrInfo[attributeName]
+            for attribute in self.attrValues:
 
                 try:
-                    modifiedAttributeValue = self.attrValues[attributeName].value
+                    attribute_id = self.attrInfo[attribute].ID
                 except (KeyError, AttributeError):
-                    modifiedAttributeValue = self.attrValues[attributeName]
+                    attribute_id = ''
+
+                try:
+                    attribute_name = self.attrInfo[attribute].name
+                except (KeyError, AttributeError):
+                    attribute_name = attribute
+
+                try:
+                    attribute_displayname = self.attrInfo[attribute].displayName
+                except (KeyError, AttributeError):
+                    attribute_displayname = ''
+
+                try:
+                    attribute_value = self.attrInfo[attribute].value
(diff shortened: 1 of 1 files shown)


Ebag333:
Looks mostly the same as the last PR, except:

```
ID,Internal Name,Friendly Name,Modified Value,Base Value
48,cpuOutput,CPU Output,450.0,360.0
153,warpCapacitorNeed,,4.065e-07,8.13e-07
102,turretSlotsLeft,Turret Hardpoints,5.0,5.0
1196,heatDissipationRateMed,heatDissipationRateHi,0.01,0.01
,jumpDriveConsumptionAmount,,0.0,
209,scanLadarStrength,Ladar Sensor Strength,0.0,0.0
1178,heatCapacityHi,,100.0,100.0
,energyWarfareResistance,,0.78,
,hullRepair,,0,
```

Items like `energyWarfareResistance` and `hullRepair` are now exported.

Some of these are real attributes (added via an effect most likely) and some are fake attributes (like `siege`).  They are displayed in the GUI, and the export essentially matches what you see in the GUI.  The one difference is that the GUI assumes that the base value is the modified value, when it should do a lookup against dgmattribs or leave blank.  The export does not do a lookup, and just leaves the base value as blank.

Credit to @ccplarrikin for finding this.

blitzmann:
ust gonna merge this assuming it works...

I don't like the try catches everywhere, there's probably a better way to handle that, but it's not that big of a deal

Ebag333:
Spoken like a true JS developer. :)
(You would hate C# or similar languages I think.  Try/Catch everywhere.)

I could use If/Else to see if it exists in the fit dict before grabbing it out of the modified dict, but functionally that's no different than what is done with the try/catch.

I also wanted to allow for further enhancement to properly grab the actual attrib base value (right now it's blank, because it's not attached to the dicts).  That takes a lookup back to attribs, so a bit more than the simple fix that @ccplarrikin requested (exporting the special attribs with their modified values).

blitzmann:
> You would hate C# or similar languages

I work with C# daily as part of my job. I have no problem with try/catch when needed, but there just seems like there would be a better way of doing what you're trying to do (like with `hasattr(a, 'str')` or `dict.get('str', default)`). But I can be bothered looking too far into it at the moment. .

PR #785, opened 2016-10-23
Properly Handle when Attributes not attached to directly to the Fit

Ebag333:
Currently there are some assumptions made about attributes that are assigned to the Fit via an effect or other mechanics.

We assume that the base value = current value, this is often (usually) not true.

We don't display the attribute ID.

We may or may not have the friendly name for the attribute.

If an attribute that has been modified against the Fit is not part of the Fit dict (where we're doing our lookups), this will do database lookups on the actual attribute to pull in some additional information.

We now properly populate the `base` column with the correct default value (from the attribute table).  If the lookup doesn't find the attribute (it's a fake attribute, used iternally for our purposes and not one generated by CCP) then we don't populate a base value at all.

We lookup the IDs and friendly names if possible.

This cleans up the item attribute pane quite a bit.

Adjusted the export to match the items pane improvements, they should now match each other (previously they were quite far apart for some attributes).

This actually makes it very easy to see CCP vs internal attributes, as our internal attribs won't have IDs (or base values).

### DIFF
--- a/gui/itemStats.py
+++ b/gui/itemStats.py
@@ -25,6 +25,7 @@ import sys
 import wx.lib.mixins.listctrl as listmix
 import wx.html
 from eos.types import Fit, Ship, Citadel, Module, Skill, Booster, Implant, Drone, Mode, Fighter
+import eos.db
 from gui.utils.numberFormatter import formatAmount
 import service
 import config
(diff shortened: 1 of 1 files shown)


Ebag333:
@blitzmann this is ready to merge. 😊

Ebag333:

Ebag333:
Closing as can no longer replicate original issue. (Probably fixed along the way.)

blitzmann:
@Ebag333 yeah this was merged with another PR it seems...

I'm still confused as to what the purpose of it was. Do you have a solid example of before vs after characteristics (screenshot are awesome)

Ebag333:
We weren't getting default values for a lot of things. Should have grabbed a screenshot. Oh well. 

Since we did not actually change the code here that fetched it, I suspect a lower level bug that got resolved. I rebased this, and never hit the alternative default lookups, so as far as I can tell it doesn't happen anymore.

PR #690, opened 2016-07-30
Stat Pane View Improvements

Ebag333:
We're at the point now where this is mostly functional, but could use some help with the formatting as I'm absolutely garbage with wxPython.

Plenty left to do.
- [x] Convert Resistances to the minimal version.
- [x] Convert Recharge Rates to the minimal version.
- [x] Finish converting the Full into Minimal versions (mostly format cleanup).
- [ ] Break out drones into their own separate pane.
- [ ] Break out fighters into their own separate pane.
- [ ] Break out the misc stats some? (Separate Sensor pane?)
- [x] Move "Requires Restart" line onto it's own (so it doesn't overlap the Crest one when selected)
- [ ] Icon for the Preferences window.
- [ ] Tooltip improvements on the Panes (as long as we're in there...)

http://puu.sh/qjx59/1f96e21026.png

PR #790, opened 2016-10-29
Mining Yield "per hour". 

IndictionEve:
A little improvement for our mining guys.
Miners usually calculate their mining yield per hour.

### DIFF
--- a/gui/builtinStatsViews/miningyieldViewFull.py
+++ b/gui/builtinStatsViews/miningyieldViewFull.py
@@ -124,9 +124,9 @@ class MiningYieldViewFull(StatsView):
     def refreshPanel(self, fit):
         #If we did anything intresting, we'd update our labels to reflect the new fit's stats here
 
-        stats = (("labelFullminingyieldMiner", lambda: fit.minerYield, 3, 0, 0, u"%s m\u00B3/s",None),
-                 ("labelFullminingyieldDrone", lambda: fit.droneYield, 3, 0, 0, u"%s m\u00B3/s", None),
-                 ("labelFullminingyieldTotal", lambda: fit.totalYield, 3, 0, 0, u"%s m\u00B3/s", None))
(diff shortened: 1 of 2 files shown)


Ebag333:
It's a reasonable change....except for two minor things. 

One, we have always done everything in "per seconds" so people are used to it, and SOME people hate change. 

Two, we don't handle it very well right now because we don't show what it looks like with crystals. So that's a larger issue. 

Ideally we should show it all, but gui space is an issue...

IndictionEve:
Thanks for your comment.
I made the change because I tried some mining fittings and the yield values meant nothing to me. After I multiplied the values i had a idea how "good" the yield really is.
1) "per second"  is simply not a useful (too small) scale for mining yield. Because of this, EFT calculates the yield in minutes. o.smium in hours.
It is no question, that for all other attributes "per seconds" are a great scale to compare.
2) I agree, mining crystals are a totally different topic where all tools i know have no proper solution for this.

Ebag333:
I agree that per hour stats are vastly more useful.  I think we should tie it into #771 and handle both at the same time.

The mining info pane is:
A) Hidden (which is awful UI)
B) Very limited info
C) Shows incorrect statistics (in the case of crystals)

It really needs a full rework.

We can use part of the functionality from #690 to split it out on it's own pane and let users turn it off if they don't want to see it.  Then we can show a lot more info, more accurate info, and a lot more info.

blitzmann:
I agree I like /hr better than /sec, seems more useful. I don't see anything wrong with bringing this change in immediately despite all the other issues the mining pane faces (and crystal calculations, which I still need to look at).

I'll take a look at this hopefully before the next release. :)

resinneublem:
Linking https://www.reddit.com/r/Eve/comments/5o2npc/miner_qol_improvement_m3s_calculated_inclient_on/ where people disagree that m3/h is useful.

Now that EVE has m3/s, it probably makes sense to revert this and stick with what EVE represents.

PR #791, opened 2016-10-29
open fittings in new page by default

IndictionEve:
This is a new option for a pyfa behavior disturbing me since I started to use it: I have to press CTRL to open a fitting in a new tab - and I always mess it up...
With the option in "Preferences=> General" you can change the behavior to "Open fittings in a new page by default".
Every time you select a fitting, it will open a new page.
If the option is set, the behavior of CTRL is inverted and will open the fitting in the selected tab.
The option is set false by default.

I know it is no big thing, but as soon as you miss click, restore your fitting tabs the way you want needs a lot of clicks.


Ebag333:
Overall looks good, but you need to set the default value. `settings.py` I believe (on my mobile). 

Look at the open PR for the stats UI improvements, you can see how it was added there.

IndictionEve:
Default values for this are set in `serviceFittingDefaultOptions`  in  `service/fit.py`
All other default values (Network, Export, Update, CREST) are set in `settings.py`.

blitzmann:
Yeah I don't know why general preferences are set in fit.py. I've been meaning to move them over for years. I'll take a look at this PR soon, but quick code review seems to be good

Issue #786, opened 2016-10-24
Usage abilities


Sellec:
As for usage improvement.

A very important ability is to drag cargo or booster or projected items from one setup to another. For example, i select some weapon in cargo (10 items) and want to move it into another ship. Drag&Drop ability does not allow to change active tab in setups window.

Another ability is a window that prompts an amount of items, moved into cargo. For example, it very uncomfortable to move one ammo into cargo, then make dblclick on it, select "Change AA Quantity". If this window would be showed when items is moved into cargo - it would be perfect. It can be editable in settings.

Another ability is to use MDI style except Tab style in setups windos. For example, i want to see 2-3 setups at the same time and my display is so large then it can be made very comfortably. This styles can be editable in settings.

Another ability is to open setups (by dblclick in setup list window) in new tab/window. Not, when you dblclick on any setup, you see it in currently selected tab. For example, when i compare two setups, i need to manually open new tab and then reselect them manually to compare fittings. For example, i enable this ability in settings and then dblclick on setup would show it in separate tab.

blitzmann:
> Another ability is a window that prompts an amount of items, moved into cargo. For example, it very uncomfortable to move one ammo into cargo, then make dblclick on it, select "Change AA Quantity". If this window would be showed when items is moved into cargo - it would be perfect. It can be editable in settings.

I don't see why not.

> Another ability is to use MDI style except Tab style in setups windos. For example, i want to see 2-3 setups at the same time and my display is so large then it can be made very comfortably. This styles can be editable in settings.

MDI (like EFT has) is in direct contradiction to the design philosophy of pyfa, which is to provide a unified interface with tabbed browsing. I do see the benefit of seeing multiple fits at one time, though this isn't something that can easily be achieved at this time due to the way the GUI is structured.

> A very important ability is to drag cargo or booster or projected items from one setup to another. For example, i select some weapon in cargo (10 items) and want to move it into another ship. Drag&Drop ability does not allow to change active tab in setups window.

This goes hand in hand with the MDI thing. Since you can't see multiple fittings at a time, you can drag and drop from one to another. There may be a way to ctrl+c selected items and simply copy them over; I don't know off the top of my head how difficult that would be.

[...]

PR #798, opened 2016-11-06
Ship browser tooltip

IndictionEve:
This will add trait tooltips to the shipbrowser.
-Category-View: no tooltips
-Ship-View: show trait as tooltip
-Fit-View&Search-View: show shiptype and trait as tooltip

the way how the browser handles the different items is "special" here and there but i think i finally got it.
Made a lot of crosstests to make sure the different stages are working fine.
Also, it was a request from @Sellec #786

### DIFF
--- a/gui/shipBrowser.py
+++ b/gui/shipBrowser.py
@@ -745,14 +745,18 @@ class ShipBrowser(wx.Panel):
             filter = subRacesFilter[ship.race] if ship.race else True
             if override:
                 filter = True
-
+            if ship.traits is not None:
+                shipTrait = ship.traits.traitText
+            else:
+                shipTrait = ""             
+                
             if self.filterShipsWithNoFits:
                 if fits>0:
                     if filter:                       
-                        self.lpane.AddWidget(ShipItem(self.lpane, ship.ID, (ship.name, ship.traits.traitText, fits), ship.race))
+                        self.lpane.AddWidget(ShipItem(self.lpane, ship.ID, (ship.name, shipTrait, fits), ship.race))
             else:
                 if filter:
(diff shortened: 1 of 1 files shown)


IndictionEve:
`Bugfix: Unable to handle ships without traits (Citadels) ba47420` 
The traits of citadels are just text inside the description. I had to modify the code to handle ships without traits.
The code is not pretty, but I want to keep the parameter variable simple (string) 
Plan B is to send the whole traits structure as parameter to the function. This would result in nicer code but an ugly structure (`Traits.traitText`) as variable.

Ebag333:
Overall looks good. 

One question though.

If I hover over the ship itself (not a ship fit), it shows just the bonus text.  If I hover over a fit, it shows the ship name, a bunch of `-` in line, then the bonus text.

Is there a reason for showing the ship name and the dashed line for fits?  Both should display the same and I think that the dashed line doesn't show very well.

IndictionEve:
The dash line is just a seperator.
Formating default tooltips is simply not possible (as far as i know)

In "ship view" you know the ship and just interested in traits.
In "fitting view" the shiptype is indicated just by an image. 
The displayed "name" is the name of the fitting.
Especially when using search (search for "pvp" for example) you are getting may different fits of different shiptypes and it is nearly not possible to seperate a "pvp" caracal from a "pvp" cerberus or a "pvp" caracal navy issue by image.
Because of this I added the shiptype to the tooltip to make this information more visible.

blitzmann:
This is great, thanks so much for the contribution!

One thing: I also didn't like the hyphens used for the separator, so instead I changed them to unicode dashes. These should look seemless, at least on Windows machines (haven't tried Linux / OSX)

Issue #755, opened 2016-09-29
UNLIMITED POWAAAAAAAAAR!!!!! missing calculations


Rylleyryelle:
ebag helped me get pyfa running because hes trying to get me off of eft for some reason.. but so far im going to have to stick to that because this just isnt working. it opens now, but its not calculating basic things.

**1-** Fighters
**1. A** Fighter DPS Not Calculated:
i put in my fighters in my carrier and it still says 0.0 DPS. i click the fighters tab and it says 3/3 light fighters, they all 3 have checkmarks, but nothing. 0 dps.
**1.B** Fighter Abilities Missing:
on top of the fighters just not applying dps at all to the display, their abilities are missing. it just says "Active Abilities -- None" for every single fighter
**1.C** Fighter Control Range:
when using fighters, pyfa seems to think the standard 60km base range is correct, this should be corrected to 400km when using fighters
**2-** Pyfa still lets you fit subcap drones into carriers (example Warrior II into an Archon)

blitzmann:
Hey there. Let me try to clear a few of these up for you :)

> 1. A Fighter DPS Not Calculated:
i put in my fighters in my carrier and it still says 0.0 DPS. i click the fighters tab and it says 3/3 light fighters, they all 3 have checkmarks, but nothing. 0 dps.

>1.B Fighter Abilities Missing:
on top of the fighters just not applying dps at all to the display, their abilities are missing. it just says "Active Abilities -- None" for every single fighter

There isn't any DPS because you probably didn't set any abilities. To do so, right click the fighter group and select the abilities to apply. Unlike other drones and module, pyfa doesn't activate anything by default for fighters, since there are multiple things to choose from. I could put a note stating `(right click for abilities)` to help with the confusion, that should be easy enough. :)

> 1.C Fighter Control Range:
when using fighters, pyfa seems to think the standard 60km base range is correct, this should be corrected to 400km when using fighters

This is a valid concern. The 60km range is dictated directly from one of the drone skills IIRC. I must have missed that fighters have their own (is it always 400?). Will check. 

> 2- Pyfa still lets you fit subcap drones into carriers (example Warrior II into an Archon)

This is true for any fit, and it's a design thing. We don't restrict any ship from fitting regular drones. You can put 30 Warrior IIs in a Retribution is you like. However, you will notice that the drone bay is over capacity for any ship that doesn't allow for it. 

[...]

PR #799, opened 2016-11-06
Activate fighter standard attack and limit active fighters

IndictionEve:
#755 @Rylleyryelle said: 
> i as a user, would much prefer to have to DISABLE what i dont want running on them, rather than having to enable every stupid bleepity bleeping thing every time i change out a fighter to tinker.
- Adding Fighters: Activate all implemented abilities by default
-- Because you don't add deactivated modules to fittings, too
-- The usability of fighters is not optimal, so let's make it a little bit easier for the daily use.
- Adding Fighters: Deactivate fighter groups if more than the maximum number of active fighters are in bay (soft limit)
--It is still possible to activate more fighter then allowed for testing, switching fighters, etc...

### DIFF
--- a/service/fit.py
+++ b/service/fit.py
@@ -658,6 +658,13 @@ class Fit(object):
             '''
             if fighter is None:
                 fighter = eos.types.Fighter(item)
+                used = fit.getSlotsUsed(fighter.slot)
+                total = fit.getNumSlots(fighter.slot)
+                for ability in fighter.abilities:
+                    ability.active = True if ability.effect.isImplemented else False
+                if used >= total:
+                    fighter.active = False
+
                 if fighter.fits(fit) is True:
                     fit.fighters.append(fighter)
                 else:
--- a/service/fit.py
+++ b/service/fit.py
@@ -660,8 +660,20 @@ class Fit(object):
                 fighter = eos.types.Fighter(item)
                 used = fit.getSlotsUsed(fighter.slot)
                 total = fit.getNumSlots(fighter.slot)
+                standardAttackActive = False;
                 for ability in fighter.abilities:
-                    ability.active = True if ability.effect.isImplemented else False
+                    if (ability.effect.isImplemented and ability.effect.handlerName == u'fighterabilityattackm'):
+                        # Activate "standard attack" if available
+                        ability.active = True
+                        standardAttackActive = True
+                    else:
+                        # Activate all other abilities (Neut, Web, etc) except propmods if no standard attack is active
+                        if (ability.effect.isImplemented
+                            and standardAttackActive == False
+                            and ability.effect.handlerName != u'fighterabilitymicrowarpdrive'
+                            and ability.effect.handlerName != u'fighterabilityevasivemaneuvers'):
+                            ability.active = True
+
                 if used >= total:
                     fighter.active = False
 


Ebag333:
If I import a carrier with 1 million fighters on them, will they all  be activated? :)

IndictionEve:
Nope, nothing changes to the current behavior.
You have to add a fighter in pyfa for the new mechanics.

IndictionEve:
Thanks for the import tip, seems like we have a issue over here #800

Ebag333:
Seems like it.

My only concern was that a number of carriers can have a _lot_ of spare fighters, it'd be more painful to turn them off manually than turn the (relatively) few that it has tubes for on.

But sounds like that's not an issue.

blitzmann:
Firstly, I am still not sold on activating them by default. It would be best to activate the 'standard' attack by default if there's an easy way to get to that programatically (would probably have to hard code effects which we want to auto-activate). So, instead of 

```
for ability in fighter.abilities:
                    ability.active = True if ability.effect.isImplemented else False
```

it would be 

```
for ability in fighter.abilities:
                    ability.active = True if ability.effect.isImplemented and ability.effect.name in ("effect1", "effect2") else False
```

(or however you get the name of the effect, can't remember off the top of my head)

I am open to counter points on this though. :)

Secondly, I do like the soft limit you've placed when adding fighters. Nice touch, and definitely want to keep that regardless of what we do with the fighter abilities.

IndictionEve:
Basically it's about why do you want to add fighters to your fitting: 
First of all you want to know the max values (dps, speed)
Adding fighters without abilities is equal to a inactive fighter group: they are there but have no impact on my fitting. But this is the reason you using this tool: to simulate the effects of modules (respectively fighters).
At the second step you start experimenting with the attributes (damage without rockets, speed without propmod, etc.)
In the end you got a feeling for the maximum rating and everything between. 

However your idea just activate the 'standard attack' looks like the best solution to me.
I asked 15 guys out of my alliance what kind of behavior they expect if they add fighter to a fitting, without telling them any options: 
11 of them expecting an active 'standard attack'  
4 of them 'everything activated'
'nothing activated ' was no option for them.
But hard code abilities is something I don't like, because you have to look after the changes  every update to make sure all standard attacks are set as wanted.

Ebag333:
>Basically it's about why do you want to add fighters to your fitting: 
>First of all you want to know the max values (dps, speed)

I agree with this.

I'd love to see a a "standard attack active" option, but since there's no default we're stuck with hard coding something, which is fugly.

Basically I'd put it in this order:
1. Standard attack on, special off
2. All attacks on
3. All attacks off

My only concern is that we make sure we don't enable all the fighters (so you don't have to click ten million times to turn *OFF* fighter attacks), the soft limit should take care of that (I hope).

If we have to hard code a list of attributes, I'd rather just turn on all the abilities than hard coding that list, since it's just another thing to maintain.

So yeah, pretty much what @IndictionEve said.

IndictionEve:
Looks like I've got it.
The indicator we looking for is `ability.effect.handlerName`.
`fighterabilityattackm` is the "standard attack" for light and heavy drones.

support fighter currently have two abilities - propmod and utility (neut, web, etc)
If there is no "standard attack" all other abilities except propmods (`fighterabilitymicrowarpdrive` and `fighterabilityevasivemaneuvers`) set active.

because of this pyfa is able to handle and activate future effects (maybe mining fighters?)

blitzmann:
I merged this in singularity branch. Excellent work.

@Ebag333's work in #804 might helping moving forward with the default effects to activate and whatnot once that's merged in. Would probably take a bit of work, and I'm not worried about it right now, but it's a thing.

PR #801, opened 2016-11-07
Handle Lockbreaker, Void, and Focused Void Bombs

Ebag333:
Bomb effects can now be projected.

For calculations that include a cycle time (cap warfare bombs), the cycle time assumes that you launch a bomb as soon as the cooldown is complete (`moduleReactivationDelay`).

### DIFF
--- a/eos/effects/usemissiles.py
+++ b/eos/effects/usemissiles.py
@@ -12,7 +12,7 @@ def handler(fit, src, context):
     src.reloadTime = 10000
 
     if "projected" in context:
-        if src.item.group.name == unicode("Missile Launcher Bomb"):
+        if src.item.group.name == u'Missile Launcher Bomb':
             # Bomb Launcher Cooldown Timer
             moduleReactivationDelay = src.getModifiedItemAttr("moduleReactivationDelay")
 
--- a/eos/effects/usemissiles.py
+++ b/eos/effects/usemissiles.py
@@ -4,7 +4,27 @@
 # Modules from group: Missile Launcher Heavy (12 of 12)
 # Modules from group: Missile Launcher Rocket (15 of 15)
 # Modules named like: Launcher (151 of 151)
-type = 'active'
-def handler(fit, module, context):
+type = 'active', "projected"
+
+
+def handler(fit, src, context):
     # Set reload time to 10 seconds
-    module.reloadTime = 10000
+    src.reloadTime = 10000
+
+    if "projected" in context:
+        if src.item.group.name == unicode("Missile Launcher Bomb"):
+            # Bomb Launcher Cooldown Timer
+            moduleReactivationDelay = src.getModifiedItemAttr("moduleReactivationDelay")
+
+            # Void and Focused Void Bombs
+            neutAmount = src.getModifiedChargeAttr("energyNeutralizerAmount")
+
+            if moduleReactivationDelay and neutAmount:
+                fit.addDrain(src, moduleReactivationDelay, neutAmount, 0)
+
+            # Lockbreaker Bombs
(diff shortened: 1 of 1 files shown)


eos/effects/usemissiles.py
```diff
@@ -4,7 +4,27 @@
...
+type = 'active', "projected"
+
+
+def handler(fit, src, context):
     # Set reload time to 10 seconds
-    module.reloadTime = 10000
+    src.reloadTime = 10000
+
+    if "projected" in context:
+        if src.item.group.name == unicode("Missile Launcher Bomb"):
```
blitzmann: Just curious, why `unicode()`? Everywhere else uses `u"string"` instead, but even here we don't really need it as we know it's going to be an ASCII string

Ebag333: `src.item.group.name` isn't ASCII.

https://puu.sh/s9PbQ/3d74247d1b.png

I could have used `u"Missile Launcher Bomb"` instead I suppose, but potato tomato.

blitzmann: It's stored as unicode, yes, but it isn't unicode. Everything in that string could be ASCII.

My main point is that it's inconsistent (not that pyfa is consistent with much). In most (all?) other effect files we use a simple strings when matching groups, skills, etc, because there isn't any reason to use unicode. And if we are gonna use unicode, `u""` convention would be preferred.

Ebag333:
>Also, why is this being done in useMissiles effect? The bombs themselves have an effect we can utilize I believe

I was originally going to use `bombLaunching` but there's a few problems with it.

1) Charges don't seem to trigger a `projected` context.
2) When they are triggered, the `fit` is the host ship.

We could get around 2 by getting the `projected onto` ship, but that's fugly.  1, not so much.

eos/effects/usemissiles.py
Ebag333: >It's stored as unicode, yes, but it isn't unicode. Everything in that string could be ASCII.

`if src.item.group.name == "Missile Launcher Bomb":`

Doesn't match.

I'll change it to `u'` format.

blitzmann:
> We could get around 2 by getting the projected onto ship, but that's fugly. 1, not so much.

I'd have to take a look at the method used to calculate the attributes for modules and how it's used with charges, but this might be pretty easy to fix. I would prefer this way as it's the convention (it's the charge doing the modifying, not the module like I'm pretty sure this conveys)

eos/effects/usemissiles.py
blitzmann: >`if src.item.group.name == "Missile Launcher Bomb":`
>Doesn't match.

wait... really? Well that changes things. How the fuck does that not work? The lambda shouldn't matter AFAIK, as it just declares an anonymous function...

Ebag333:
>I'd have to take a look at the method used to calculate the attributes for modules and how it's used with charges, but this might be pretty easy to fix. I would prefer this way as it's the convention (it's the charge doing the modifying, not the module like I'm pretty sure this conveys)

Well, I don't know about convention but apparently doing it this way is common enough that someone wrote at `getModifiedChargeAttr` function. :)

We need to know a lot of things from the projecting fit, and things off both the module and charges.  So even if we went off the charge effect, we'd have to find the parent module, and get modified attributes off that.

It's a lot more complicated, and would require some changes to the way that we handle projection (to allow projecting charges and not just modules/ships).

It's also not really the charge that's doing the modifying, but the module itself.  It's not too different than how remote sebos work really.

blitzmann:
I still think we should handle projecting charges, but the GUI doesn't accept anything but modules, and it seems like more hassle then it's worth. This is fine for now. Merged.

PR #804, opened 2016-11-09
Make Effects Togglable

Ebag333:
This does a few things.

1. "Implemented" got renamed to Active, and instead of a 2 value (Yes/No) option, it's a 3 value option (Yes, No, Blank).  Yes and No cover the old `isImplemented` Yes option (and represent if the effect is currently being calculated or not), while blank represents the old `isImplemented` No option.
2. Added a Type column.  This displays the types, and is mostly for debugging, but super useful if someone wants to toggle on drug side effects.
3. Added a run time column.  This is only shown when debug mode is on (so only shows when the ID column does).  Super useful for quickly seeing what stage an effect ran under.
4. Resized columns to better fit the various information they show.
5. Moved the debug only double left click to a double right click.  (Sorry @blitzmann !)
6. You can double left click on any effect to toggle it on/off (switch Active state from True to False and vice-versa).  If the effect isn't implemented, nothing happens.  (Way to bury the lede.)
7. Commented out the old booster side effect code.  This is maybe 50% implemented, and hasn't been touched in going on 4 years now.  Commenting out this code will speed up the execution of Pyfa a bit, and if someone down the line ever implements it, it can easily be put back.

This is implemented by introducing a new state called `activeByDefault`.  This is attached to the effect directly, which has some interesting side effects (more on that in a moment, also, pun intended).  So you can find it under (few quick examples):
```
fit.ship.item.effects[].activeByDefault
fit.modules[].item.effects[].activeByDefault
```

You can get the attribute (and also set the attribute) directly against *.effects[].

To use, simple add to an effect file (above `def handler`):
`activeByDefault = False`

This can be added to _any_ effect, though currently aside from booster side effects I can't think of any that we would want to add this to.

If this is not set:
We simply assume that missing/none = True, and set it accordingly (much as we set `runTime` to `Normal `if not otherwise set).  Nearly all effect files will fall under this category.

If this is set to `True`:
We would enable it anyway, but hey, it's double enabled.  No effect files are currently configured this way (and probably will never be).

If this is set to `False`:
Basically we simply skip adding the effect to the effect handler when the effect is called, much as if the run time didn't match or other criteria failed.

Okay, now for the caveats.

1. Toggling this doesn't trigger a refresh the fit calcs.  Not sure how to do this, maybe @blitzmann can chime in.  So once you toggle the effects, you need to modify the fit and trigger a refresh (simply turning a module off then back on works).
2. This _*is not*_ stateful between sessions.  If you close and reopen Pyfa, it will use the default values. (No plans to make this stateful between sessions as it's primarily a debugging tool.)
3. This _*IS*_ stateful within a session.  This is because it's tied to a _effect_, not a particular fit.  So if I turn on side effects for a drug, then _all_ fits with that same drug (or at least the same effects) will have those side effects (once recalculated).

And finally:
How to toggle effects

### DIFF
--- a/eos/gamedata.py
+++ b/eos/gamedata.py
@@ -85,15 +85,25 @@ class Effect(EqBase):
 
     @property
     def activeByDefault(self):
-        '''
-        The runTime that this effect should be run at.
+        """
+        The state that this effect should be be in.
         This property is also automaticly fetched from effects/<effectName>.py if the file exists.
         the possible values are:
         None, True, False
-        None and False are equivalent. True is the default if the effect is also implemented.
 
-        effects that are not active will not be calculated.
-        '''
+        If this is not set:
+        We simply assume that missing/none = True, and set it accordingly
+        (much as we set runTime to Normalif not otherwise set).
+        Nearly all effect files will fall under this category.
+
+        If this is set to True:
+        We would enable it anyway, but hey, it's double enabled.
+        No effect files are currently configured this way (and probably will never be).
+
+        If this is set to False:
+        Basically we simply skip adding the effect to the effect handler when the effect is called,
+        much as if the run time didn't match or other criteria failed.
+        """
         if not self.__generated:
             self.__generateHandler()
 
@@ -101,8 +111,10 @@ class Effect(EqBase):
 
     @activeByDefault.setter
     def activeByDefault(self, value):
-        # Just assign the input values to the ``numbers`` attribute.
-        # You *could* do something more interesting here if you wanted.
+        """
+        Just assign the input values to the activeByDefault attribute.
+        You *could* do something more interesting here if you wanted.
+        """
         self.__activeByDefault = value
 
     @property
--- a/eos/gamedata.py
+++ b/eos/gamedata.py
@@ -84,6 +84,28 @@ class Effect(EqBase):
         return self.__runTime
 
     @property
+    def activeByDefault(self):
+        '''
+        The runTime that this effect should be run at.
+        This property is also automaticly fetched from effects/<effectName>.py if the file exists.
+        the possible values are:
+        None, True, False
+        None and False are equivalent. True is the default if the effect is also implemented.
+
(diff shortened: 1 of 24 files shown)


eos/gamedata.py
```diff
@@ -84,6 +84,28 @@ def runTime(self):
         return self.__runTime
 
     @property
+    def activeByDefault(self):
+        '''
+        The runTime that this effect should be run at.
+        This property is also automaticly fetched from effects/<effectName>.py if the file exists.
```
blitzmann: Someone forgot to change the docstring :)

Ebag333: Fixed. :)

Ebag333:
>Did we really have all the booster side effect effects created? 

Yup.

>Wow, had no idea. So really, the booster side effect stuff is really like 80% done.

Nope.

The effects are the easiest part, and remember that I touched every side effect that was already created and updated them, so there's only like 20ish of them.

The side effect stuff is basically a table, some structure around the object (for storing and linking it), and that's it. 

Here's what I think is left to be done:
1. There's nothing to toggle them on/off functionally, so all those functions need to be built out.
2. The table in the DB probably needs to be extended a bit, you would have to add/drop rows rather than simply changing a value to cache whether the side effect is on/off.  As we have seen elsewhere, constantly adding/dropping rows is a good way to introduce corruption.
3. Of course none of the GUI stuff is done at all.

.....basically there is still a boatload of work to do on it.  I'd still say it's 50% complete, and that might be a generous estimate.

Ebag333:
>The effects are the easiest part

Speaking of which, I should go through and make sure that all the effects are implemented, and that they're actually correct.

blitzmann:
@Ebag333 could you possibly get this updated against sisi branch? If so, I can try to get it tested and merged with the pre-release tomorrow. If not, then we can get it merged later, not a big deal. :)

Ebag333:
Closing as it's been replaced with #832

Issue #803, opened 2016-11-08
Add to projected tab via right click menu


Ebag333:
Dragging and dropping ship fits to the projection window is super painful, especially on touchpads.  Often it ends up opening the fit, which means it takes about 5 clicks to get back to where you were to try again.  Adding it to the right click menu should be easy, and would help a lot with that minor pain point.

PR #810, opened 2016-11-13
Add "projected" to selected fitting via shipbrowser right click menu

IndictionEve:
Merry Christmas @Ebag333 :)
Solve #803

New item in shipbrowser context menu `Set as 'Projected' to selected fitting`


blitzmann:
Few things here:

1) Shouldn't show unless there is an active fit
2) When projecting, should switch to the projection tab.

Despite this, I'm going to merge it since I need similar functionality for new command boosts as well. Thanks!

PR #811, opened 2016-11-14
Requirements for Pyfa

Ebag333:
Caught most of the main ones, I think.
(Yes I'm sneaking EVE-Gnosis in here, just so I don't have a conflict when I submit the cap/shield sim fix)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,4 +4,3 @@ python-dateutil
 urllib3
 requests==2.11.1
 sqlalchemy
-EVE-Gnosis>=2016.11.17post2
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,7 @@
+matplotlib
+PyYAML
+python-dateutil
+urllib3
+requests
+sqlalchemy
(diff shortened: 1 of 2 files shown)


resinneublem:
Oh nice. I'd dig something like this too!

blitzmann:
This will be tied in when #812 gets merged

Ebag333:
This should be able to be merged before 812 because we don't reference the `requirements.txt` anywhere in `setup.py`.  Only IDEs like pyCharm will pick it up and warn you about missing prereqs.

(At least, in a normal build scenario.  For Pyfa's weird build this may not hold true.)

Ebag333:
@blitzmann removed EVE-Gnosis as that's going to have to have other PRs merged, so not quite as quick as I'd hoped....

blitzmann:
What is requirements_test?

Ebag333:
>Add requirements for getting this to work on Travis-Ci

Basically if you are running tests against it outside of an IDE that will handle it for you (like pyCharm).

PR #816, opened 2016-11-15
Limit lanaguage from multi to en-us

Ebag333:
Lets this run in 32 bit python.

### DIFF
--- a/eos/saveddata/fit.py
+++ b/eos/saveddata/fit.py
@@ -32,7 +32,7 @@ import time
 import copy
 from utils.timer import Timer
 from eos.enum import Enum
-
+from gnosis.formulas.formulas import Formulas
 
 import logging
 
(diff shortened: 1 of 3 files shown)


blitzmann:
1) Seems there's other commits in this that shouldn't be in there. Plz fix :)
2) I'm okay with this in general for now. There are other changes that need to be made though, I think in the json2sql script.

Ebag333:
Ugh, no idea how that got in there. 

These changes were all that was needed to make a build work without error.

Ebag333:
>1) Seems there's other commits in this that shouldn't be in there (new cap stuff). Plz fix :)

Fixed.

blitzmann:
> These changes were all that was needed to make a build work without error.

Could have sworn this was going to break it https://github.com/pyfa-org/Pyfa/blob/master/scripts/jsonToSql.py#L68

But you're right, it does work. :D

PR #821, opened 2016-11-20
Adds a new notes pane for each fit

### LINKED ISSUE #364: Suggestion: Offer a description/notes field for fits
Just something small to allow people to write some notes down about the fit

resinneublem:
This allows users to save text notes for each fit. They may do this for history, or usage, etc.

Closes #364

### DIFF
--- a/eos/db/migrations/upgrade16.py
+++ b/eos/db/migrations/upgrade16.py
@@ -1,0 +1,14 @@
+"""
+Migration 16
+
+- Alters fits table to introduce notes attribute
+"""
+
+import sqlalchemy
+
+def upgrade(saveddata_engine):
+    # Update fits schema to include notes attribute
+    try:
+        saveddata_engine.execute("SELECT notes FROM fits LIMIT 1")
+    except sqlalchemy.exc.DatabaseError:
+        saveddata_engine.execute("ALTER TABLE fits ADD COLUMN notes VARCHAR;")
--- a/gui/additionsPane.py
+++ b/gui/additionsPane.py
@@ -25,6 +25,7 @@ from gui.fighterView import FighterView
 from gui.cargoView import CargoView
 from gui.implantView import ImplantView
 from gui.projectedView import ProjectedView
+from gui.notesView import NotesView
 from gui.pyfatogglepanel import TogglePanel
 from gui.gangView import GangView
 from gui.bitmapLoader import BitmapLoader
@@ -57,6 +58,7 @@ class AdditionsPane(TogglePanel):
         projectedImg = BitmapLoader.getImage("projected_small", "gui")
         gangImg = BitmapLoader.getImage("fleet_fc_small", "gui")
         cargoImg = BitmapLoader.getImage("cargo_small", "gui")
+        notesImg = BitmapLoader.getImage("skill_small", "gui")
 
         self.drone = DroneView(self.notebook)
         self.notebook.AddPage(self.drone, "Drones", tabImage = droneImg, showClose = False)
@@ -79,9 +81,12 @@ class AdditionsPane(TogglePanel):
         self.gangPage = GangView(self.notebook)
         self.notebook.AddPage(self.gangPage, "Fleet", tabImage = gangImg, showClose = False)
 
+        self.notes = NotesView(self.notebook)
+        self.notebook.AddPage(self.notes, "Notes", tabImage = notesImg, showClose = False)
+
         self.notebook.SetSelection(0)
 
-    PANES = ["Drones", "Fighters", "Cargo", "Implants", "Boosters", "Projected", "Fleet"]
+    PANES = ["Drones", "Fighters", "Cargo", "Implants", "Boosters", "Projected", "Fleet", "Notes"]
     def select(self, name):
         self.notebook.SetSelection(self.PANES.index(name))
 
--- a/gui/notesView.py
+++ b/gui/notesView.py
@@ -1,0 +1,45 @@
+import wx
+
+import service
+import gui.globalEvents as GE
+import gui.mainFrame
+
+class NotesView(wx.Panel):
+    def __init__(self, parent):
+        wx.Panel.__init__(self, parent)
+        self.lastFitId = None
+        self.mainFrame = gui.mainFrame.MainFrame.getInstance()
+        mainSizer = wx.BoxSizer(wx.VERTICAL)
+        self.editNotes = wx.TextCtrl(self, style=wx.TE_MULTILINE)
+        mainSizer.Add(self.editNotes, 1, wx.EXPAND)
+        self.SetSizer(mainSizer)
+        self.mainFrame.Bind(GE.FIT_CHANGED, self.fitChanged)
+        self.Bind(wx.EVT_TEXT, self.onText)
+        self.saveTimer = wx.Timer(self)
+        self.Bind(wx.EVT_TIMER, self.delayedSave, self.saveTimer)
+
+    def fitChanged(self, event):
+        sFit = service.Fit.getInstance()
+        fit = sFit.getFit(event.fitID)
+
+        self.Parent.Parent.DisablePage(self, not fit or fit.isStructure)
+
+        if event.fitID is None and self.lastFitId is not None:
+            self.lastFitId = None
+            event.Skip()
+            return
+        elif event.fitID != self.lastFitId:
+            self.lastFitId = event.fitID
+            self.editNotes.SetValue(fit.notes or "")
+
+    def onText(self, event):
+        # delay the save so we're not writing to sqlite on every keystroke
+        self.saveTimer.Stop() # cancel the existing timer
+        self.saveTimer.Start(1000, True)
+
+    def delayedSave(self, event):
+        sFit = service.Fit.getInstance()
+        fit = sFit.getFit(self.lastFitId)
+        newNotes = self.editNotes.GetValue()
+        fit.notes = newNotes
+        wx.PostEvent(self.mainFrame, GE.FitChanged(fitID=fit.ID))
--- a/eos/db/saveddata/fit.py
+++ b/eos/db/saveddata/fit.py
@@ -45,6 +45,7 @@ fits_table = Table("fits", saveddata_meta,
                    Column("targetResistsID", ForeignKey("targetResists.ID"), nullable=True),
                    Column("modeID", Integer, nullable=True),
                    Column("implantLocation", Integer, nullable=False, default=ImplantLocation.FIT),
+                   Column("notes", String, nullable = True),
                    )
 
 projectedFits_table = Table("projectedFits", saveddata_meta,
--- a/eos/saveddata/fit.py
+++ b/eos/saveddata/fit.py
@@ -1085,6 +1085,7 @@ class Fit(object):
         copy.name = "%s copy" % self.name
         copy.damagePattern = self.damagePattern
         copy.targetResists = self.targetResists
+        copy.notes = self.notes
 
         toCopy = (
             "modules",


eos/db/migrations/upgrade16.py
```diff
@@ -0,0 +1,14 @@
...
+
+- Alters fits table to introduce notes attribute
+"""
+
+import sqlalchemy
+
+def upgrade(saveddata_engine):
+    # Update fits schema to include notes attribute
+    try:
+        saveddata_engine.execute("SELECT notes FROM fits LIMIT 1")
```
resinneublem: I ripped this off from https://github.com/pyfa-org/Pyfa/blob/master/eos/db/migrations/upgrade13.py

gui/additionsPane.py
```diff
@@ -57,6 +58,7 @@ def __init__(self, parent):
         projectedImg = BitmapLoader.getImage("projected_small", "gui")
         gangImg = BitmapLoader.getImage("fleet_fc_small", "gui")
         cargoImg = BitmapLoader.getImage("cargo_small", "gui")
+        notesImg = BitmapLoader.getImage("skill_small", "gui")
```
resinneublem: This isn't ideal, but I just put this in as a placeholder. I don't know if you all know of a pen or a paper icon we could use for this.

gui/notesView.py
```diff
@@ -0,0 +1,41 @@
+import wx
+
+import service
+import gui.globalEvents as GE
+import gui.display as d
+import gui.mainFrame
+
+class NotesView(d.Display):
+    DEFAULT_COLS = []
```
resinneublem: I tried excluding this as well as setting this to `None` and neither seemed to work

gui/notesView.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __init__(self, parent):
```
resinneublem: I am very new to wxPython so I could use some review on this area.

gui/additionsPane.py
```diff
@@ -79,9 +81,12 @@ def __init__(self, parent):
         self.gangPage = GangView(self.notebook)
         self.notebook.AddPage(self.gangPage, "Fleet", tabImage = gangImg, showClose = False)
 
+        self.notes = NotesView(self.notebook)
+        self.notebook.AddPage(self.notes, "Notes", tabImage = notesImg, showClose = False)
+
         self.notebook.SetSelection(0)
 
-    PANES = ["Drones", "Fighters", "Cargo", "Implants", "Boosters", "Projected", "Fleet"]
+    PANES = ["Drones", "Fighters", "Cargo", "Implants", "Boosters", "Projected", "Fleet", "Notes"]
```
resinneublem: I've noticed one small bug and I can't figure out how to fix it.

If your preferences are set to reload previously open fits, when this opens the fit, it goes to the "Fleet" tab. I'm really not sure why or what code is setting that.

gui/additionsPane.py
Ebag333: Take a look in `\imgs\icons`.  That's what we have currently.

Not sure we have anything better.

resinneublem: Bummer. Not seeing anything.
 
I recognize some of the icons in gui, I just can't remember the name of the set to search and grab an editor looking icon.

gui/notesView.py
Ebag333: IIRC, every view expects at least one column.  Leaving this blank should be fine.

gui/additionsPane.py
Ebag333: So this is a minor quibble, but it's a bit of a pet peeve of mine.  Because git blame goes line by line, if anyone ever adds/removed a single field from this, all the blame goes to that person.

Where I've come across this I've been changing it to:
```
PANES = [
    "Drones",
    "Fighters",
    "Cargo",
    "Implants",
    "Boosters",
    "Projected",
    "Fleet",
    "Notes",
]
```

You'll get git blamed for every line (but you were going to be anyway), and in the future adding a new tab won't assign blame for all tabs.  It also makes comparisons easier down the road.

eos/db/migrations/upgrade16.py
```diff
@@ -0,0 +1,14 @@
...
+    except sqlalchemy.exc.DatabaseError:
+        saveddata_engine.execute("ALTER TABLE fits ADD COLUMN notes STRING;")
```
Ebag333: `STRING` isn't exactly a data type in SQLite.

Historically we've been using `VARCHAR` (which just causes it to use `TEXT`).  You probably want to use `TEXT`, though `BLOB` might be an option as well.

https://www.sqlite.org/datatype3.html

I would say use either `VARCHAR` or `TEXT`.

Ebag333:
So the bigger problem with this is that every single time a change is made, this is written to the database.  This means typing on this is _very_ slow.

Have you considered an apply button to save changes, so it's not trying to save it in real time?

gui/additionsPane.py
resinneublem: It's quite easy to view previous blames by using git references. I do it all the time: https://github.com/resinneublem/Pyfa/blame/1ae509e189d455369eef39327c712e0fe81b5e91%5E/gui/additionsPane.py

`1ae509e189d455369eef39327c712e0fe81b5e91^` is the key

gui/notesView.py
```diff
@@ -0,0 +1,41 @@
...
+        self.mainFrame.Bind(GE.FIT_CHANGED, self.fitChanged)
+        self.Bind(wx.EVT_TEXT, self.onText)
+
+    def fitChanged(self, event):
+        sFit = service.Fit.getInstance()
+        fit = sFit.getFit(event.fitID)
+
+        self.Parent.Parent.DisablePage(self, not fit or fit.isStructure)
+
+        if event.fitID is None and self.lastFitId is not None:
```
resinneublem: This was cargo culted from the other additional panels. fitChanged fires a lot.

It's quite readable:

* If the new fit is none, and the last fit was set, set last fit to none and skip the remainder of the event
* If the new fit is different than the last known fit, set the text value

I started by using `currentFit` instead of storing the `lastFitId` but I changed it to stay consistent.

gui/additionsPane.py
resinneublem: Why use `big` here when the other icons use `small`?

resinneublem:
> So the bigger problem with this is that every single time a change is made, this is written to the database. This means typing on this is very slow.
>
> Have you considered an apply button to save changes, so it's not trying to save it in real time?

Hrm yeah, I can see some input lag while typing in it (goes away when I disable the save code). I'll give that a shot.

gui/additionsPane.py
blitzmann: I like the current skillbook icon. But yeah, if you find another icon from the EVE image export, simply skrink it to 16x16 and include it. But as it stands now, looks good

gui/additionsPane.py
blitzmann: This doesn't happen to me... but I'm not too concerned about it.

gui/additionsPane.py
Ebag333: You wouldn't :)

This is fine, I didn't notice which icon you'd used and didn't check the other tabs to see the size.

gui/notesView.py
blitzmann: The problem here is that you're Inheriting `d.Display`. That is what we use for our lists (fitting view, cargo view, etc... anywhere that is a list of item with informational columns (I guess table is a better term for this lol)). What you want to do is something like the current fleet view, where you inherit wx.Panel (IIRC) and build a simple page out of it that includes a textbox

gui/additionsPane.py
Ebag333: As I said it's not just git blame.  Compares are much easier down the line.

And yes, I know that you can chase the history of changes and see what folks did, but that takes more effort/time.

blitzmann:
Also just realized that there was talk about adding a save button. Personally, I like it autosaving so that the user doesn't forget to click save and lose a bunch of work.

Ebag333:
Autosaving is a better solution, but more work.

The PR that @blitzmann mentioned is #735.  I actually thought about that after I suggested the save button, but it is quite a bit complicated compared to a button.

The imports are `sched, time`, and since those are base python packages I certainly hope they're available!

735 is a lot more complex than what you need, that creates a timer to create a job (because of the nature of how that works).  You probably just need a timer.

Hmm, the only issue might be if the user types quickly then clicks away, the temp data might get garbage collected and lost.

Maybe change the write to DB to a write to var attached to `fit`, then create a job to run after a second or two to write that to the DB.

You can look at #804 and see how I added the var `activeByDefault`, which doesn't get saved to the DB but persists while Pyfa is running.

gui/notesView.py
resinneublem: Ah I might try that. Again wx (and pyfa) noob here

resinneublem:
Cool. Yeah I like the timer idea. I'll give that a shot.

Barring that I'll fallback to a static text with a dual edit/save view. The one thing I was a little unsure on the autosave is that I tend to "select all" and then accidentally overwrite something. It would suck to lose that info or expect them to undo it.

blitzmann:
> The imports are sched, time, and since those are base python packages I certainly hope they're available!

If they are base python, should work then. But again, this is strictly GUI stuff, so wx.Timer would work just as well. Doesn't matter to me. And I'm okay with a save button if we can't get it to save automatically with a timer

Ebag333:
>The one thing I was a little unsure on the autosave is that I tend to "select all" and then accidentally overwrite something

CTRL-A doesn't work in the text box.  No idea why (maybe bound to something else in wxPython)?  So that's at least slightly less likely to happen.

resinneublem:
Alright guys. take a look at the updates PR with the wx.Timer. Seems to work well.

resinneublem:
One other question, which isn't entirely related to this PR. But would it be a good idea to start a CHANGELOG.md file?

For example this PR could include an entry in it and it would make @blitzmann's job easier to put together the patch notes when it's released. https://github.com/pyfa-org/Pyfa/releases

Here's an example of what I'm talking about from the Rails project: https://github.com/rails/rails/pull/27058/files#diff-1ed2907b3b8f148c2533558a77673ffaR1

resinneublem:
@blitzmann I'm working on the conflict now. I'll rebase against your master

blitzmann:
@resinneublem I tried to fix the conflict. I apparently can edit directly to your branch? may not have worked, dunno

blitzmann:
Also, never been a fan of change logs. That's just one more thing we have to maintain. I try my best to contain all change logs in the release notes anyway

resinneublem:
No worries on the conflict. My change is just a one liner. So it'll be easier just to add to what you currently have. Gimme a min or two

blitzmann:
Also, are you force pushing these changes? 

Baaaaad :)

resinneublem:
> Also, are you force pushing these changes? 

Of course I am! :)

This feature doesn't need N commits to get merged. Clean git history bro

Ebag333:
>This feature doesn't need N commits to get merged. Clean git history bro

That kinda sorta defeats the whole purpose of git.

gui/notesView.py
Ebag333: Fair enough.  Probably something that could be optimized, but not worth worrying about right now.

zmoazeni:
> That kinda **takes advantage of one of the greatest strengths** of git.

Fixed that for ya.

resinneublem:
> That kinda **takes advantage of one the greatest advantages** of git.

fixed that for ya

blitzmann:
Thanks @resinneublem! This is gonna be a great addition :D

resinneublem:
Thank for the help guys!

Issue #826, opened 2016-11-20
null-sec.com is no more?


resinneublem:
Fixing this depends on the outcome of https://github.com/pyfa-org/Pyfa/issues/825
 
By default the HTML export writes links to null-sec.com which is down.

> null-sec.com’s server DNS address could not be found.

* https://www.reddit.com/r/Eve/comments/54jla3/what_happened_to_wwwnullseccom/

Looks like it was recent thing. Should we remove this default and option?

Issue #825, opened 2016-11-20
Should HTML export continue to be supported?


resinneublem:
Thinking out loud here. Is there a reason to continue supporting the HTML export now that the IGB is gone?

https://community.eveonline.com/news/patch-notes/patch-notes-for-yc118.9-release

> The in-game browser (IGB) has been removed. Please check the blog Bidding Farewell to the in-game brower for more info.

blitzmann:
It still offers a way to browse fits outside of pyfa, however it uses nullsec as it's fitting website and that no longer exists.

blitzmann:
#826 I think we should switch to Osmium for now. I don't know if it's worth keeping HTML export around though. I don't have usage metrics of it or anything.

resinneublem:
I can put together a PR dropping null-sec.com and defaulting to osmium.

I could see certain alliances or groups like Warp to Me finding it useful for sharing sets of fits outside of the game. But I don't know if they're actively using it.

https://wl.warptome.net/fits/ is an example, but even that looks customized.

blitzmann:
I'm going to say that, for now, we will continue supporting the html export. There's no reason at the moment to get rid of it, and it might prove useful for some folks. Gonna close this issue for now.

PR #827, opened 2016-11-20
Drops null-sec.com as an option

resinneublem:
It seems like it would be a good idea to make this user configurable so they could enter whatever domain they want.

However with this entire HTML export feature being questionable, I didn't go that far. https://github.com/pyfa-org/Pyfa/issues/825

closes: https://github.com/pyfa-org/Pyfa/issues/826

### DIFF
--- a/gui/builtinPreferenceViews/pyfaHTMLExportPreferences.py
+++ b/gui/builtinPreferenceViews/pyfaHTMLExportPreferences.py
@@ -17,7 +17,6 @@ class PFHTMLExportPref ( PreferenceView):
             "in a regular web browser to open them at NULL-SEC.com or Osmium."
     desc2 = "Enabling automatic exporting will update the HTML file after any change "+\
             "to a fit is made. Under certain circumstance, this may cause performance issues."
-    desc3 = "Preferred website to view fits while not using in-game browser can be selected below."
     desc4 = "Export Fittings in a minmal HTML Version, just containing the Fittingslinks " +\
             "without any visual styling or javscript features"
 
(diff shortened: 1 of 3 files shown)


resinneublem:
> I don't think we should drop the settings for the website export. We can leverage it in a number of ways, and if we wanted to add additional sites we'd end up just adding it back in anyway.

Sure. So why not wait until we have more than one option before bringing it back?

It doesn't make sense to keep around all this code and have a drop down for just one value.

Ebag333:
Because it's already there now and can be easily expanded on.

Yes it's only 40ish lines of code across 3 files, but I _hate_ doing work twice. :)

If we're going to drop the HTML export, then we should just drop it entirely (since users can export via CREST to their fitting database in game, then export out of that to another service).

I'm all for not keeping code around just for the sake of keeping code around, but this isn't really worth ripping out.  There's no performance gains, and it likely just creates more work down the line when we end up adding it back in.

resinneublem:
> Because it's already there now and can be easily expanded on.
> Yes it's only 40ish lines of code across 3 files, but I hate doing work twice. :)

Sorry pal, 100% disagree. It's as easy as a `git revert` or looking at this PR to port back the code. It's not starting from 0.

Leaving in stale code is just keeping around technical debt. You don't know if we're going to expand with another website or remove the section altogether. So let's make the best decision with the knowledge we have now. 

> I'm all for not keeping code around just for the sake of keeping code around, 

That's precisely what you're arguing for here :)


> ... but this isn't really worth ripping out. There's no performance gains, and it likely just creates more work down the line when we end up adding it back in.

You're forgetting the drain on the user having to see this and understand it. It's a UI component that allows them to select between multiple websites...but...there's only one website. "Why can't I choose something else? Am I doing something wrong? Is something not installed? Is it broken?"

As for "creating more work later". YAGNI, you're assuming this area is going to be expanded on.

> If we're going to drop the HTML export, then we should just drop it entirely (since users can export via CREST to their fitting database in game, then export out of that to another service).

I'm not arguing against this point. I just don't know if https://github.com/pyfa-org/Pyfa/issues/825 is going to be decided quickly.

Ebag333:
>Sorry pal, 100% disagree. It's as easy as a git revert or looking at this PR to port back the code. It's not starting from 0.

Easy for you, maybe.  A lot of people who submit PRs have little to no experience with Git. 

>You're forgetting the drain on the user having to see this and understand it. It's a UI component that allows them to select between multiple websites...but...there's only one website. "Why can't I choose something else? Am I doing something wrong? Is something not installed? Is it broken?"

It also exposes it to the user so they can see what website it's exporting to.

Anyway, I've said my two bits.  @blitzmann can chime in with his thoughts.

blitzmann:
I'm fine with not having a drop down for now. Later on we may want to allow the option for the user to put in their own link, but it's not priority for now.

Question though : what happens to those folks who currently have NULLSEC selected as their export? Will it now default to osmium, or is there going to be an error?

resinneublem:
It will just use osmium regardless of what they had selected. The code that used the setting no longer reads the setting.

blitzmann:
Gotcha. Will take a look at this soon - it's one of the PRs that needs to be done before release because it's currently broken.

Thanks!

Issue #818, opened 2016-11-16
Remembering filter choices


resinneublem:
I heard from a corpmate last night that one bummer about the search is:

* You search for a specific module
* You don't want officer modules so you deselect the filter at the bottom
* You find you item and then do another search
* The filters are reset

Keeping the filter choice at least for the duration of the app being open makes sense. (For example, if it was reset every time the app opened, but not every time a search occurred)

Ebag333:
The other annoying thing to me is that the behavior for searching changes the default behavior.  That changed behavior then sticks when you go back to browsing.

resinneublem:
> The other annoying thing to me is that the behavior for searching changes the default behavior.

Sorry, I'm not following. Do you have an example?

Ebag333:
>Sorry, I'm not following. Do you have an example?

Search for an item.  It turns on all categories.
Click on a market group.  All categories will be turned on (default when clicking on groups is only `Normal`).
"Closing" the search reverts to the default behavior.

resinneublem:
Ohh. I think I did run into that while testing. Pretty sure #829 fixes that too. But you might want to checkout the branch and verify.

It mainly retains the pushed state across searches. And continues to have all meta buttons pushed on startup.

blitzmann:
Eh, this closed by mistake. Keeping it open until release

PR #829, opened 2016-11-20
Remembers user's meta filtering choices

resinneublem:
There is a lot of logic that overwrites the user's choices of meta filtering on every search.

That's a little too clever. Especially if the user is looking for different variations of a certain meta.

If a user wants to include the other meta buttons they can re-add them after filtering.

closes: https://github.com/pyfa-org/Pyfa/issues/818

### DIFF
--- a/gui/marketBrowser.py
+++ b/gui/marketBrowser.py
@@ -63,6 +63,7 @@ class MarketBrowser(wx.Panel):
         self.metaButtons = []
         for name in self.sMkt.META_MAP.keys():
             btn = wx.ToggleButton(p, wx.ID_ANY, name.capitalize(), style=wx.BU_EXACTFIT)
+            btn.SetValue(True)
             setattr(self, name, btn)
             box.Add(btn, 1, wx.ALIGN_CENTER)
             btn.Bind(wx.EVT_TOGGLEBUTTON, self.toggleMetaButton)
@@ -175,7 +176,6 @@ class MarketTree(wx.TreeCtrl):
         self.marketBrowser.searchMode = False
         sMkt = self.sMkt
         mg = sMkt.getMarketGroupByItem(item)
-        metaId = sMkt.getMetaGroupIdByItem(item)
 
         jumpList = []
         while mg is not None:
@@ -197,7 +197,7 @@ class MarketTree(wx.TreeCtrl):
             self.Expand(item)
 
         self.SelectItem(item)
-        self.marketBrowser.itemView.selectionMade(forcedMetaSelect=metaId)
+        self.marketBrowser.itemView.selectionMade()
 
 class ItemView(d.Display):
     DEFAULT_COLS = ["Base Icon",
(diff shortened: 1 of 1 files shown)


resinneublem:
There was one other thing I encountered while working on this: https://github.com/pyfa-org/Pyfa/blob/a0a0ec9b1fb4e166a93396f8282d3bba565da951/gui/marketBrowser.py#L80

To include multiple meta buttons, the user has to hold down "ctrl". That doesn't totally make sense to me. I would expect "shift" to be the modifier that they would hold down.

I didn't change that in this PR. But if you all want me to, I will.

blitzmann:
I'm gonna sit on this PR for now. I am not sure I like this change. To me, when you type in something different for a search, it should "reset" and search all items. But at the same time, persisting the filters feels correct as well... Give this a while to brew. I'll also ask around and see what opinions others have. :)

> To include multiple meta buttons, the user has to hold down "ctrl". That doesn't totally make sense to me. I would expect "shift" to be the modifier that they would hold down.

I expect ctrl would be it. Whenever you select things in a GUI (such as a multiselect in HTML, or better yet in Excel), shift selects from the currently selected to the cursor and everything in between, while ctrl is selecting each individual item.

jalep:
Personally this is a feature I would like to see implemented. Most of the time, I find when a search returns every meta level item it just slows me down.

resinneublem:
@blitzmann totally fine to let this continue marinating, but I want to mention something else as I've been thinking about the pain I run into with this filter reset.

I would hazard a guess that 90%+ fits use mods that are in the "normal" filter. And you get into a different mindset when you "want to bling" a ship. But even then, I would think it's even more rare for someone to care about officer mods when blinging. So normal fit vs bling fit vs officer fits are all different mindsets when you fit a ship.

For each of those mindsets you're trying different mods. Neuts vs Prop vs Tank. Or even when in a "normal" mindset you're making fitting/dps decisions like 250mm vs 280mm arties. So each search really isn't independent. You still have a goal for the fit. And I think persisting the user's choice on that filter helps express the intention of that goal and makes it easier to find the mods you care about faster.

As I said above, I'm totally fine keeping the default (all filters selected/show all) on startup. That makes it less confusion for users who don't care about these filters.

blitzmann:
Thought about this some more, and I think I'm more open to the idea. Just have to test it now and I'll merge it for the next version release. :)

blitzmann:
Missed the latest version release. :(

Gonna merge it now into dev though :)

blitzmann:
Actually, @resinneublem , I just noticed: disabled buttons show up as toggled. Searching 200mm

image

vs this change:

image

I don't know how I didn't catch it before, but there it is. I'm reverting this change for now, please look into it when you can and submit another PR against the dev branch :)

resinneublem:
Hrm, I wonder if that's different in Windows. I'll look.

On ubuntu, "Complex" is disabled and not toggled.

I'll spin up a new PR after I reproduce.

resinneublem:
Yeah I see it now on Windows. I'll work on a fix that looks good on both.

Thanks for the catch!

PR #830, opened 2016-11-21
Industrial core for Rorq

Ebag333:
For @blitzmann

### DIFF
--- a/eos/effects/industrialcoreeffect2.py
+++ b/eos/effects/industrialcoreeffect2.py
@@ -18,7 +18,7 @@ def handler(fit, src, context):
                                   src.getModifiedItemAttr("industrialCoreRemoteLogisticsDurationBonus"),
                                   )
     fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Capital Shield Emission Systems"),
-                                  "capacitorNeed",
+                                  "maxRange",
                                   src.getModifiedItemAttr("industrialCoreRemoteLogisticsRangeBonus"),
                                   )
 
@@ -75,10 +75,6 @@ def handler(fit, src, context):
                                  "maxVelocity",
                                  src.getModifiedItemAttr("industrialCoreBonusDroneVelocity"),
                                  )
-    fit.drones.filteredItemBoost(lambda drone: drone.item.requiresSkill("Ice Harvesting Drone Operation"),
-                                 "duration",
-                                 src.getModifiedItemAttr("industrialCoreBonusDroneIceHarvesting"),
-                                 )
     fit.drones.filteredItemBoost(lambda drone: drone.item.requiresSkill("Drones"),
                                  "damageMultiplier",
                                  src.getModifiedItemAttr("industrialCoreBonusDroneDamageHP"),
@@ -97,4 +93,3 @@ def handler(fit, src, context):
                                  )
 
     #  Todo: remote impedance (no reps, etc)
-2
--- a/eos/effects/industrialcoreeffect2.py
+++ b/eos/effects/industrialcoreeffect2.py
@@ -54,10 +54,7 @@ def handler(fit, src, context):
                                   )
 
     #  Command Burst Range Bonus
-    fit.modules.filteredItemBoost(lambda mod: mod.item.group.name in
-                                              (
-                                                  "Command Burst",
-                                              ),
+    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Leadership"),
                                   "buffDuration",
                                   src.getModifiedItemAttr("industrialCoreBonusCommandBurstRange"),
                                   )
--- a/eos/effects/industrialcoreeffect2.py
+++ b/eos/effects/industrialcoreeffect2.py
@@ -93,3 +93,9 @@ def handler(fit, src, context):
                                  )
 
     #  Todo: remote impedance (no reps, etc)
+    fit.ship.increaseItemAttr("warpScrambleStatus", src.getModifiedItemAttr("siegeModeWarpStatus"))
+    fit.ship.boostItemAttr("remoteRepairImpedance", src.getModifiedItemAttr("remoteRepairImpedanceBonus"))
+    fit.ship.increaseItemAttr("disallowTethering", src.getModifiedItemAttr("disallowTethering"))
+    fit.ship.boostItemAttr("sensorDampenerResistance", src.getModifiedItemAttr("sensorDampenerResistanceBonus"))
+    fit.ship.boostItemAttr("remoteAssistanceImpedance", src.getModifiedItemAttr("remoteAssistanceImpedanceBonus"))
+    fit.ship.increaseItemAttr("disallowDocking", src.getModifiedItemAttr("disallowDocking"))
--- a/eos/effects/industrialcoreeffect2.py
+++ b/eos/effects/industrialcoreeffect2.py
@@ -24,6 +24,11 @@ def handler(fit, src, context):
                                   src.getModifiedItemAttr("industrialCoreRemoteLogisticsRangeBonus"),
                                   stackingPenalties=True
                                   )
+    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Capital Shield Emission Systems"),
+                                  "falloffEffectiveness",
+                                  src.getModifiedItemAttr("industrialCoreRemoteLogisticsRangeBonus"),
+                                  stackingPenalties=True
+                                  )
 
     #  Local Shield Repper Bonuses
     fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Capital Shield Operation"),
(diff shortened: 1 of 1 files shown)


eos/effects/industrialcoreeffect2.py
```diff
@@ -1,29 +1,100 @@
...
+    fit.ship.multiplyItemAttr("scanResolution", src.getModifiedItemAttr("scanResolutionMultiplier"))
+
+    #  Remote Shield Repper Bonuses
+    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Capital Shield Emission Systems"),
+                                  "duration",
+                                  src.getModifiedItemAttr("industrialCoreRemoteLogisticsDurationBonus"),
+                                  )
+    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Capital Shield Emission Systems"),
+                                  "capacitorNeed",
+                                  src.getModifiedItemAttr("industrialCoreRemoteLogisticsRangeBonus"),
```
blitzmann: You're also missing two effect related to this: `maxRange` and `falloffEffectiveness`, using the `industrialCoreRemoteLogisticsRangeBonus` attribute

eos/effects/industrialcoreeffect2.py
```diff
@@ -1,29 +1,100 @@
 # industrialCoreEffect2
 #
 # Used by:
-# Variations of module: Industrial Core I (2 of 2)
+# Module: Industrial Core I
```
Ebag333: I didn't change it. :)

I'll often set them to *SOMETHING* but that's only for my own reference.

eos/effects/industrialcoreeffect2.py
Ebag333: >this should be industrialCoreRemoteLogisticsDurationBonus, not the range bonus

The `src.attribute` is correct.  The module attribute is not, copy/pasta error from local repper.

Fixed (soon to be submitted).

>You're also missing two effect related to this: maxRange and falloffEffectiveness, using the industrialCoreRemoteLogisticsRangeBonus attribute

How do you know that the range bonus effects `falloffEffectiveness`?

eos/effects/industrialcoreeffect2.py
```diff
@@ -1,29 +1,100 @@
...
+                                  src.getModifiedItemAttr("industrialCoreBonusMiningBurstStrength"),
+                                  )
+
+    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Mining Foreman"),
+                                  "warfareBuff4Value",
+                                  src.getModifiedItemAttr("industrialCoreBonusMiningBurstStrength"),
+                                  )
+
+    #  Command Burst Range Bonus
+    fit.modules.filteredItemBoost(lambda mod: mod.item.group.name in
```
Ebag333: Other items require the Leadership skill (fighters, for example).

Why not filter on just the group? Lots of other effects do so.

Alternatively we can filter on skills and target the 5 command burst skills.

Ebag333:
>Most of them we don't use in pyfa, either because there's just no need, or because we don't support the feature yet. But it's still a good idea to include these so that they show up in attributes and less work will need to be done to support them in the future. :)

Remote impedance isn't done at _all_ yet.  Which we really need to do.  There's an issue open for it.

>I beleive things like the Siege effect for dreads has these that you can refer to

We already do `fit.extraAttributes["siege"] = True`

eos/effects/industrialcoreeffect2.py
```diff
@@ -1,29 +1,100 @@
...
+                                  src.getModifiedItemAttr("industrialCoreLocalLogisticsDurationBonus"),
+                                  )
+    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Capital Shield Operation"),
+                                  "shieldBonus",
+                                  src.getModifiedItemAttr("industrialCoreLocalLogisticsAmountBonus"),
+                                  )
+
+    # Mining Burst Bonuses
+    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Mining Foreman"),
+                                  "warfareBuff1Value",
```
blitzmann: That's actually probably my bad. It has Value, but I'm trying to set a Multiplier. Or some shit. So it's broken. Awesome :D

Ebag333: >That's actually probably my bad.

Since they have both attributes wouldn't it be CCPs bad? :D

eos/effects/industrialcoreeffect2.py
blitzmann: Because that's not how the effect is defined. We're filtering on modules, not fighters, so we don't have to worry about it applying to fighters.

blitzmann:
>Remote impedance isn't done at all yet. Which we really need to do. There's an issue open for it.

Hence we we need those funky values, if only for completeness in the attributes.

> We already do fit.extraAttributes["siege"] = True
I meant the impedance and disable attribtes. I don't actually know off the top of my head what the siege extra attribute does... it might be old and unused.

Ebag333:
>I don't actually know off the top of my head what the siege extra attribute does... it might be old and unused.

It's used in effects at least.

```
    if fit.extraAttributes["siege"]:
        fit.ship.increaseItemAttr("commandBonusEffective", ship.getModifiedItemAttr("shipBonusORECapital2"),
                                  skill="Capital Industrial Ships")
```

Since that's a passive effect, kinda makes sense?  I would think it'd make more sense to just make it active but.... vOv

It's not used by anything (anymore), that's the old `shipcommandbonuseffectivemultiplierorecapital2.py`

blitzmann:
I guess my question is why is it included in this effect? How did this get brought up? I brought up Siege so that you could go to the siege effect and see how we set the other values that are currently not set. <_<

Ebag333:
> guess my question is why is it included in this effect? How did this get brought up?

It was in the effect before I got here, I swear!

Seriously tho, running the Indy core is basically like sieging, so it makes sense to do it.

eos/effects/industrialcoreeffect2.py
blitzmann: I believe there are instances where if you modify an attribute that isn't there, it will actually add it to the list. vov its something I need to look at, but the boosts seem to be working okay

eos/effects/industrialcoreeffect2.py
blitzmann: >How do you know that the range bonus effects falloffEffectiveness?

Because eos and the dgmeffects file tells me. 

> The src.attribute is correct. The module attribute is not, copy/pasta error from local repper.

Both of us were correct. It was a mismatch, but there are a total of four modifiers that utilize `Capital Shield Emission Systems`, you have two and are missing the other two.

This is what I have:
```
fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Capital Shield Emission Systems"), "falloffEffectiveness", src.getModifiedItemAttr("industrialCoreRemoteLogisticsRangeBonus"), stackingPenalties=True)
    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Capital Shield Emission Systems"), "maxRange", src.getModifiedItemAttr("industrialCoreRemoteLogisticsRangeBonus"), stackingPenalties=True)
    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Capital Shield Emission Systems"), "capacitorNeed", src.getModifiedItemAttr("industrialCoreRemoteLogisticsDurationBonus"))
    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Capital Shield Emission Systems"), "duration", src.getModifiedItemAttr("industrialCoreRemoteLogisticsDurationBonus"))
```

blitzmann:
@Ebag333 I'm also showing stacking penalties being applied on these modifiers:

scanResolution
Drones > damageMultiplier
Drones > maxVelocity
Leadership > maxRange
Capital Shield Emission Systems > maxRange and falloffEffectiveness
Capital Shield Operation > shieldBonus

Which makes sense since these attributes are usually penalized in other effects.

blitzmann:
@Ebag333 any news on the changes?

I can add them myself, but I've rather you finish it if possible. If not, let me know and I'll add the missing data tonight before pre-release

blitzmann:
For documentation

This modifier is not needed
```
    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Leadership"),
                                  "falloffEffectiveness",
                                  src.getModifiedItemAttr("industrialCoreBonusCommandBurstRange"),
                                  stackingPenalties=True
                                  )
```

Additionally, you were missing a modifier for `capacitor need`:

```
    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Capital Shield Emission Systems"),
                                  "capacitorNeed",
                                  src.getModifiedItemAttr("industrialCoreRemoteLogisticsDurationBonus")
                                  )
```

Otherwise looks good. i've fixed these issues and have merged this into sisi branch

PR #832, opened 2016-11-22
Effect toggle (rebased)

Ebag333:
Seems it's just easier to create a new PR.


Ebag333:
I'm not quite sure how I got this into a new PR.  But when merging it back to the original PR it wanted to modify over 2k files.  This way it's only the actual 24 we're touching.

Also, don't ask me why the two commits show twice....

See #804 for history.

PR #848, opened 2016-11-26
Run recalc as a timer service

Ebag333:
Replaces #735

So this does a couple things.

First off, we have a new thread that simply runs through a recalc.  There is a .5 second delay between runs, or a 1 second delay if we're in debug mode (because logging slows down the processing a lot, so since this is threaded we can run it faster than everything else can run).

The second half of this is we create a scheduled task to flag the thread that a recalc needs to be run, if something requests a recalc.  This creates the job on a 1 second delay, or 2 second if running in debug mode.

If a recalc is requested, we perform the following steps:
1. Check to see if a scheduler to run the recalc job already exists, if it does then destroy it and create a new one.  This allows the end user to indefinitely extend the delay before the recalc job is run, so we don't try and run the job while they're clicking around on things.
2. Once the scheduler timer is up, we flag the job to actually run the recalculations.
3. The scheduled task runs, saves to DB, recalcs, and refreshes the GUI.  If everything goes okay, it flags itself to not run anymore, if any exceptions occur we do _not_ set this so the recalculations will run again.

@blitzmann I think this is ready to merge.  We can play with the timing on it (shorten/lengthen it) if you want it to feel more/less responsive.  The down side to making it any shorter is that the recalc job might run while the user is still clicking around.  Of course if we take too long to trigger it, the user might think that Pyfa is slow.

This _drastically_ helps with things such as adding 8 guns, or 20 drones, etc.  The 1 second timer for the scheduler seems to work for me for doing things such as duplicating objects (turrets with ammo, for example), a shorter delay results in the job running before I'm done with the task I'm doing.

The threaded job could be shortened as well.  However, if I ran it too quickly (especially when in debug mode), it'll run fast enough to step on itself and this causes things like the DB isn't ready, so we get weird DB errors, or the fit object isn't fully created so `fit.clear()` fails.  This _may_ be why it's run 2-3 times, because things _seem_ to calculate correctly running it only once if we slow things down a bit.

And of course, if there _is_ an error on one of the DB commits, or `fit.clear()`, etc, we will rerun the recalc job again, to hopefully clear it up.  This seems to work fine, but by slowing it down a bit (hence the .1 second delays you see in the job) we don't seem to run into these.  Even though we're adding 200 ms to the job (plus the time it takes to actually run it), we're no longer running it 3 times, so it's a net win.

Play around with it, and let me know what you think.

### DIFF
--- a/service/fit.py
+++ b/service/fit.py
@@ -1209,11 +1209,9 @@ class Fit(object):
             args = []
             #  Slow down processing to account for the longer time debug mode takes
             if config.debug:
-                recalcJobID = schedule.enter(2, 1, self.triggerRecalc, args)
+                recalcJobID = schedule.enter(1.75, 1, self.triggerRecalc, args)
             else:
-                recalcJobID = schedule.enter(1, 1, self.triggerRecalc, args)
-
-                # schedule.run()
+                recalcJobID = schedule.enter(1.5, 1, self.triggerRecalc, args)
         else:
             logger.debug("Fit empty, skipping job creation.")
 
@@ -1232,6 +1230,9 @@ class RecalcJob(threading.Thread):
 
     def run(self):
         self.resume()  # unpause self
+        # Slow down the first run, otherwise it errors out
+        # AttributeError: 'MainFrame' object has no attribute 'fitMultiSwitch'
+        time.sleep(1)
         while True:
             if self.paused:
                 # logger.debug("Nothing to run")
@@ -1242,25 +1243,27 @@ class RecalcJob(threading.Thread):
 
                 # Slow down processing to account for the longer time debug mode takes
                 if config.debug:
-                    time.sleep(1)
-                else:
                     time.sleep(.5)
+                else:
+                    time.sleep(.25)
             else:
                 logger.debug("Starting Recalc Job")
                 #  do stuff
-                time.sleep(.1)
                 self.iterations += 1
 
-                mainFrame = gui.mainFrame.MainFrame.getInstance()
-                sFit = Fit.getInstance()
+                try:
+                    mainFrame = gui.mainFrame.MainFrame.getInstance()
+                    sFit = Fit.getInstance()
 
-                fitID = mainFrame.getActiveFit()
-                if fitID:
-                    fit = eos.db.getFit(fitID)
-                else:
-                    fit = sFit.getFit(mainFrame.getActiveFit())
+                    fitID = mainFrame.getActiveFit()
+                    if fitID:
+                        fit = eos.db.getFit(fitID)
+                    else:
+                        fit = sFit.getFit(mainFrame.getActiveFit())
+                except AttributeError:
+                    # Mainframe doesn't exist yet most likely
+                    fit = None
 
-                pass
                 skipPause = False
                 if fit:
                     logger.debug("=" * 10 + "recalc" + "=" * 10)
--- a/service/fit.py
+++ b/service/fit.py
@@ -117,8 +117,8 @@ class Fit(object):
             "showMarketShortcuts": False,
             "enableGaugeAnimation": True,
             "exportCharges": True,
-            "openFitInNew":False
-            }
+            "openFitInNew": False
+        }
 
         self.serviceFittingOptions = SettingsProvider.getInstance().getSettings(
             "pyfaServiceFittingOptions", serviceFittingDefaultOptions)
@@ -1213,21 +1213,12 @@ class Fit(object):
             else:
                 recalcJobID = schedule.enter(1, 1, self.triggerRecalc, args)
 
-            #schedule.run()
+                # schedule.run()
         else:
             logger.debug("Fit empty, skipping job creation.")
 
-
     def triggerRecalc(self):
-        global recalcJobID
-        if schedule.empty():
-            logger.debug("Scheduler empty")
-            pass
-        else:
-            pass
-
         logger.debug("Running recalc job")
-        # RecalcTimer = RecalcJob()
         self.RecalcTimer.resume()
 
 
@@ -1243,7 +1234,7 @@ class RecalcJob(threading.Thread):
         self.resume()  # unpause self
         while True:
             if self.paused:
-                #logger.debug("Nothing to run")
+                # logger.debug("Nothing to run")
                 if not schedule.empty():
                     if schedule.queue[0].time < time.time():
                         #  Check if we're past the scheduled time for the job, then run it
@@ -1317,7 +1308,7 @@ class RecalcJob(threading.Thread):
                         # fit.inited = True
                         logger.debug("=" * 10 + "recalc end" + "=" * 10)
 
-                #  Don't run the recalc again, unless something went wrong
+                # Don't run the recalc again, unless something went wrong
                 if not skipPause:
                     self.pause()
 
--- a/service/fit.py
+++ b/service/fit.py
@@ -19,9 +19,13 @@
 
 import locale
 import copy
+import config
 import threading
 import logging
 import wx
+import gui
+import gui.globalEvents as GE
+import sched, time
 from codecs import open
 
 import xml.parsers.expat
@@ -39,6 +43,7 @@ from service.settings import SettingsProvider
 from service.port import Port
 
 logger = logging.getLogger(__name__)
+schedule = sched.scheduler(time.time, time.sleep)
 
 
 class FitBackupThread(threading.Thread):
@@ -93,6 +98,11 @@ class Fit(object):
         self.character = Character.getInstance().all5()
         self.booster = False
         self.dirtyFitIDs = set()
+
+        #  Recalc job
+        self.recalcJobID = []
+        self.RecalcTimer = RecalcJob()
+        self.RecalcTimer.start()
 
         serviceFittingDefaultOptions = {
             "useGlobalCharacter": False,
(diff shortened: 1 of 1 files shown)


Ebag333:
Just some time metrics.

This is running from pyCharm, full debugging (pyCharm debug on and Pyfa debug on).

Task was launching Pyfa, opening a Catalyst fit, removing 7 guns, then duplicating 8th gun into the 7 empty slots.

Current branch (`pyfa.org\master`):
(I stripped the text between recalcs as otherwise it would have been quite long, a whopping 134 lines long.)
```
2016-11-26 01:39:53,740 root                     INFO     Starting pyfa
2016-11-26 01:40:26,661 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:30,956 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:34,532 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:35,788 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:37,117 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:38,272 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:39,414 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:40,509 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:41,569 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:48,447 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:50,134 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:51,601 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:53,288 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:55,075 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:56,628 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:58,361 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:40:58,892 eos.saveddata.fit        DEBUG    Timer - Fit: 366, Catalyst - Done with fit calculation - 0.00ms (529.00ms elapsed)
```

Total time (not including the startup time) is 32 seconds.

And here's with it as a job:
(No lines have been removed, so you can see how many times recalcs were requested. The first `Starting Recalc job` is just running it once to set it up, the real action doesn't happen until the first time it's scheduled)
```
2016-11-26 01:42:47,414 root                     INFO     Starting pyfa
2016-11-26 01:42:53,795 service.fit              DEBUG    Starting Recalc Job
2016-11-26 01:43:15,723 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:16,092 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:16,092 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:18,306 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:18,308 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:19,286 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:19,286 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:19,881 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:19,881 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:20,493 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:20,493 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:21,098 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:21,099 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:21,661 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:21,663 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:22,352 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:22,354 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:24,440 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:24,440 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:25,921 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:25,923 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:27,098 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:27,099 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:28,391 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:28,391 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:29,667 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:29,668 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:30,941 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:30,943 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:32,220 service.fit              DEBUG    Job already scheduled. Cancelling existing job
2016-11-26 01:43:32,221 service.fit              DEBUG    Scheduling fitting job
2016-11-26 01:43:34,947 service.fit              DEBUG    Running recalc job
2016-11-26 01:43:35,947 service.fit              DEBUG    Starting Recalc Job
2016-11-26 01:43:36,049 service.fit              DEBUG    ==========recalc==========
2016-11-26 01:43:36,165 eos.saveddata.fit        DEBUG    Starting fit calculation on: Fit(ID=366, ship=Catalyst, name=Catalyst) at 0x6fae350, withBoosters: True
2016-11-26 01:43:36,661 eos.saveddata.fit        DEBUG    Timer - Fit: 366, Catalyst - Done with runtime: early - 494.00ms (494.00ms elapsed)
2016-11-26 01:43:37,289 eos.saveddata.fit        DEBUG    Timer - Fit: 366, Catalyst - Done with runtime: normal - 629.00ms (1123.00ms elapsed)
2016-11-26 01:43:37,306 eos.saveddata.fit        DEBUG    Timer - Fit: 366, Catalyst - Done with runtime: late - 18.00ms (1141.00ms elapsed)
2016-11-26 01:43:37,308 eos.saveddata.fit        DEBUG    Timer - Fit: 366, Catalyst - Done with fit calculation - 1.00ms (1142.00ms elapsed)
Command: 
{}
2016-11-26 01:43:37,410 service.fit              DEBUG    ==========refresh gui==========
2016-11-26 01:43:37,415 service.fit              DEBUG    ==========recalc end==========
```

Total time, 22 seconds.

The end user interaction ended at 01:43:32, almost 5 seconds before the job completed.  4 of those seconds is waiting for the job to start, so we could probably shave another 2-3 seconds off that time by tuning the various delays in running the job.

Ebag333:
So ran some tests.

The scheduler task creation basically controls how long we give the user before the recalc job kicks off.  With a 1 second timer, you could do things like add drones or turrets (just fast double clicking) but longer tasks (like duplicating an existing turret) would take too long.

I also sped up how often the thread loops.  Can probably speed it up even further, but the gains would be minimal.

blitzmann:
@Ebag333 haven't looked at this latest version (and very briefly looked over the text here), but I did revisit the old version a few nights ago and was kinda miffed at the very noticeable  delay when off lining / onlining a module to see the difference in stats. 

Couple questions:
* Did you happen to add a preference to do this? If not, there should be. :)
* I am seeing a lot of `time.sleep()`. This concerns me. Without taking a good look at it or fiddling with it on my local machine, I'm assuming these are there for a lot of "catch up" events. This makes me uneasy - it's usually bad practice to intentionally slow down execution, and usually if it needs to happens it's only don'e once. But we seem to be using it to help time for the other thread? I feel like this is the wrong way to do it (not that I know off the top of my head what the right way is)
* Did you ever solve the GUI refresh issue?  don't see anything in the code about catching the event or anything like that.

Ebag333:
>I did revisit the old version a few nights ago and was kinda miffed at the very noticeable delay when off lining / onlining a module to see the difference in stats.

Well, that would be because the old version never triggered a GUI refresh.

>Did you happen to add a preference to do this? If not, there should be. :)

No, and I don't think there should be.  After tinkering with the times the vast majority of the time the "feel" is significantly faster, the only time there is a slight delay that is noticeable is if you do one simple thing (like say online/offline a module).  That's of course due to the recalc not running immediately as it did before.  Any time you did multiple actions (which, most users will tend to do) it feels faster because we're only running the recalc job once, not one time for each action.

So the more actions you queue up, the worse it feels, which is why adding 5 drones or 8 turrets was extremely painful, because you could double click 8 times very rapidly and queue up 8 recalcs before the first one even fired off.

Also, most users won't understand it.  "Run recalc as a threaded task, what does that mean?"

>I am seeing a lot of time.sleep(). This concerns me. Without taking a good look at it or fiddling with it on my local machine, I'm assuming these are there for a lot of "catch up" events. This makes me uneasy - it's usually bad practice to intentionally slow down execution, and usually if it needs to happens it's only don'e once. But we seem to be using it to help time for the other thread? I feel like this is the wrong way to do it (not that I know off the top of my head what the right way is)

The time sleeps are unfortunately required.  Some of them won't be noticeable at all, and some are just to keep things sane, because the threaded job will run as fast as we let it.

Some of them are because we can step on ourselves, which I suspect might be why we run the recalc 3 times.

For example:
```
try:
	eos.db.commit()
	# Delay to let the DB catch up
	time.sleep(.1)
except:
	#  Something went wrong, let the job run again to try and recalc
	skipPause = True
	logger.debug("DB commit error")
```

We do two things here to protect ourselves, one give the DB time to finish writing (because if we don't, then we get a DB error) and two if somehow we _still_ get a DB error, we'll run the recalc job again.

Without the 100 ms delay I would occasionally get DB errors, usually lock errors or similar.

Same sort of thing with:
```
fit.clear()
fit.calculateModifiedAttributes(withBoosters=True)
fit.fill()
```

Here we batch them into a single try because usually if one fails it's `clear()`, and I was seeing errors like `AttributeError: 'MainFrame' object has no attribute 'fitMultiSwitch'`.  Basically again, moving too quickly so other components of Pyfa didn't fully have everything built out yet.  Also with these, because we refresh the GUI immediately afterward, I would occasionally see the GUI refresh with the old numbers.  Both those scenarios are covered by another 100 ms delay.

So now we're up to a total of 200 ms of delay in the recalc job.  But since we aren't running it 2-3 times on first load, or multiple times because we were quickly clicking on things and queued it up, the total time savings ends up being significant.

>Did you ever solve the GUI refresh issue? don't see anything in the code about catching the event or anything like that.

This seems to do the trick:
`wx.PostEvent(mainFrame, GE.FitChanged(fitID=fitID))`

You can actually see the GUI refresh. :)

Ebag333:
@blitzmann can we revisit this?

>Is is just my ****** pc or is this extremely laggy/slow? Works well otherwise, but god its laggy. +1 for not being as lazy as eft though.
>(as in, if i unheat 5 modules at the same time it takes up to 10 seconds before the entire things is useable again and for the mods to be no longer overheated)

I'm not the only one feeling this pain. :)

ftsartek:
This is a good concept, and would like to see it implemented. Ebag just needs to fix the graph interactions with this if you add the cap life/and repairing graphs.

Ebag333:
>This is a good concept, and would like to see it implemented. Ebag just needs to fix the graph interactions with this if you add the cap life/and repairing graphs.

The graph issues are part of the graph window, and that needs an entire rework.

I believe what you told me on Slack was:
>okay release Pyfa loads fits a lot slower

Though to be fair, the current pre-release/release runs the recalcs 2-3 times while this version only runs it once.  So it's got a lot more work to do! :)

Ebag333:
This is still something that I want to do, but it should be held off until we handle threads through a central handler.

Let's leave this open, however, so it doesn't get dropped, and there's good info in here on what we need to move to the threaded job.

Ebag333:
Closing as we have #976 open.

Issue #780, opened 2016-10-21
Mac Deprecated version no longer works on macOS sierra version 10.12


JoshuaNotowitz:
As is it says in the tittle the program doesn't launch it just bounces in the dock.  This wouldn't be a problem but, at least on my machine, the non deprecated version is measurable worse in regards to performance and ease of use.

blitzmann:
Unfortunately I don't think there's anything I can do about it, unless we bring on a developer with mac experience. I simply don't have the time myself to get a copy of 10.12, go through the process of installing it as a hackintosh, and testing / developing on it.

Could you open the package file, browser to `Contents/MacOS`, and run the `pyfa` file in the terminal? Hopefully it pops up with an error that is useful.

Also, i know the wx3 version is less than ideal for anyone >10.10 (or 10.11, don't remember) due to performance, but what exactly are you having problems with?

Ebag333:
Isn't there a beta wxPython version for newer macs?

JoshuaNotowitz:
> Joshua$ /Users/Joshua/Downloads/pyfa\ 2.app/Contents/MacOS/pyfa ; exit;
Segmentation fault: 11
logout
Saving session...
...copying shared history...
...saving history...truncating history files...
...completed.

> [Process completed]

Is what it says terminal when I try to launch it.  And as far as problems its mostly just a general slow down as compared to before with the deprecated version.  Search results load slower and ship fits load slower.

blitzmann:
> Isn't there a beta wxPython version for newer macs?

Are you talking about wxPython Pheonix? That's a thing, but it's incompatible with Classic (and might be only python 3, not sure). It could help, and converting wouldn't be too diffult, just time consuming, and I don't know if it'll help.

> Is what it says terminal when I try to launch it. And as far as problems its mostly just a general slow down as compared to before with the deprecated version. Search results load slower and ship fits load slower.

yeah, that error doesn't help at all. =/ You're gonna have to go with the new version. I know it's not optimal, but my hands are tied at the moment. I know that it's slow and sluggish. There is an option in the preferences to disable animations that help with the stats panel updating.

Or, you could always try the Windows release in Wine. >_>

blitzmann:
Would you be able to run pyfa from source? Do you have the tools / know how to do so?

JoshuaNotowitz:
I don't think I would be able to I'm not very technically skilled 😄

blitzmann:
Duplicate of #954

Issue #548, opened 2016-03-27
Putting Pyfa inside a folder with special characters breaks it.


Altrue:
(On Windows 7)

Hi,
I'm a french guy, as such my Download folder is named "Téléchargements". Trying to launch Pyfa from inside this folder (Téléchargements/Pyfa) will display a nasty error message. But "Telechargement" is fine.

It's not major or anything but I thought I should let you know ;)

neutronst4r:
I just run "python2 Downloads/P%yfá-mastér/pyfa.py" on my Arch system and it works fine without error messages. So seems to be Windows specific.

Ebag333:
Windows has rather strict limits on what you can and can't do with folder names.

Unfortunately, there isn't a good way around this.  I'm planning on looking at some folder stuff anyway (there's issues with the way Pyfa handles it currently) so I might look into this at the same time.

blitzmann:
I feel like this would be an easy fix and probably has to do with unicode issue or whatever.

Ebag333:
@blitzmann I think I figured out how to fix this...

blitzmann:
@Ebag333 Awesome! it really shouldn't be too difficult, just ensuring that we are being safe with unicode paths. I'll await your pull request - I'd like to get this merged sooner rather than later if you've already got a fix in for it :)

Ebag333:
It's kinda sorta not difficult?

The fix is basically this (for the save location:
```
    root = os.path.expanduser(os.path.join("~", ".pyfa"))
    if type(root) == str:  # leave unicode ones alone
        try:
            root = root.decode('utf8')
        except:
            root = root.decode('windows-1252')
```

The _catch_ (and there always is one isn't there) is that we bloody use the built in os.path.join all over the freaking place.

I'm writing two functions (one for Pyfa location, one for save location) to handle it, and should handle appending names/etc cleanly as well as returning just the path, so will then need to convert the 64 times we use `os.path.join` (along with the dozens of other path functions) to point to the two functions in `config.py`.

So yeah, relatively easy fix, painful with a lot of refactoring.

blitzmann:
@Ebag333 this doesn't seem like a correct fix. =/

Give me a bit of time to look into it before you go through the trouble. I'd like to research this a bit more for my own benefit as well (unicode kicks my ass)

blitzmann:
[...]

PR #854, opened 2016-11-27
Handle unicode, utf8, and windows-1252

Ebag333:
Change most of the various `os.path` functions looking up the Pyfa root or saved data paths to point back to two functions in `config.py`.

This lets use more cleanly handle these with different unicode charcters and formatting types, and we have a central location for this if down the road we need to do something special here.


blitzmann:
For future fyi, please don't include all the formatting changes in things like this - branch them off into a separate cleanup branch and request PRs from that. It makes it difficult to see what has changed and what hasn't, and for a change as low-level as this, it's important not to miss anything.

That being said, I'll take a look at this hopefully before release. :)

Ebag333:
Fair enough, the reformatting was a bit more extensive than I thought it'd be, and I knew you were going to yell at me for it. :D

Ebag333:
@blitzmann is there a reason that you took out the decodes for different file systems? Or am I reading that last commit wrong?

Ebag333:
never mind. I see what you did there. 

```
mac-deprecated fails to launch, mac launches and works, has some issues with resolution that may be related to retina screens, seen here; https://puu.sh/sMHKs/b7d0514a47.png

systemspecs for reproduction: 
- Mid 2015 15” Macbook Pro (w/o discrete graphics)
- macOS 10.12.1
- No pyfa otherwise installed beforehand.
```

Correct me if I'm wrong, but the depreciated version is for older OSx versions, right?

blitzmann:
Yes, apparently it no longer works in later versions of os x. see #780

I did manage to find a couple os x VMs last night, but haven't tested them to see if they actually still work. If they do, I'll try to test tonight as well. The retina resolution has been a problem since forever...

blitzmann:
image

This doesn't see to work in Windows, at least with the packaged version. I used the example in #548. Haven't tried with source yet.

blitzmann:
Also, while it doesn't solve the issue it meant to solve, I'm going to keep it in the dev branch. It doesn't break current functionality as far as I can tell.

Ebag333:
that is one I tried that worked fine before you rewrote it. Not sure what changed to break it again. (testing on Windows)

blitzmann:
On a more serious note:

```
Missing icon file: /home/ryan/Pyfa/imgs\gui\pyfa.png
  File "pyfa.py", line 135, in <module>
    MainFrame(options.title)
  File "/home/ryan/Pyfa/gui/mainFrame.py", line 135, in __init__
    i = wx.IconFromBitmap(BitmapLoader.getBitmap("pyfa", "gui"))
  File "/usr/lib/python2.7/dist-packages/wx-3.0/wx/_gdi.py", line 1342, in IconFromBitmap
    val = _gdi_.new_IconFromBitmap(*args, **kwargs)
ValueError: invalid null reference in method 'new_IconFromBitmap', expected argument 1 of type 'wxBitmap const &'

```

On my Linux Mint setup. This is much more problematic, as pyfa simply fails to run period. :)

Ebag333:
Urgh, I wish you'd just left it until I had a chance to look at it.  Now I've gotta figure out what's changed between where my branch was and where you left this (before the revert).

Ebag333:
Yeah, hate to say it but something you changed broke the fix.  Going to poke at this a bit and see if I can figure out what did....

Probably why it's not working on Linux either.

blitzmann:
> Yeah, hate to say it but something you changed broke the fix. Going to poke at this a bit and see if I can figure out what did....

Bullshit. I checked if my -one change- broke it before I even posted, thanks. I even just now isolated your changes only and it's still broken.

Your shit's broken. Own it, fix it, move on.

blitzmann:
FYI, this is probably where it is breaking. you most likely overlooked this bit:

https://github.com/pyfa-org/Pyfa/blob/master/gui/bitmapLoader.py#L77

Why this doesn't work on Linux but it works on Windows, I don't have a clue right now. But it's an area to start dropping `print`s in.

Ebag333:
FYI as is it's broken on Windows. When I committed the PR, this particular test `Téléchargements` worked fine.

Ebag333:
>FYI, this is probably where it is breaking. you most likely overlooked this bit:
>https://github.com/pyfa-org/Pyfa/blob/master/gui/bitmapLoader.py#L77

This did get missed in the original PR looks like.

blitzmann:
> FYI as is it's broken on Windows. When I committed the PR, this particular test Téléchargements worked fine.

Be more specific. It worked with the skels? Or it worked from source? Because I can tell you without a doubt - I just built a package with your unicode path commit, and that use case does not work. Please outline your testing and what you did to determine that it worked - maybe we're both missing something here.

For reference, this is the original commit - 0d4f24

Ebag333:
Replaced the archive files with the updated .py files. It worked.

I'm 99.9% sure that this error I ran into before. I may have fixed it and simply not committed that fix (which would be dumb on my part).  It's hard for me to tell what's going on with the PR because that branch for some reason has a bajillion commits after it.  Reverting back to the original commit has this problem, but I'm 99% sure that I didn't have a PR with just a single commit.

Given that this all was done near a month ago though, who knows.  Memory is faulty.

blitzmann:
Just realized: `/home/ryan/Pyfa/imgs\gui\pyfa.png` Notice the inconsistent slash directions. This might be why it's breaking.

Ebag333:
>Just realized: /home/ryan/Pyfa/imgs\gui\pyfa.png Notice the inconsistent slash directions. This might be why it's breaking.

To be fair I never claimed to test it on Linux. :)

Not sure how Linux handles alternating slashes.  This very well could be the root of that issue.

blitzmann:
> It's hard for me to tell what's going on with the PR because that branch for some reason has a bajillion commits after it. Reverting back to the original commit has this problem, but I'm 99% sure that I didn't have a PR with just a single commit.

You did.

Are you still using GitHub app? I don't know how they display stuff, but with GitExtensions it's very easy to see what happened:

image

then a few week later: 

image

You can see that I brought the branch up to date with dev, made my commit, and then merged back into dev. Until that point, that one commit was it. 

> To be fair I never claimed to test it on Linux. :)

https://pyfa.slack.com/archives/D2Q139FTJ/p1481484319000065

I took this to mean that you were going to test it.

Ebag333:
So there were definitely commits that I either never committed or committed locally but never synched.  At least I'm _guessing_ that's what happened, I know that some of these issues I solved before, and I see specific ones that would have failed with a simple test (even in Windows, no need for Mac/Linux) that I know were working before.

Anyway, doesn't matter.

For the issue with:
```
    certName = "cacert.pem"
    os.environ["REQUESTS_CA_BUNDLE"] = getPyfaPath(certName)
    os.environ["SSL_CERT_FILE"] = getPyfaPath(certName)
```

`os.environ` chokes if you try and shove something into it like paths straight up decoded from `windows-1252`.  I don't remember how I solved that before (if I had to guess, probably either using `str(` or `unicode(`), but from a quick search it seems it's a known problem and most people handle it by using `os.environ["REQUESTS_CA_BUNDLE"] = getPyfaPath(certName).encode('utf8')` to put it into a format it handles.

Other things choke if you pass it `utf8` (like anything that uses `os.path`), so we can't just convert it _all_ to `utf8`...that'd be too easy.

@blitzmann I probably stepped on what you did figuring this out, so for that I apologize.  I cleaned up all references to `os.path.join` so they shouldn't exist in the codebase anymore (_except_ in the case of `eos\config`, which I have not touched as it _seems_ to work fine as is, and I wasn't sure if you wanted to point that at the root `config` or not).  

It actually did give me a chance to clean up a few paths that don't point to the standard locations, where we'd done the same fix but outside of `config`.  Now they use the function from `config` so we don't have the same checks scattered through the code base.

I did change the function name as `_getPath` was used twice, and `getPath` is too close to many other things.  If you prefer something other than `parsePath` feel free to change it, just make sure you get the files outside of `config` that point to it.

I think I hit all of the locations where we're getting paths, definitely did for anything use `os.path.join`.  As of right now it's working fine for the test of `Téléchargements`, but I haven't tested it for Mac or Linux.

PR #855, opened 2016-11-27
Fix Mining Rig Effect for Ice Harvesters

Ebag333:
Fixes erroneously applying amount bonus to Ice Harvesters.

Removes unnecessary skill level flag.


blitzmann:
> Removes unnecessary skill level flag.

This is necessary: see Mining Drone Operation skill. Please try not to tweak things if not, like, 92% sure :)

Will merge and fix

PR #867, opened 2016-12-01
Add Skill sets for Alpha Clones

Tommallionus:
Please add option for simulating fitting under Alpha Clones max skills for each of four races (like all lvl5 now, just sth like "Alpha Clone Gallente Max skills"). 

Thank You! 
Tommallionus


Ebag333:
I don't think you meant to submit a pull request. :)

There are multiple issues open for Alpha clones already.

TL;DR: CCP hasn't given a way for us to handle this yet.  We are working on some work arounds between now and when they do.

PR #868, opened 2016-12-01
Preference Panes

Ebag333:
This PR contains a number of enhancements to the Preference panes.

1. Fold in the preference pane settings from #690.  This gives the end user the ability to turn on and off individual panes, and lets out build out each pane separately.  By breaking out the underlying functionality, we can enhance the panes (or add new ones) once we can all agree on how exactly they should look.
2. Adds a logging preferences pane.  Right now all this does is show if debug mode is on or not, but I'd like to enhance it with exposing the log path.  Also right now this is a read only setting, the user cannot change it (and it even warns them of this).  Future enhancement, we can make this togglable while running (should be fairly easy to do).
3. Add a database preferences pane.  This is mostly to just expose where the DBs are located so it's easier for users to find them to send to @blitzmann .  Ideally, we'd make the checkbox to toggle storing the personal DB in the executable path actually work from inside Pyfa...but that's harder to do.  Right now we're just letting the user know if it's checked or not.
4. New fitting engine preference pane.  This right now only contains the `Factor in Reload Timer` but once we move to the new cap/tank sim engine I'm planning on adding the ability to do other various things, like not factoring in reactivation delay, possibly a stagger toggle, etc.  The idea is to break out the actual fitting engine modifications from the User/UI choices, and provide a little more detailed info so users have a better idea of what they're doing when they click that box. 

@blitzmann this should be ready to be merged.  The only visible impact to the user is additional preference panes, no changes to the stats panes themselves.

### DIFF
--- a/config.py
+++ b/config.py
@@ -18,10 +18,10 @@ debug = False
 saveInRoot = False
 
 # Version data
-version = "1.24.1"
-tag = "git"
-expansionName = "Singularity"
-expansionVersion = "1099916"
+version = "1.25.0"
+tag = "Stable"
+expansionName = "Ascension"
+expansionVersion = "1.9"
 evemonMinVersion = "4081"
 
 pyfaPath = None
--- a/config.py
+++ b/config.py
@@ -18,8 +18,8 @@ debug = False
 saveInRoot = False
 
 # Version data
-version = "1.25.1"
-tag = "Stable"
+version = "1.25.2"
+tag = "git"
 expansionName = "Ascension"
 expansionVersion = "1.9"
 evemonMinVersion = "4081"
--- a/config.py
+++ b/config.py
@@ -18,8 +18,8 @@ debug = False
 saveInRoot = False
 
 # Version data
-version = "1.25.0"
-tag = "Stable"
+version = "1.25.1"
+tag = "git"
 expansionName = "Ascension"
 expansionVersion = "1.9"
 evemonMinVersion = "4081"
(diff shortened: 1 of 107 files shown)


Ebag333:
Note to self.  The Fitting Engine preference pane would be the _perfect_ place for a toggle to show mining lasers with or without crystals.

blitzmann:
I'll take a look at this soon, but just a heads up it's probably not going to make it into release tomorrow. Which is fine, I'd like a bit more time to go over it and maybe discuss ways we can utilize a fitting engine config pane. :)

Ebag333:
I'm not expecting it tomorrow. :)

It really would be nice if this got merged sooner rather than later, so I can build add more preferences for running the Gnosis engine.  Because of the design of Gnosis, we have a _ton_ of control over how we're generating the simulation matrix it returns.  Right now a lot of it is set to my assumptions, but I'd like to give some of that control back to the users.  (To be fair, it's no less control than the current sim.)

blitzmann:
yeah... I haven't even looked at the capsim stuff yet. That particular one is one that I want to get others opinions on as well, so it's gonna take some time (hopefully we can have an answer before the release after this one)

Ebag333:
Rebased this PR, so it's up to date and ready to be merged.

Ebag333:
@blitzmann I pointed this PR at the dev branch.

Ebag333:

Ebag333:
Replaced with #996

PR #886, opened 2016-12-08
Add ability to look at tactical mode item info

Ebag333:
This is very useful for troubleshooting.  We also allow users to view item info on every other "module", so it's a little out of place that we don't for these.

@blitzmann this should be ready to be merged.

### DIFF
--- a/eos/saveddata/mode.py
+++ b/eos/saveddata/mode.py
@@ -53,3 +53,8 @@ class Mode(ItemAttrShortcut, HandledItem):
             for effect in self.item.effects.itervalues():
                 if effect.runTime == runTime and effect.activeByDefault:
                     effect.handler(fit, self, context=("module",))
+
+    def getValidCharges(self):
+        # Modes don't have charges, but it is queried for so return nothing.
+        validCharges = set()
+        return validCharges
--- a/eos/saveddata/mode.py
+++ b/eos/saveddata/mode.py
@@ -53,8 +53,3 @@ class Mode(ItemAttrShortcut, HandledItem):
             for effect in self.item.effects.itervalues():
                 if effect.runTime == runTime and effect.activeByDefault:
                     effect.handler(fit, self, context=("module",))
-
-    def getValidCharges(self):
-        # Modes don't have charges, but it is queried for so return nothing.
-        validCharges = set()
-        return validCharges
--- a/gui/builtinViews/fittingView.py
+++ b/gui/builtinViews/fittingView.py
@@ -26,7 +26,7 @@ import gui.display as d
 from gui.contextMenu import ContextMenu
 import gui.shipBrowser
 import gui.multiSwitch
-from eos.types import Slot, Rack, Module
+from eos.types import Slot, Rack, Module, Mode
 from gui.builtinViewColumns.state import State
 from gui.bitmapLoader import BitmapLoader
 import gui.builtinViews.emptyView
@@ -331,7 +331,7 @@ class FittingView(d.Display):
 
     def removeItem(self, event):
         row, _ = self.HitTest(event.Position)
-        if row != -1 and row not in self.blanks:
+        if row != -1 and row not in self.blanks and isinstance(self.mods[row], Module):
             col = self.getColumn(event.Position)
             if col != self.getColIndex(State):
                 self.removeModule(self.mods[row])
@@ -342,12 +342,7 @@ class FittingView(d.Display):
     def removeModule(self, module):
         sFit = service.Fit.getInstance()
         fit = sFit.getFit(self.activeFitID)
-        try:
-            populate = sFit.removeModule(self.activeFitID, fit.modules.index(module))
-        except ValueError:
-            # This module isn't in our list of modules, don't remove anything. Likely a special snowflake.
-            logger.debug("Failed attempt to remove %s from fit" % module.item.name)
-            populate = None
+        populate = sFit.removeModule(self.activeFitID, fit.modules.index(module))
 
         if populate is not None:
             self.slotsChanged()
@@ -497,20 +492,17 @@ class FittingView(d.Display):
             mod = self.mods[self.GetItemData(sel)]
 
             # Test if this is a mode, which is a special snowflake of a Module
-            if hasattr(mod, "_Mode__item"):
+            if isinstance(mod, Mode):
                 srcContext = "fittingMode"
-                # Skip the normal processing
-                mod.isEmpty = True
 
-                itemContext = sMkt.getCategoryByItem(mod.item).name
+                itemContext = "Tactical Mode"
                 fullContext = (srcContext, itemContext)
                 if not srcContext in tuple(fCtxt[0] for fCtxt in contexts):
                     contexts.append(fullContext)
 
                 selection.append(mod)
 
-
-            if not mod.isEmpty:
+            elif not mod.isEmpty:
                 srcContext = "fittingModule"
                 itemContext = sMkt.getCategoryByItem(mod.item).name
                 fullContext = (srcContext, itemContext)
--- a/gui/builtinViews/fittingView.py
+++ b/gui/builtinViews/fittingView.py
@@ -31,8 +31,11 @@ from gui.builtinViewColumns.state import State
 from gui.bitmapLoader import BitmapLoader
 import gui.builtinViews.emptyView
 from gui.utils.exportHtml import exportHtml
+from logging import getLogger, Formatter
 
 import gui.globalEvents as GE
+
+logger = getLogger(__name__)
 
 #Tab spawning handler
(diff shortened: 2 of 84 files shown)


gui/builtinViews/fittingView.py
```diff
@@ -339,7 +339,11 @@ def removeItem(self, event):
     def removeModule(self, module):
         sFit = service.Fit.getInstance()
         fit = sFit.getFit(self.activeFitID)
-        populate = sFit.removeModule(self.activeFitID, fit.modules.index(module))
+        try:
+            populate = sFit.removeModule(self.activeFitID, fit.modules.index(module))
+        except ValueError:
+            # This module isn't in our list of modules, don't remove anything. Likely a special snowflake.
+            populate = None
```
blitzmann: Instead of this, the user shouldn't see the remove module option for the stance to begin with. Revert this and add that plzkthx

eos/saveddata/mode.py
```diff
@@ -53,3 +53,8 @@ def calculateModifiedAttributes(self, fit, runTime, forceProjected=False):
             for effect in self.item.effects.itervalues():
                 if effect.runTime == runTime and effect.activeByDefault:
                     effect.handler(fit, self, context=("module",))
+
+    def getValidCharges(self):
+        # Modes don't have charges, but it is queried for so return nothing.
+        validCharges = set()
+        return validCharges
```
blitzmann: Curious as to where this gets called for stances?

gui/builtinViews/fittingView.py
```diff
@@ -488,12 +491,26 @@ def spawnMenu(self):
 
         while sel != -1 and sel not in self.blanks:
             mod = self.mods[self.GetItemData(sel)]
+
+            # Test if mod.isEmpty exists.
+            try:
+                mod.isEmpty
+            except AttributeError:
+                mod.isEmpty = False
```
blitzmann: I'd rather have something like `if hasattr(mod, 'isEmpty') and not mod.isEmpty` rather than try catch here. Also, make a small note that simply says that t3d modes don't have this property, otherwise I'll be looking at this 2 years from now wondering why the fuck it's here lol

gui/builtinViews/fittingView.py
Ebag333: it's there for double click (also, sanity check in case anything else gets added that isn't a module). I'll check and make sure modes don't have a remove option in the submenu. 

I think that we need it regardless, it fails to the state we want (not doing anything), we are just trapping the stacktrace here.

gui/builtinViews/fittingView.py
Ebag333: both good thoughts.

eos/saveddata/mode.py
Ebag333: probably because stances are just an item and often get treated as such. I can generate a stacktrace and find where it's called.

Ebag333:
```
  File "C:\Python27\lib\site-packages\wx-3.0-msw\wx\_core.py", line 16766, in <lambda>
    lambda event: event.callable(*event.args, **event.kw) )
  File ".\gui\builtinViews\fittingView.py", line 531, in spawnMenu
    menu = ContextMenu.getMenu(selection, *contexts)
  File ".\gui\contextMenu.py", line 69, in getMenu
    if m.display(srcContext, selection):
  File ".\gui\builtinContextMenus\moduleAmmoPicker.py", line 31, in display
    currCharges = mod.getValidCharges()
AttributeError: 'Mode' object has no attribute 'getValidCharges'
```

`contextMenu` just loops through a set number of things.  We could do a  special exception in there to ignore modes, but seems easier, better, and more standardized to simply return an empty `getValidCharges` on modes, which is what other items do if they have no charges.

That way if CCP ever mucks with modes and adds charges to it, or we get other special modules, less special exceptions and easier to handle.

gui/builtinViews/fittingView.py
Ebag333: Okay, so we no longer show the remove option in the menu.  So used to seeing it, that it looked normal with it there and weird without it.

We still need this try/catch though, because of double click to remove.  Like I said, when it exceptions because of this it fails in the way we want it to (keeps going and doesn't remove anything), but I think it's cleaner to catch the exception and handle it.

I can add debug logging if we want to record these scenarios.

gui/builtinViews/fittingView.py
```diff
@@ -492,10 +492,11 @@ def spawnMenu(self):
...
 
-            # Test if mod.isEmpty exists.
-            try:
-                mod.isEmpty
-            except AttributeError:
+            # Test if mod.isEmpty does not exist.
+            # Certain special module can be missing this trait
+            # Example: T3D modes
+            if not hasattr(mod, 'isEmpty'):
+                # Set it if missing, prevents later stack traces.
```
blitzmann: ha, I didn't mean set it if `hasattr` evaluated to false. I meant to not even assign the property `isEmpty` and use the `hasattr` like so:

`if hasattr(whatever) and not mod.isEmpty:`

This way we aren't setting a property that shouldn't be set. `isEmpty` applies to Modules only, not modes. No reason to have it around.

gui/contextMenu.py
```diff
@@ -69,6 +69,17 @@ def getMenu(cls, selection, *fullContexts):
...
+
+                    # Check the selected item to see if it has special reasons for not showing the menu item
+                    skip_menu_item = False
+                    for sel in selection:
+                        if hasattr(sel, "_Mode__item") and texts == "Remove Module":
+                            # Don't show remove for modes, these are special modules that cannot be removed
+                            skip_menu_item = True
+
+                    if skip_menu_item == True:
+                        continue
```
blitzmann: This is not the right way to do it. :) Look in `builtinContextMenus/remove.py` or whatever it's called. That file should have a method that determines if that context menu item displays or not. You can call an `isinstance()` to determine if it shows or not (you may have to import the type from eos depending on current imports). Actually, it might be better to have the context menu context set as `fittingMode` instead of `fittingModule` - that way you can easily determine which context menus are displayed depending on the context and don't have to worry about importing type definitions. You can then bypass the entirety of the whole `isEmpty` thing and set a new context for modes.

By the way, for future reference, python has a wonderful `for...else` ability. It basically replaces things where you need a boolean flag in a for loop.

blitzmann:
@Ebag333 I never know if these things ping you, so I'm doing it. I made some more comments. :)

eos/saveddata/mode.py
blitzmann: It still doesn't feel right to me. If modes ever get charges, then we simply add it (at that point we would probably just inherit from Module). Having it defined here for no purpose other than to work around a bug makes me thing we should conditionally account for it where it's called.

Also, going the route I suggested in another comment about giving the mode it's own context menu... context (*cringe*) would solve this, as we can then selectively determine if we even run the offending code (the module ammo picker). The current problem is that the context menu sees the mode as a module, and all modules are assumed to have charge properties. And this context menu runs because it things the item is a module, and thus assumes it has all the module stuff. Having a separate context for modes would be the best way to go IMO, and it's not difficult at all to do. :)

Let me know if you need additional help about it :D

Ebag333:
@blitzmann sometimes it does sometimes it doesn't.  vOv

Implemented a fitting mode.  It's set in `fittingView` and then we can use that to skip adding `Remove` to the menu.

Think that's what you wanted. :)

gui/builtinViews/fittingView.py
Ebag333: Well, if we wanted to treat modes like a module it's required because it's used several places. vOv

Anyway, implemented `fittingMode` so now (at least for the purpose of context menus) modes are no longer items.

blitzmann:
Fuck it, just going to reply here.

> Well, if we wanted to treat modes like a module it's required because it's used several places. vOv

I don't want to treat modes like modules. That's the point. They aren't a module. The show up in the fittin window s due solely to the fact that there's no better place for it, but they aren't modules and should not have "module properties" so to speak

> modes are no longer items. 

I don't know what this means. Ofc they are items. =/

Also, you shouldn't have to touch `gui/contextMenu.py`. You misunderstand how the context menu stuff works. When we spawn a context menu, see send it a "context" - basically it's information pertaining to the source of the spawn. So in our situation, we want to create a `fittingMode` context so that we can easily toggle which menu items are triggers and which aren't. The context menus themselves are a plug in system - they get imported and registered to the context menu class. When a menu spawns, it loops through each of them and calls a method in the class to determine if that menu displays or not.

In the change you did, you created `fittingMode` but then you also added it to the criteria o display in `itemRemove.py::display()` and then hardcoded for it not to display in `contextMenu.py`. All that should happen here is simply not adding `fittingMode` to that display criteria.

I will clean it up and try to push to your branch so that you can better review what I mean, then I'll merge it.

PR #890, opened 2016-12-11
fixed parameters in call when adding neuts to structures.

Drezil:
When adding Neuts to an Engeneering-Complex you get the error that "float" has no attribute xx..

It seems like "container" was not passed by accident, as indicated by the firstline-comment that it was not used previously.

HTH

### DIFF
--- a/eos/effects/structureenergyneutralizerfalloff.py
+++ b/eos/effects/structureenergyneutralizerfalloff.py
@@ -10,4 +10,4 @@ def handler(fit, container, context):
                                     and container.state >= State.ACTIVE) or hasattr(container, "amountActive")):
         amount = container.getModifiedItemAttr("energyNeutralizerAmount")
     time = container.getModifiedItemAttr("duration")
-    fit.addDrain(time, amount, 0)
+    fit.addDrain(container, time, amount, 0)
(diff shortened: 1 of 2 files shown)


Ebag333:
Huh. This shouldn't work then.  `addDrain` requires 3 parameters (4th is optional).  It shouldn't work for any structure, not just engineering complexes.

Nice catch. 👍

Drezil:
and found an off-by-one index when adding boosts onto other ships... (in my case: Orca mining-boosting a venture)...

probably an artifact as it is very rarely called and the first index in the tuple is always a string and not the value which it was compared against.

Error was: abs() could not handle type "str".

Ebag333:
@Drezil can you revert that last commit?  Blitzman is actively working on this one, and might have already caught this.

@blitzmann can you take a look at this?  Looks like he's correct to me, but you might have already fixed this and know more about how you setup command bursts than I do.

```
-        if warfareBuffID not in self.commandBonuses or abs(self.commandBonuses[warfareBuffID][0]) < abs(value):
 +        if warfareBuffID not in self.commandBonuses or abs(self.commandBonuses[warfareBuffID][1]) < abs(value):
``

blitzmann:
Thanks @Drezil . No worries about reverting and submitting two different ones in this case, but for future reference that would be the way tog o :)

Issue #379, opened 2015-10-10
Cannot start on custom locale when installation path contains non-English characters


WiseFerret:
My pyfa is installed in /home/yotam/הורדות/pyfa.

Trying to bypass issue #378 (https://github.com/DarkFenX/Pyfa/issues/378), I try to force 
English locale. This is what I get:

```
yotam@ape:~/הורדות/pyfa$ LC_ALL=C ./pyfa.py
Traceback (most recent call last):
  File "./pyfa.py", line 79, in <module>
    config.defPaths()
  File "/home/yotam/הורדות/pyfa/config.py", line 69, in defPaths
    sys.modules['__main__'].__file__))), sys.getfilesystemencoding())
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd7 in position 12: ordinal not in range(128)
```

When I install pyfa in my home folder (/home/yotam/pyfa), which does not include non-English characters, the same commands launches pyfa in the desired locale perfectly.

blitzmann:
I thought we fixed this like last year. Will look into it. In the meantime... run via an ASCII path

blitzmann:
Can you try `LC_ALL=en_US.UTF-8` instead? Using `LC_ALL=C` did not work for me (I assume because out files are written in ASCII, not Unicode? I'm not too familiar with locales and how they inheritely work)

WiseFerret:
When I run LC_ALL=en_US.UTF-8 ./pyfa.py the program loads fine even when the path contains non-ascii characrters; however, it suffers from the same issue described in issue #378. The interface is right-to-left, and is broken.
It seems that LC_ALL=C is needed on right-to-left locales.

blitzmann:
Specifically what is your locale set to?

WiseFerret:
LANG=he_IL.UTF-8
LANGUAGE=he_IL:he
LC_CTYPE="he_IL.UTF-8"
LC_NUMERIC=he_IL.UTF-8
LC_TIME=he_IL.UTF-8
LC_COLLATE="he_IL.UTF-8"
LC_MONETARY=he_IL.UTF-8
LC_MESSAGES="he_IL.UTF-8"
LC_PAPER=he_IL.UTF-8
LC_NAME=he_IL.UTF-8
LC_ADDRESS=he_IL.UTF-8
LC_TELEPHONE=he_IL.UTF-8
LC_MEASUREMENT=he_IL.UTF-8
[...]

PR #910, opened 2016-12-19
Unicode Paths 2: The Reckoning

Ebag333:
See #854 for history.

@blitzmann if you want to keep the original PR you should be able to cherry pick the second commit on this one.

This has been tested (on Windows) against: 
- `Téléchargements` works
- `הורדות` (Hebrew) - pyCharm itself chokes on this one. Pyfa doesn't like it either: 
`Missing icon file: C:\Users\Ebag333\Documents\GitHub\Ebag333\??????\imgs\gui\pyfa.png`

For testing the one that seems to break most reliably is the cert path, so you can simply drop the check in `config` from:
```
if isFrozen():
    certName = "cacert.pem"
    os.environ["REQUESTS_CA_BUNDLE"] = getPyfaPath(certName).encode('utf8')
    os.environ["SSL_CERT_FILE"] = getPyfaPath(certName).encode('utf8')
```

This may be helpful:
https://docs.python.org/2/library/codecs.html

### DIFF
--- a/config.py
+++ b/config.py
@@ -31,18 +31,20 @@ gameDB = None
 
 
 class StreamToLogger(object):
-   """
-   Fake file-like stream object that redirects writes to a logger instance.
-   From: http://www.electricmonk.nl/log/2011/08/14/redirect-stdout-and-stderr-to-a-logger-in-python/
-   """
-   def __init__(self, logger, log_level=logging.INFO):
-      self.logger = logger
-      self.log_level = log_level
-      self.linebuf = ''
+    """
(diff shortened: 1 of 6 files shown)


blitzmann:
#379 having a user test hebrew path. Once we get an okay from them, we can probably merge this in.

blitzmann:
User mentioned in issue that it works with hebrew on linux. I would assume it also works on OS X considering it's super similar to linux.

Will do a quick test myself and merge this into dev this weekend. Thanks!

Ebag333:
This might make some Russians happy. 😁

PR #913, opened 2016-12-20
Remembers user's meta filtering choices

resinneublem:
There is a lot of logic that overwrites the user's choices of meta
filtering on every search.

That's a little too clever. Especially if the user is looking for
different variations of a certain meta.

If a user wants to include the other meta buttons they can re-add them
after filtering.

Extra care has to be taken with Windows. If a button is Enabled=False AND
SetValue=True, then it looks like it's Enabled=True, but clicking it
doesn't do anything. So we handle that logic in it's own class

closes: https://github.com/pyfa-org/Pyfa/issues/818
take 2 of: https://github.com/pyfa-org/Pyfa/pull/829

---

This branch is against the current Pyfa master. To quickly see the difference between my last branch and this branch you can run: `$ git diff cfa57a70dc17a8f80bff2e354498372fa42534cd -- gui/marketBrowser.py`

### DIFF
--- a/gui/marketBrowser.py
+++ b/gui/marketBrowser.py
@@ -31,6 +31,25 @@ ItemSelected, ITEM_SELECTED = wx.lib.newevent.NewEvent()
 RECENTLY_USED_MODULES = -2
 MAX_RECENTLY_USED_MODULES = 20
 
+class MetaButton(wx.ToggleButton):
+    def __init__(self, *args, **kwargs):
+        super(MetaButton, self).__init__(*args, **kwargs)
+        self.setUserSelection(True)
+
+    def setUserSelection(self, isSelected):
+        self.userSelected = isSelected
+        self.SetValue(isSelected)
+
+    def setMetaAvailable(self, isAvailable):
+        self.Enable(isAvailable)
+        # need to also SetValue(False) for windows because Enabled=False AND SetValue(True) looks enabled.
+        if not isAvailable:
+            self.SetValue(False)
+
+    def reset(self):
+        self.Enable(True)
+        self.SetValue(self.userSelected)
+
 class MarketBrowser(wx.Panel):
     def __init__(self, parent):
         wx.Panel.__init__(self, parent)
@@ -62,7 +81,7 @@ class MarketBrowser(wx.Panel):
         vbox.Add(p, 0, wx.EXPAND)
         self.metaButtons = []
         for name in self.sMkt.META_MAP.keys():
-            btn = wx.ToggleButton(p, wx.ID_ANY, name.capitalize(), style=wx.BU_EXACTFIT)
+            btn = MetaButton(p, wx.ID_ANY, name.capitalize(), style=wx.BU_EXACTFIT)
             setattr(self, name, btn)
             box.Add(btn, 1, wx.ALIGN_CENTER)
             btn.Bind(wx.EVT_TOGGLEBUTTON, self.toggleMetaButton)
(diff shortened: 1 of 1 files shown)


gui/marketBrowser.py
```diff
@@ -94,7 +110,7 @@ def toggleMetaButton(self, event):
             # Do 'nothing' if we're trying to turn last active button off
             if len(activeBtns) == 1 and activeBtns.pop() == ebtn:
                 # Keep button in the same state
-                ebtn.SetValue(True)
```
resinneublem: @blitzmann I really didn't feel good about all the `SetValue` and `Enabled` calls scattered across this file. So I specialized the ToggleButton and made it pretty clear what we want to do, when we want to do it.

I tested this on Ubuntu and Windows and it works pretty well. macOS behaved like Ubuntu in the first round, so I don't anticipate any problems there.

resinneublem: Also tested on macOS Sierra. I'm starting to look into the copy/paste issue and the tooltips and verified figured I would verify this.

blitzmann:
Something seems to be broken. Reproduction:

1) Start pyfa
2) Navigate to Ship Equiptment > Propulsion > Afterburners
3) Be default everything is selected. Click on one filter to only select that one, and the ctrl+click on all other filters to select all again
4) After this, search "200". The first filter that you selected to re-enable everything is used, and all others are deselected. 

This is especially bad if you select Complex first, then ctrl+click all others, and then search "200". There are no Complex modules that meet the 200 search criteria, and so nothing shows up.

Let me know if you need more details

resinneublem:
Hrm. I can't seem to get this right can I?

Investigating again.

resinneublem:
Yep, I can reproduce. Thanks man.

resinneublem:
@blitzmann take a look at the latest commit. The problem was that I wasn't setting the user selection in every branch.

I refactored the click handler to make it easier to understand and click tested it out. Seems to work out well.

There is one user experience issue I think we should talk about though:

1. Open app
2. Click Market > Ship Equipment > Propulsion > Afterburners
3. Filter by Officer
4. Click Micro Jump Drives

There are no Officer Micro Jump Drives so nothing shows up. But the Officer fitter button is disabled so it's a little ambiguous to the user why they're not seeing anything.

selection_002

I don't think that's a problem with this logic, I think it's just a gap with the existing design. Some thoughts:

* Add a 5th "All" button. That might change up the some of the Ctrl-Click behavior though.
* Display a note in the list that we're filtering out "x other items".
* Probably a better design idea. I'm not a great designer.

---

Normally I'd squash these commits into one since the previous commit is faulty, but I just added a new one after our last conversation.

blitzmann:
Would there be a way to reset the filter if it goes disabled? And if it's the only one set, then revert back to select all (or maybe revert to normal)? This might be the reason why it was decided to reset the filters to all when searching. I don't know of a good way to handle this atm. =/

resinneublem:
@blitzmann I honestly think this is more of a design/user communication issue where the existing button layout is restraining what we're trying to accomplish.

The issue as I see it is: The user tells us what they care about (only certain meta groups), and sometimes we show them an empty list based on their choices (either clicking categories in the tree view or searching). Instead of ignoring their wishes (by resetting the filters for them), we should better communicate that there are other items that they may be ignoring.

One other reason why I think resetting is less desirable is because on the 3rd character of a search, we immediately search for items. But sometimes the intermediary search is more restrictive than the final search terms. If someone wasn't sure if it was "repeating cannon" and instead typed "cannon" (only Normal and Faction) then "cannon repeating" (Normal, Faction, Officer) we would be ignoring their request to filter only officer items.

Another design idea is to have a "badge" with each button listing the number of items per meta group that is found. So if they search for "Micro Jump" and don't see any teams, at least they would see "Normal (3)".

I don't see this as a deal breaker for this PR being merged. But I think the discussion is worthwhile.

blitzmann:
Not sure what the best action to take is tbh. If we can easily get it to say "No results based on filters" or something that might be helpful, but not a priority, and something tells me it might be a pain in the ass to do. 

If you want to tackle that in a separate PR go for it. I'll go ahead and merge this one. There may be a couple edge cases that might cause some confusion, but I think we can address them when users complain - they may have more ideas on how to handle those particular cases as well. :)

Issue #899, opened 2016-12-12
paste into notes


epmd34:
When the notes pane is active copy,paste cut etc should act on the clipboard as text, not try to import / export fittings.

Ebag333:
Unfortunately we hijack copy/paste, because of the import/export fittings option.  It's been that way for years, and it's set very low level.  I'm not sure we can override that, and if we do we're changing the default behavior of the whole app (ctrl-c does one thing most places, something else in on specific case, that sort of thing tends to be bad UI and confusing to end users).

You can right click to bring up the context menu (at least in Windows, not sure on Linux/Max).

blitzmann:
We need to fix this for #490 as well, sooner rather than later.

We should be able to specifically define C&P shortcuts on a per-widget level. We want to keep the global shortcut for fittings, but this definitely needs to be fixed when we need it.

resinneublem:
@epmd34 What platform are you running in? OSX?

epmd34:
OSX

blitzmann:
@epmd34  Ah, yeah, then that's the problem. I thought this was a Windows bug, so this is really a duplicate of #490. Still need to find a solution for this tho...

DarkFenX:
Closed as duplicate of #490 (which should be fixed starting from next release)

Issue #887, opened 2016-12-10
Paste from clipboard doesn't work when entering API for new character (OSX)


matthewpennell:
In the current OSX build, on adding a new character, paste (Cmd+V or Edit->Paste) does not work in the API key and verification code fields. This means that you cannot import a character's skills without retyping both strings manually.

blitzmann:
Duplicate #490

Unfortunately it's difficult for me at this point to do anything or develop on OSX. I will try to make this a priority for the next big release though.

Issue #877, opened 2016-12-04
Importing api key / Verification code for new char is not working on mac


hugohase:
hi,

i try to create a new char with the char editor and try to ctrl c / ctrl v cmd c / cmd v the api infos from the website but the input fields don't accept this.

if i try to copy a string from an existing char the popup opens for the format and it exports a fit not the key.

blitzmann:
Probably a duplicate of #490. Haven't had a chance to fix it yet ;_;

Issue #490, opened 2016-01-02
Conflict with Copy/Paste shortcuts


paxswill:
When I'm trying to add an API key to Pyfa, I copy the info from CCP's Api key site, but I'm unable to use ⌘V (command-V) to paste in the key ID or vCode as that is in use by the "From Clipboard" command. Could those commands either have their shortcuts disabled while the API editing window is focused (probably hacky and leaves other corner cases unsolved) or possibly change the "To Clipboard" and "From Clipboard" to ⇧⌘V/⇧⌘C (shift-command-V/C). I am still able to right-click the text field and paste from there, so adding help text to tell people to do that might be a quick fix as well.

blitzmann:
Are you using the basic `-mac` version or the `-mac-wx3` version, and what is your OSX version? This is the first I'm hearing of the shortcut completely being overridden rather than the intended added functionality. I'll have to look into this more

paxswill:
-mac-wx3, 10.10.5

resinneublem:
Yeah I can reproduce too. Will do some googling to see about a work around.

resinneublem:
As a short term work around you can right click and choose "Paste" on the text field.

There may be a way to disable the menu items but I don't know you'd be able to deterministically tell when to disable/enable.

RealSalmon:
This is also an issue with the -mac version in v1.18.0 (OS X 10.10.5)

blitzmann:
I still have not had a change to check, but this may be related to the fix for the menu as stated here: https://github.com/pyfa-org/Pyfa/issues/456#issuecomment-161885962

blitzmann:
I can confirm this is present in current release. With the current master it works, but probably due to a recent switch from Frame to Dialog that may or may not be sticking around. Anyway, this issue here is event propagation - need to find a way to stop paste events from being handled by the main window if they are happening in a text ctrl.

resinneublem:
I did some digging into this, and it looks like the problem is https://github.com/pyfa-org/Pyfa/commit/0a10f2d72ad82a383667ebe2a1f15491c48c6f87

Specifically the override https://github.com/pyfa-org/Pyfa/blob/master/gui/mainMenuBar.py#L54-L56

If I comment that out, copy/paste works just fine. When you're using a text control, it behaves as normal (even though the Edit menu highlights). And when you're not in a text control, it will import/export the fit.

[...]

PR #915, opened 2016-12-20
Stop overriding copy/paste on macOS

resinneublem:
For some reason this causes issues on wxMac 3.0.2.0. The menu accelerators hijack the shortcut and won't allow typical text manipulation inside of text controls.

There's some discussion about this happening for other people in:

* http://trac.wxwidgets.org/ticket/11320
* http://trac.wxwidgets.org/ticket/14953

I have not tested this in wxMac 2.8

This is to fix:

* https://github.com/pyfa-org/Pyfa/issues/490
* https://github.com/pyfa-org/Pyfa/issues/899
* https://github.com/pyfa-org/Pyfa/issues/877
* https://github.com/pyfa-org/Pyfa/issues/887

But I'm not sure if it reverts the fix for: https://github.com/pyfa-org/Pyfa/issues/456#issuecomment-161885962

### DIFF
--- a/gui/mainMenuBar.py
+++ b/gui/mainMenuBar.py
@@ -51,10 +51,6 @@ class MainMenuBar(wx.MenuBar):
         self.toggleOverridesId = wx.NewId()
         self.importDatabaseDefaultsId = wx.NewId()
 
-        if 'wxMac' in wx.PlatformInfo and wx.VERSION >= (3,0):
-            wx.ID_COPY = wx.NewId()
-            wx.ID_PASTE = wx.NewId()
-
         self.mainFrame = gui.mainFrame.MainFrame.getInstance()
         wx.MenuBar.__init__(self)
 


resinneublem:
Hrm, I can't even open the `mac-deprecated` version on macOS Sierra. The `mac` version opens just fine.

I ran: `python dist.py -s ../../Pyfa-skel -b ../../Pyfa -d ../../ -z` after cloning the Pyfa-skel repository.

I don't know what I'm doing wrong since the latest stable release downloaded from github opens just fine.

resinneublem:
Hrm strange. Something happened and now I can't open the latest stable `mac-deprecated` now.

I blasted `~/.pyfa` and tried again. Still fails.

I wonder if anyone else can open it. I don't have a non-Sierra machine to test.

blitzmann:
Yeah, known issue. See #780

blitzmann:
Could you run the `pyfa.app/Contents/MacOS/pyfa` binary from terminal and see if anything shows?

resinneublem:
> Could you run the pyfa.app/Contents/MacOS/pyfa binary from terminal and see if anything shows?

Immediately segfaults.

blitzmann:
> Immediately segfaults.

Yeah, and unfortunately I have no knowledge of how to handle those. =/ `deprecated` is basically dead in the water for 10.12+, which wouldn't be so bad if 10.11 didn't completely fuck performance for the regular wx3 build...

I'll be making build packages and testing this pull request soon, haven't forgotten about it :)

blitzmann:
@resinneublem I haven't forgotten about this, I just still need to test it, especially to see hwo it interacts with the other issues. But we got a report in #954 that the new pre-release fixes the mac-deprecated version, can you test as well? 

Thanks!

blitzmann:
Unfortunately this isn't going to work... We have the following situation (tested on 10.7... yes I know this is extremely old)

(current master)
pyfa (wx2): Shortcuts overwrite default event. CMD+V in fit notes will try to import a fit from the clipboard, while CMD+C will not copy the current text but instead bring up the fit export dialog
pyfa (wx3): Exact same as above

(with this fix)
pyfa (wx2): Same as the above two
pyfa (wx3): CMD+V/C work as expected in text ctrls, but the to and from clipboard actions are disabled (reoccurance of #456)

We may just have to change the shortcuts for OS X to be something like SHIFT+CMD+C. At the same time, I'm going to try to make a real push in developing on OS X and making new binaries that don't suck for 10.11+, and hopefully getting rid of the wx2.8 build so we don't have to support it.

Unless we can find a way to only attach these shortcuts when a text control is not currently in focus...

blitzmann:
I tested another version of this last night with no success. I'm going to close this PR since I don't think it's going to work. I did note (and this is in  a comment on one of the issues) that this only happens in the packaged release - works fine when running through pycharm. 

I don't know what the problem is, but I'm going to re-evaluate when I rebuild the binaries (either through py2app or the outstanding request to use pyinstaller).

Issue #917, opened 2016-12-21
Can't create new character


tmockler69:
On ver 1.26.0, every time I attempt to add a new character, nothing happens.

I enter the character name, and click OK, the dialog box disappears, but the character is not added to the list.

Am I doing something wrong?

Ebag333:
Probably not.  Can you upload your user database somewhere and email me a link? ebagola@gmail.com

tmockler69:
It's a brand new installation....I'm guessing the user database is empty.

What folder would I find the file you need?

Ebag333:
So brand new, never used before?

Here's how you create chars:
https://puu.sh/sWVeX/ed842b8dda.gif

If that's not working, then something is clearly wrong.  Are you running under Windows, Linux, Mac?

tmockler69:
Yeah, something is clearly wrong.  As I said before, I click on the new character button, enter the name, click OK, the dialog box disappears, but the new name is not added to the list.

I'm running Windows 10

Ebag333:
Try opening the folder you installed Pyfa into, right click, `Run as Administrator`.  See if that makes a difference.

If that doesn't, create a shortcut for the EXE, and add this to the shortcut (under right click, properties):
`-debug`

That should output some information to `c:\Users\<YourUsername>\.pyfa` (which is also where your user database goes).

tmockler69:
Run as Adminstrator does not make a difference.

Tried to add -debug to the command line on the shortcut and received the following dialog:

![capture](https://cloud.githubusercontent.com/assets/17017122/21407246/8244c0a8-c79d-11e6-96f3-c7c1b08acf69.JPG)

Ebag333:
Can you try it with `-debug` outside of the quotes?

blitzmann:
BTW It's `-d`, or `--debug`

tmockler69:
OK....the debug switch is now working

blitzmann:
That being said, even if it's a brand new install... there have been times when people "try out" pyfa and then drop it, only to reinstall it months / years later. That user database can persist that long as long as you don't manually clean it up. 

Please send me a copy of `%userprofile%\.pyfa\saveddata.db` to the email in the README. Many times, characters not adding is due to some sort of database corruption.

OR if it truly is a new install, then you don't really care about the database anyway. Delete the database and try to launch pyfa again (I would still like a copy of the database before you nuke it so that I can try to analyze what went wrong and maybe prevent this in the future)

:)

[...]

Issue #777, opened 2016-10-20
Database corruption issue


Ebag333:
As per #773 (and many others), users are getting corrupted databases, where either data is getting out of sync or there's actual corruption (missing/bad data).

Need to do some investigating on this, and see about doing some standard database repairs......

Ebag333:
We started down this road with the implementation of the default profiles.

Can likely extend this methodology easily to other things (chars).

I have personally seen where you can end up with a "missing" damage profile, so you always show HP (even if EHP is toggled) until a new damage profile is selected.

blitzmann:
Agreed, we can do more about database maintenance. At the very least, allow the user to run an integrity check manually.

I think we should look at the current constraints. Things like character skills being orphaned shouldn't happen as we cascade the delete within a transaction. I don't know why some things break like that.

As for #773, I don't think it's database corruption. Symptoms don't make sense to me.

Ebag333:
>I think we should look at the current constraints. Things like character skills being orphaned shouldn't happen as we cascade the delete within a transaction. I don't know why some things break like that.

SQLite doesn't have proper DB constraints like MS SQL or MySQL would, so you are allowed to do things that full blown DBs would prevent you from doing.  (But hey, at least it's not Mongo.)

There _are_ a number of things we don't seem to take advantage of though.  Like flagging a column as being unique, limited amount of foreign keys set, and there _is_ a check constraint...but those can be dangerous.  I also have no idea how SQLAlchemy handles those (though it should handle at least two of those three just fine).

Foreign keys are particularly interesting because they're very hit and miss what's set and what's not.  Those can do a lot to help prevent bad data from sneaking in.

Ebag333:
>I have personally seen where you can end up with a "missing" damage profile, so you always show HP (even if EHP is toggled) until a new damage profile is selected.

Well, figured this out.

We assume that `Uniform` will be damage pattern 1.  But what if it's recreated at some point in time?  (Or, for that matter, any damage pattern.)
https://puu.sh/rP8Fq/6c53fd18e7.png

We should do validation on launch, if the damage pattern ID is set but doesn't exist, we can handle it one of two different ways.

1) Change it to the (new/current) uniform damage pattern ID.
2) Nullify the field.

I vote for option 2, and then change the damage pattern code so that if the damage pattern lookup is null we set it to uniform.

blitzmann:
>But what if it's recreated at some point in time?

[...]

PR #920, opened 2016-12-27
Database Validation and Cleanup

Ebag333:
This addresses three issues:

1. Character skills exist for a character that doesn't.  See #917
2. Damage pattern on a fit is set to a damage pattern that doesn't exist.  See #777
3. Character on a fit is set to a character that doesn't exist. (Can't find which issue mentioned this.)

@blitzmann I know that you wanted to avoid running database validation and cleanup each time, but I don't think it's avoidable.  We are running into scenarios where people can't even launch Pyfa, so they can't get to a point when they _could_ click a button to fix it.

And yes, I know that the first two issues won't cause that scenario.  But since a lot of users won't ever take the time to post here or on the forums about it (just rage quit and go to another tool), I think that this is the best approach.

Plus, this takes basically nil time.  For a clean DB:
```
2016-12-26 21:39:18,069 root                     DEBUG    Starting database validation.
2016-12-26 21:39:18,072 eos.db.saveddata.databaseRepair DEBUG    Start databsae validation and cleanup.
2016-12-26 21:39:18,075 eos.db.saveddata.databaseRepair DEBUG    Running database cleanup for character skills.
2016-12-26 21:39:18,082 eos.db.saveddata.databaseRepair DEBUG    Running database cleanup for orphaned characters attached to fits.
2016-12-26 21:39:18,089 eos.db.saveddata.databaseRepair DEBUG    Running database cleanup for orphaned damage patterns attached to fits.
2016-12-26 21:39:18,098 root                     DEBUG    Completed database validation.
```

Takes all of 26 milliseconds and likely most of that is from the actual logging.

Ran against my personal DB (where I figured out the uniform damage profile issue), only took literally a handful of milliseconds more to clean up hundreds of records.

Now, this doesn't _fix_ the root issue.  But since we don't know what causes it exactly, and we have people who are currently broken and non-functional, we might as well clean it up to get them back on their feet.

Broke this out into it's own class so we can easily extend it with more checks.

### DIFF
--- a/eos/db/saveddata/databaseRepair.py
+++ b/eos/db/saveddata/databaseRepair.py
@@ -37,17 +37,13 @@ class DatabaseCleanup:
         # See issue #917
         try:
             logger.debug("Running database cleanup for character skills.")
-            results = saveddata_engine.execute("SELECT * FROM characterSkills "
+            results = saveddata_engine.execute("SELECT COUNT(*) AS num FROM characterSkills "
                                                "WHERE characterID NOT IN (SELECT ID from characters)")
 
-            # Count how many records exist.  This is ugly, but SQLAlchemy doesn't return a count from a select query.
-            result_count = 0
-            for _ in results:
-                result_count += 1
+            if results.fetchone()['num'] > 0:
+                delete = saveddata_engine.execute("DELETE FROM characterSkills WHERE characterID NOT IN (SELECT ID from characters)")
+                logger.error("Database corruption found. Cleaning up %d records.", delete.rowcount)
 
-            if result_count > 0:
-                logger.error("Database corruption found. Cleaning up %d records.", result_count)
-                saveddata_engine.execute("DELETE FROM characterSkills WHERE characterID NOT IN (SELECT ID from characters)")
         except sqlalchemy.exc.DatabaseError:
             logger.error("Failed to connect to database.")
 
@@ -58,14 +54,9 @@ class DatabaseCleanup:
         # See issue #777
         try:
             logger.debug("Running database cleanup for orphaned damage patterns attached to fits.")
-            results = saveddata_engine.execute("SELECT * FROM fits WHERE damagePatternID not in (select ID from damagePatterns)")
+            results = saveddata_engine.execute("SELECT COUNT(*) AS num FROM fits WHERE damagePatternID not in (select ID from damagePatterns)")
 
-            # Count how many records exist.  This is ugly, but SQLAlchemy doesn't return a count from a select query.
-            result_count = 0
-            for _ in results:
-                result_count += 1
-
-            if result_count > 0:
+            if results.fetchone()['num'] > 0:
                 # Get Uniform damage pattern ID
                 uniform_results = saveddata_engine.execute("select ID from damagePatterns WHERE name = 'Uniform'")
 
@@ -80,10 +71,10 @@ class DatabaseCleanup:
                 elif uniform_result_count > 1:
                     logger.error("More than one uniform damage pattern found.")
                 else:
-                    logger.error("Database corruption found. Cleaning up %d records.", result_count)
-                    saveddata_engine.execute("UPDATE 'fits' SET 'damagePatternID' = ? "
+                    update = saveddata_engine.execute("UPDATE 'fits' SET 'damagePatternID' = ? "
                                              "WHERE damagePatternID NOT IN (SELECT ID FROM damagePatterns)",
                                              uniform_damage_pattern_id)
+                    logger.error("Database corruption found. Cleaning up %d records.", update.rowcount)
         except sqlalchemy.exc.DatabaseError:
             logger.error("Failed to connect to database.")
 
@@ -92,14 +83,10 @@ class DatabaseCleanup:
         # Find orphaned character IDs. This solves an issue where the chaaracter doesn't exist, but fits reference the pattern.
         try:
             logger.debug("Running database cleanup for orphaned characters attached to fits.")
-            results = saveddata_engine.execute("SELECT * FROM fits WHERE characterID NOT IN (SELECT ID FROM characters)")
+            results = saveddata_engine.execute("SELECT COUNT(*) AS num FROM fits WHERE characterID NOT IN (SELECT ID FROM characters)")
 
             # Count how many records exist.  This is ugly, but SQLAlchemy doesn't return a count from a select query.
-            result_count = 0
-            for _ in results:
-                result_count += 1
-
-            if result_count > 0:
+            if results.fetchone()['num'] > 0:
                 # Get All 5 character ID
                 all5_results = saveddata_engine.execute("SELECT ID FROM characters WHERE name = 'All 5'")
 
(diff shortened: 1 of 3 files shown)


blitzmann:
Doesn't seem too bad. Eventually I would only want this to run if we ever run into a sql exception, which would require messing with custom exception handlers. Buy for now this should do.

Also, your record counting... there should be a much easier way. Len(results) should work?

Ebag333:
>Also, your record counting... there should be a much easier way. Len(results) should work?

You'd think, wouldn't you?

`{TypeError}object of type 'ResultProxy' has no len()`

Searching for it seems that there's no real good solutions for it.  Some folks write custom functions to do it, but it's more complicated that need be IMHO.

So the various `.count`, `len`, etc don't work for this.  Could convert the object type to a dict, but that'd basically end up being just as complex (really more so).

It's not pretty, but it takes a fraction of a millisecond to do (even if there's hundreds of records), and most times there won't be any records at all.

Ebag333:
Also, FYI, this should be ready to merge.

blitzmann:
Ah, yeah you'd have to do a `.fetchall()` to get the actual results - that should get you either a `len()` or a `rowcount` property maybe.

Or you can do a `SELECT COUNT(*)` into the result and use that to drive whether or not to run the other query.

Or just remove the counting completely - a simple DELETE statement that covers the condition should suffice.

But none of this is breaking and can remain as is - like I said, eventually I want to change how this is triggered, so if anything it can be changed then.

Ebag333:
>Ah, yeah you'd have to do a .fetchall() to get the actual results - that should get you either a len() or a rowcount property maybe.

I can fiddle with that, but my understanding is that the `returnrows` and `rowcount` properties are never populated for a select, so I'm doubtful that'd work.

I haven't tried `.fetchall()` so I can see if that makes a difference.

>Or you can do a SELECT COUNT(*) into the result and use that to drive whether or not to run the other query.

You still end up with a result you have to loop through or otherwise extract the first record, so it's not really much more efficient.  Would be a single loop instead of possible a few hundred, but lines of code wise it's not any different.  It'd basically end up being what I did for getting the ID for Uniform damage profile and the All V character.

>Or just remove the counting completely - a simple DELETE statement that covers the condition should suffice.

This makes me uncomfortable.  I don't like blindly deleting things, it's a bad practice with SQL data and I want it well documented if we *DO* delete anything.

Speaking of which, I'm assuming that logging errors will always output to the log, correct?

>But none of this is breaking and can remain as is - like I said, eventually I want to change how this is triggered, so if anything it can be changed then.

Yeah.  I'm just getting tired of all the new issues being opened with the same root causes.  I just want to get something in place to get this cleaned up so we don't spend our time troubleshooting the same thing over and over again, and we don't have users rage quitting without even telling us (which I'm sure happens).

I think that we should get this implemented and extend it as we figure out other serious Pyfa breaking scenarios.  What little time this costs us starting up can be saved elsewhere (like fixing the triple loop looking for if you have a fit for each ship in each class).

blitzmann:
>  I'm assuming that logging errors will always output to the log, correct?

You assume too much.

Logging in pyfa is hit and miss. I made an attempt once to actually get it to work with errors, but doesn't seem to work all the time (like, 90% of the time it doesn't). This is actually an area that I want to revisit soon and harden up - logging all errors to a file, or if run with `-d` redirect all output to a file.

I first want to get #247 looked at once more and tested on the different platforms - as long as it notifies the user of the exception, logging it to disk is less important.

Ebag333:
>You assume too much.

Let me rephrase.

Info and error is _supposed_ to log to the file.  Debug _should_ only log if `-d` is there.

I know that logging isn't perfect.  @a-tal had a fit when he saw we used the built in logging at all, because it's slow and heavy.  And from debug stepping through it, I can't disagree.

Not sure what we would use to replace it, but even if `error` doesn't properly output now I wanted to get it in place so it's there to properly output once we fix logging and error reporting.

blitzmann:
@Ebag333 made a couple modifications to the PR; see latest commit. I didn't like the idea of SELECT * as that pulls back all data from the table, which could potentially be a performance issue. I find this cleaner. If you don't see anything wrong with these changes, I'll go ahead and merge. :)

Ebag333:
Ran against an empty DB:

```
  File "C:\Program Files (x86)\JetBrains\PyCharm Community Edition 2016.3\helpers\pydev\pydevd.py", line 1596, in <module>
    globals = debugger.run(setup['file'], None, None, is_module)
  File "C:\Program Files (x86)\JetBrains\PyCharm Community Edition 2016.3\helpers\pydev\pydevd.py", line 974, in run
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "./pyfa.py", line 125, in <module>
    import service.prefetch
  File ".\service\prefetch.py", line 68, in <module>
    database_cleanup_instance.OrphanedFitDamagePatterns(eos.db.saveddata_engine)
  File ".\eos\db\saveddata\databaseRepair.py", line 59, in OrphanedFitDamagePatterns
    if results.fetchone()['num'] > 0:
TypeError: 'NoneType' object has no attribute '__getitem__'
```

I'm getting `{TypeError}'NoneType' object has no attribute '__getitem__'` regardless if there's valid data, bad data, or no data at all.

I do see that results is returning a valid count, but I'm not sure that the method you're using to try and extract it works.

I submitted a commit with an alternate way of getting it. it's basically the same method used to grab the IDs (though we don't check if more than 1 record was returned, since we'll only ever get one).

blitzmann:
What do you mean an empty DB? I run this against a brand new saveddata.db and it runs fine. Even if I delete the characters and damage patterns that are created...

After looking at the SQLAlchemy docs, I think there's a better way of doing this anyway - made a commit. Please try to break it again (and explain in more detail how to broke the other one please, because I can't figure it out lol)

Ebag333:
>After looking at the SQLAlchemy docs, I think there's a better way of doing this anyway - made a commit. Please try to break it again (and explain in more detail how to broke the other one please, because I can't figure it out lol)

Uhh, I deleted the DB, ran Pyfa, it broke. vOv

Like I said initially, I tried a BUNCH of ways to get the data out of the results because I wanted the row count. The ONLY way I could get it to work was by looping through the results. Googling and searching through Sqlalchemy docs seemed to confirm that. 

if you found a better way I'm all for that. You must have a slight difference in your environment than I do, maybe a different version of Sqlalchemy or something. But since we have users that will have various versions, the low tech solution (even if it's slightly fugly) is probably the best way to go if it works reliably. 

Anyway, I'll try the latest commit. I'm curious to see how you solved this, because I can think of a good number of scenarios where I would like to query the DB directly, and don't want to have to loop through results just to get a simple count.

Ebag333:
>Uhh, I deleted the DB, ran Pyfa, it broke. vOv

On mobile so can't edit (grrr github) but I also tried it on a DB with data. It just doesn't work in my environment at all.

Ebag333:
>if row and row['num'] > 0:

Random Python thought, but instead of 
`if row and row['num'] > 0:`
Shouldn't we be able to do
`if row:`
since in Python false = None = 0?

Hmm that doesn't necessarily test existence though. I've had problems with that before, but that's not an issue here since we'll always get a number back.

Ebag333:
Okay, so did a few more things.

I added a check and fix for if there's null values (we were only checking missing, which doesn't actually catch null) for both damage pattern and character IDs on fits.

I broke out the query logic (where we're actually looking for results) into it's own function in `queries.py`.  Makes sense since we're already doing it twice, and will likely do it many more times down the road.

The function also converts the resultsproxy into a standard dict because eff that shit.  So now the object that gets returned we can do a simple `__len__()` and hey, there's the number of results we have, how easy is that?  (Technically it's a named dict for each row and the container object is a simple list, seemed the easiest and cleanest way to handle it.)

This function isn't designed to handle everything for you and be very standardized like the other queries (like say `getCrestCharacter`).  This expects you to know what you're doing and be ready to handle the return yourself (so know what the actual table column names are, or specifically alias them in the query).  But it should be good for one off scenarios, like that stupid takes-5-to-60-seconds-to-load loop within a loop within a loop where we're just getting a list of all the ship groups that have a fit in them so we can filter it down.


I did tighten up the `if` statements for the counts as mentioned above, but didn't mess with the queries.  There's no real reason to use the `executeDatabaseQuery` function for that, it'd be heavier than needed, and we couldn't use `__len__()` anyway unless we go back to querying all rows.  Better to keep it at just 2 lines of code.

Anyway, I think we have a winner here.  If you're okay with what I did breaking it out, can ya merge it in?  I want to use that new query function so I can shoot that long load issue in the head. :D

PR #921, opened 2016-12-27
Loading item as a booster if implant fails.

### LINKED ISSUE #919: import from clipboard has issues with drugs/boosters
When I try and import any fit that contains drugs/boosters pyfa won't do it.  Removing the  drugs from the paste fixes the issue.  For example:

```
[Armageddon Navy Issue, Test Drugs]

[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]

[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]

[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]

[Empty Rig slot]
[Empty Rig slot]
[Empty Rig slot]


Improved Mindflood Booster
Improved Frentix Booster

```
Will not import

Ebag333:
Fixes #919

This is ready to merge.


blitzmann:
Two things:

1) I can't check right now, but can't we check to see, instead of going off category name, if the item is of group name Booster or Implant or something? That would be much cleaner than a try...except. 

2) What about folks that want to keep boosters in cargo? Personally, I think if an EFT formatted fit has boosters, it should go into the cargohold rather than as an actual booster. It makes more sense to me, especially since I don't think EFT exports boosters and this is probably coming from EVE (in which case it will come from cargo since EVE doesn't export currently active boosters)

DarkFenX:
Can't you just rely on implantness/boosterness attribute?

resinneublem:
> What about folks that want to keep boosters in cargo?

+1 for keeping it in the cargo.

Ebag333:
>I can't check right now, but can't we check to see, instead of going off category name, if the item is of group name Booster or Implant or something? That would be much cleaner than a try...except.

I thought about that, but I got yelled at last time I used market group. :P

I don't really like the try/catch because if the catch fails we're up a creek and don't handle _that_.  I can't think of anything else it would be, since it's really just an artifact we created by treating boosters and implants separately when they're treated the same in game.

>What about folks that want to keep boosters in cargo? 

Unfortunately the EFT format doesn't distinguish between cargo and implants.

>Personally, I think if an EFT formatted fit has boosters, it should go into the cargohold rather than as an actual booster. It makes more sense to me, especially since I don't think EFT exports boosters and this is probably coming from EVE (in which case it will come from cargo since EVE doesn't export currently active boosters)

You can't add items in game to cargo.  So it's most likely coming from us, since we recently added the ability to export boosters.

It also already keeps it in the cargo if it's in the cargo.  This is _only_ to handle it if it's _not_ in the cargo.

For example (trimmed empty slots for space):
```
[Rorqual, Rorqual Drones]

High-grade Ascendancy Alpha
Antipharmakon Thureo

Antipharmakon Kosybo x1
High-Grade Asklepian Alpha x1
```

Top two are actually being used, bottom two are in the cargo.  Exporting and importing handles it correctly for Pyfa (from Pyfa, into Pyfa).

Since EFT doesn't export implants, and in game doesn't, I consider the `EFT (Implants)` export to be an "extended" export that is only officially supported by Pyfa.  Super weird things happen when you import items with cargo/implants into game (for cargo, it's there but invisible, which could make for really evil shared fits).

TL:DR; we already handle cargo vs implant correctly, we just need to handle implant vs booster better.

blitzmann:
> I thought about that, but I got yelled at last time I used market group. :P

Not a market group, but the items group. There is a difference. :)

@DarkFenX, we could use implantness/boosterness attributes. I forgot that implants and boosters have this, and it might be the only way to truly tell them apart programatically. Doing a quick check, implants can have at least a few group names ("cyberimplant", "Cyber Resource Processing", etc) so it's not reliable. 

For example, `if "implantness" in item.attributes` will yeild true for implants, `"boosterness"` for boosters.

> Top two are actually being used, bottom two are in the cargo. Exporting and importing handles it correctly for Pyfa (from Pyfa, into Pyfa).

k, makes more sense now. Thanks for the clarification. I'll double check soon hopefully and merge it in. If you could get rid of that try...catch, that'd be great :)

Ebag333:
>If you could get rid of that try...catch, that'd be great :)

Done.  Thanks for pointing out the attributes, that's a much cleaner way of handling it.

Also added some error catching and logging, so now a bad implant line won't fail the entire import.

blitzmann:
FYI, If ucked the merge up with this one. I didn't realized it was based on dev branch at first, so I switch merged it into master branch to get it into codebase faster. That merged a bunch of other stuff that isn't as well vetted, so I reset the merge and force pushed master again. 

This has been merged into dev branch. :)

Ebag333:
Well, that's what we want to do right? Develop against dev and then merge into master. 

If you prefer to pick and choose, which can make sense for some fixes, could always cherry pick the commit and close the PR so we don’t get it showing up twice.

blitzmann:
I haven't really thought of how I want to handle the dev branch, tbh. I think I want to use it as a staging area for new features or big changes. For small bug fixes like this,it can go directly in master.

PR #923, opened 2016-12-30
Tooltip: Total dps tanked

IndictionEve:
Added tool tips to the reinforced labels which show the total tankable HP based on the capacitor available time `(HP/s * CapLastTime)`.

Example

The tooltip is only available if the fitting is not capstable.

This tool tip makes it much easier to compare passive tanked  ships vs active tanked ships.
Additionally I added some more text to the "passive tank" tool tip.

### DIFF
--- a/gui/builtinStatsViews/rechargeViewFull.py
+++ b/gui/builtinStatsViews/rechargeViewFull.py
@@ -105,7 +105,9 @@ class RechargeViewFull(StatsView):
                 lbl = getattr(self, "labelTank%s%sActive" % (stability.capitalize(), name.capitalize()))
                 if tank is not None:
                     lbl.SetLabel("%.1f" % tank["%sRepair" % name])
-                    if (tank["%sRepair" % name] > 0) & (not fit.capStable): #show tooltip only if fit is not capstable
+                    if (stability == "reinforced") & \
+                       (tank["%sRepair" % name] > 0) & \
+                       (not fit.capStable): #show tooltip only if fit is not capstable
                         totalDPSTank = tank["%sRepair" % name] * fit.capState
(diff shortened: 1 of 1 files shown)


Ebag333:
I'm not sure that this really is a stat we should be exposing. 

>This tool tip makes it much easier to compare passive tanked ships vs active tanked ships.
I don't think that people compare active vs passive tanked ships this way. Most people I know look at the buffer for passive ships, active tanked ships are used primarily for small gang PvP where you can tank up to a certain DPS number. 

Ancil reppers can be used for buffer, but we already expose that number. 

The number that you get for how long cap lasts is a bit garbage anyway. Unless it's generated by being completely neuted out, you can run your repper long after you hit that number. That's why lots of folks only shoot for 2-3 minutes cap times, they know that they can run stuff afterwards, you just can't run it full out. 

There are also issues with the cap time and repper HP/s generated, because bad assumptions are made currently. Like assuming that ancil reppers have a 10s reload time (it's 60).

You can look at #812 where I call out a number of these issues. 

I think it's an interesting idea, but not one that matches the way numbers are generally used, and shows a number that won't match up to actual in game numbers. I think that it will cause more confusion than it clears up.

blitzmann:
@IndictionEve just FYI, since I haven't commented yet, I've been waiting on this until we get some new code related to this issue in the dev branch, then I'll evaluate this one. Thanks, and sorry for the delay!

Gochim:
@blitzmann Hi, I'm opening another grave, even deeper that the previous one, I know :)

Were there any updates regarding this issue?

Marzhin971:
If this feature is added, it would be more relevant to name it "Total tanked Damage" or "Total tanked EHP", "Total" and "DPS" are mutually exlclusive.

PR #927, opened 2017-01-01
ECM Burst Projectors effect

Ebag333:
Change to use skill rather than group, and modify correct attribute.

Another ECM burst bug that's been this way for years.  I swear, no one uses these things.

Thanks to saeka for reporting the bug.

### DIFF
--- a/eos/effects/skillremoteecmdurationbonus.py
+++ b/eos/effects/skillremoteecmdurationbonus.py
@@ -6,5 +6,28 @@ type = "passive"
 
 
 def handler(fit, skill, context):
-    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Burst Projector Operation"),
-                                  "durationECMJammerBurstProjector", skill.getModifiedItemAttr("projECMDurationBonus") * skill.level)
+    # We need to make sure that the attribute exists, otherwise we add attributes that don't belong.  See #927
+    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Burst Projector Operation") and
+                                              mod.item.getAttribute("duration"),
+                                  "duration",
+                                  skill.getModifiedItemAttr("projECMDurationBonus") * skill.level)
+
+    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Burst Projector Operation") and
+                                              mod.item.getAttribute("durationECMJammerBurstProjector"),
+                                  "durationECMJammerBurstProjector",
+                                  skill.getModifiedItemAttr("projECMDurationBonus") * skill.level)
+
+    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Burst Projector Operation") and
+                                              mod.item.getAttribute("durationTargetIlluminationBurstProjector"),
+                                  "durationTargetIlluminationBurstProjector",
+                                  skill.getModifiedItemAttr("projECMDurationBonus") * skill.level)
+
+    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Burst Projector Operation") and
+                                              mod.item.getAttribute("durationSensorDampeningBurstProjector"),
+                                  "durationSensorDampeningBurstProjector",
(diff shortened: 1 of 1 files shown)


blitzmann:
I just checked my effects script, it's reporting 4 more modifiers than what we have:

```
type = "passive"
def handler(fit, src, context):
    lvl = src.level
    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Burst Projector Operation"), "duration", src.getModifiedItemAttr("projECMDurationBonus") * lvl)
    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Burst Projector Operation"), "durationSensorDampeningBurstProjector", src.getModifiedItemAttr("projECMDurationBonus") * lvl)
    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Burst Projector Operation"), "durationTargetIlluminationBurstProjector", src.getModifiedItemAttr("projECMDurationBonus") * lvl)
    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Burst Projector Operation"), "durationECMJammerBurstProjector", src.getModifiedItemAttr("projECMDurationBonus") * lvl)
    fit.modules.filteredItemBoost(lambda mod: mod.item.requiresSkill("Burst Projector Operation"), "durationWeaponDisruptionBurstProjector", src.getModifiedItemAttr("projECMDurationBonus") * lvl)

```

Can you double check if these things do anything?

Ebag333:
Not sure where the values are coming from TBH. 

`duration` for `projECMDurationBonus` isn't valid. CCP probably changed the attribute used. 

The other 3 don't make sense to key off `projECMDurationBonus`. I mean, it might just be `CCP logic` but it doesn't pass the sniff test for me. 

I will take a look again and see if there are other projected effects, don't have Pyfa in front of me at the moment, but I think you are seeing remnants from something that was never implemented or was removed. 

Either way, that wouldn't impact this PR.

Ebag333:
Okay, so your list is right.

The _issue_ that we have is that if we apply these as is, we'll be creating empty values in the attributes tab for the other 4 attributes that don't exist for that module.

Not quite sure on a clean way to handle it yet,  My OCD tripped because the old effect had duration, which created an empty value in the attributes tab for ECM burst.

So I'll see what we can do with this, definitely should hit them all up, but I don't want to add values that shouldn't be there.

Ebag333:
@blitzmann think I got it.  It works correctly setup this way, and we don't get all 5 attributes added to all the modules when only 1 is used.

blitzmann:
Actually, just took a look at this. You can remove the four odd modifiers. These must be internally used by CCP in their implementation of these modules, but for pyfa we only need to change the duration modifier, which works without the others.

This sometimes happens, especially for skill effects - there's modifiers that are used by CCP for their stuff, but don't necessarily have anything to do with pyfa. I got a little confused because the initial commit in this PR changed the `duration` modification to `durationECMJammerBurstProjector`.

BTW, using `mod.item.getAttribute()` is pretty clever. There's a few places where we modify attributes that don't exist simply because the effect tells us to. If we ever run into these, we should remember to do this.

Ebag333:
> I got a little confused because the initial commit in this PR changed the duration modification to durationECMJammerBurstProjector.

Well remember that I initially was just fixing projected ECM.  I broke some of the others by removing duration. 

Don't ask me why CCP is being difficult and using special attributes for duration. That reminds me though, I need to add these special durations to the Gnosis handler. 

So as of now this should be good and the skill will apply to all the projected ewar properly.

blitzmann:
>That reminds me though, I need to add these special duration to the Gnosis handler.

No you don't. The `duration` modifier should be enough (the others seems to be redundant), which is why I'm partial to not even include them since they aren't used in anything else. ¯\_(ツ)_/¯

Ebag333:
>No you don't. The duration modifier should be enough (the others seems to be redundant), which is why I'm partial to not even include them since they aren't used in anything else. ¯_(ツ)_/¯

Uhhh, duration doesn't exist on all the modules.  ECM, Damps, TP (Illumunation), Disruptor all use their own.  The rest use duration.

Issue #928, opened 2017-01-01
cx_Freeze: Python error in main script


porowns:
Hey there, recently got this error and the solutions to previous instances of this error didn't resolve it. 
![d9ff6e7193d260ad845c5d06baa59b6b](https://cloud.githubusercontent.com/assets/16302735/21582402/dbfcf73a-d023-11e6-81b0-58cb6b40c1b1.png)

Tried reinstalling a few times to a few different locations, couldn't figure it out.

blitzmann:
Did this just start recently? Was it working before? 

The actual issue is a case of database corruption. We don't know what causes it (hense the questions to gather more info), but we have a "fix" for it that should be coming in the next release. 

Could you send me a copy of your `%userprofile%/.pyfa/saveddata.db` file to the email in the README? I'll go ahead and fix it up and send it back to you. :)

porowns:
Yeah, I installed Pyfa, originally tried to import a character with an EVEMON file, then wasn't able to interact at all with adding/duplicating/deleting characters. Could not create any new characters or anything.

After that, I renamed the All 5's to try and edit it, and this issue started occuring.

Ebag333:
@blitzmann because EVEMon is out of date, I bet it has skills that we don't. So then the import fails, and poof corruption. 

A working theory at least, and something to look into...

porowns:
@Ebag333 I no longer remember which specific character file I used, but I just imported them all and Pyfa seems to be working as intended. Must have been an odd outlier case. I can't seem to reproduce the issue.

Ebag333:
>but I just imported them all and Pyfa seems to be working as intended.

Drat.

porowns:
@Ebag333 Scratch that, just reproduced the issue, emailing you the file.

Ebag333:
Anything special that you did to reproduce? Or just straight up import the file?

porowns:
I did a clean installation of Pyfa, and imported that file as the first thing I did. The import did not work (no error message, just nothing happened), and after closing and reopening Pyfa, character editor is completely broken.

blitzmann:
> I did a clean installation of Pyfa,

When you say clean installation, do you mean you nuked your user profile data as well?

[...]

PR #929, opened 2017-01-01
Add some sanity checks and validation of imported character data

Ebag333:
Now checks to make sure that the skill we're trying to import exists, and that the level being set is between (or equal to) 0 and 5.

We also now output more logging data, previously errors were simply printed to the console, which being a threaded process wouldn't show in the main console and end users would never see the error.

If a skill that is being imported isn't in our list (generated based on the All 5 character), or the skill level is out of range, we simple log an error and continue on.

Hopefully helps with issues like #928

### DIFF
--- a/service/character.py
+++ b/service/character.py
@@ -272,8 +272,11 @@ class Character(object):
         return char
 
     def rename(self, char, newName):
-        char.name = newName
-        eos.db.commit()
+        if char.name in ("All 0", "All 5"):
+            logger.info("Cannot rename built in characters.")
+        else:
+            char.name = newName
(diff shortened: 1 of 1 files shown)


Ebag333:
I also added a check on the rename to make sure we're not renaming built in chars, since the GUI isn't always properly locking the rename button for them.  Should help stop people from doing something they shouldn't.

blitzmann:
This doesn't address the gui issue of being able to rename the all 5 character. I'll create an issue about that one. It's not priority as this PR ensures it's not allowed.

Thanks!

Issue #932, opened 2017-01-03
Stack trace when cloning modules


Ebag333:
This is against the dev branch. Getting this stack trace:

```
Traceback (most recent call last):
  File ".\gui\builtinViews\fittingView.py", line 104, in OnData
    self.dropFn(x, y, data)
  File ".\gui\builtinViews\fittingView.py", line 190, in handleListDrag
    self.swapItems(x, y, int(data[1]))
  File ".\gui\builtinViews\fittingView.py", line 400, in swapItems
    sFit.cloneModule(self.mainFrame.getActiveFit(), mod1.position, mod2.position)
  File ".\service\fit.py", line 616, in cloneModule
    if new.fits(fit):
  File ".\eos\saveddata\module.py", line 401, in fits
    for attr in self.itemModifiedAttributes.keys():
  File "C:\Python27\lib\_abcoll.py", line 410, in keys
    return list(self)
  File ".\eos\modifiedAttributeDict.py", line 131, in __iter__
    all = dict(self.__original, **self.__modified)
TypeError: 'NoneType' object is not iterable
```

Fit is:
```
[Loki, Instalock Blapper]

Damage Control II
Gyrostabilizer II
Gyrostabilizer II

True Sansha Warp Disruptor
50MN Quad LiF Restrained Microwarpdrive
Sensor Booster II, Scan Resolution Script
Sensor Booster II, Scan Resolution Script
Sensor Booster II, Scan Resolution Script

Information Command Burst II, Sensor Optimization Charge
Information Command Burst II, Sensor Optimization Charge
720mm Howitzer Artillery II, Republic Fleet EMP M
[Empty High slot]
720mm Howitzer Artillery II, Republic Fleet EMP M
720mm Howitzer Artillery II, Republic Fleet EMP M
720mm Howitzer Artillery II, Republic Fleet EMP M

[Empty Rig slot]
[Empty Rig slot]
[Empty Rig slot]

Loki Propulsion - Interdiction Nullifier
Loki Engineering - Power Core Multiplier
Loki Electronics - Tactical Targeting Network
Loki Offensive - Turret Concurrence Registry
Loki Defensive - Warfare Processor
```

Some items clone successfully, some don't.  Cloned the burst fine, can't clone the Arty.  This fit was cloned from another fit, so _shouldn't_ be a DB corruption issue.

In `module\saveddata.py`, line 401:
`        for attr in self.itemModifiedAttributes.keys():`

[...]

PR #933, opened 2017-01-04
Database Corruption: Multiple Items in Single Slot on Fit

Ebag333:
Fixes an issue where duplicate items will get added to a single slot.  See #932

Not sure where this comes from, possible source of database corruption is:
1) Cloning existing fit
2) Swapping T3 subsystems (as this rapidly adds/removes slots from the fit)
3) Race condition where the fit is calculated while still populating (because we trigger calculations multiple times)

This is definately the brute force method by just nuking the conflicting modules.

Unfortunately, I don't really see another way to do it.  When we have multiple modules, we don't know which module is good and which is bad.  We could run through a series of deletions to clean up some more obvious ones (like removing nulls before modules), but that then starts to become expensive time wise, especially since this sort of SQL query is expensive.

Even in my fairly large DB, this amounted to just a handful of fits, so it's a rare issue.  While I _hate_ removing data, the issue should be rare enough and the amount of data removed small enough that I'm leaning toward the simpler method.

Even though this isn't a very efficient SQL query, it's still pretty fast.  I wouldn't want to start chaining a bunch of these though.

```
2017-01-04 12:33:16,773 eos.db.saveddata.databaseRepair DEBUG    Running database cleanup for character skills.
2017-01-04 12:33:16,792 eos.db.saveddata.databaseRepair DEBUG    Running database cleanup for orphaned characters attached to fits.
2017-01-04 12:33:16,822 eos.db.saveddata.databaseRepair DEBUG    Running database cleanup for orphaned damage patterns attached to fits.
2017-01-04 12:33:16,835 eos.db.saveddata.databaseRepair DEBUG    Running database cleanup for multiple modules in a single slot on a fit.
2017-01-04 12:33:17,098 eos.db.saveddata.databaseRepair ERROR    Database corruption found. Cleaning up 49 records.
```

### DIFF
--- a/eos/db/saveddata/databaseRepair.py
+++ b/eos/db/saveddata/databaseRepair.py
@@ -29,9 +29,6 @@ class DatabaseCleanup:
 
     @staticmethod
     def OrphanedCharacterSkills(saveddata_engine):
-        # Finds and fixes database corruption issues.
-        logger.debug("Start databsae validation and cleanup.")
-
         # Find orphaned character skills.
         # This solves an issue where the character doesn't exist, but skills for that character do.
         # See issue #917
@@ -43,6 +40,31 @@ class DatabaseCleanup:
 
             if row and row['num']:
                 delete = saveddata_engine.execute("DELETE FROM characterSkills WHERE characterID NOT IN (SELECT ID from characters)")
+                logger.error("Database corruption found. Cleaning up %d records.", delete.rowcount)
+
+        except sqlalchemy.exc.DatabaseError:
+            logger.error("Failed to connect to database.")
+
+    @staticmethod
+    def DuplicateModulesOnFit(saveddata_engine):
+        # Finds fits with multiple modules in a single slot.
+        # See issue #932
+        try:
+            logger.debug("Running database cleanup for multiple modules in a single slot on a fit.")
+            results = saveddata_engine.execute("SELECT count(*) AS num FROM modules m1 "
+                                               "JOIN (SELECT count(*) as count, fitID, position from modules "
+                                               "GROUP BY fitID, position) as m2 "
+                                               "ON m1.fitID = m2.fitID and m1.position = m2.position "
+                                               "WHERE m2.count > 1")
+            row = results.first()
+
+            if row and row['num']:
+                delete = saveddata_engine.execute("DELETE FROM modules WHERE ID IN "
+                                                  "(SELECT ID FROM modules m1 "
+                                                  "JOIN (SELECT count(*) as count, fitID, position from modules "
+                                                  "GROUP BY fitID, position) m2 "
+                                                  "ON m1.fitID = m2.fitID and m1.position = m2.position "
+                                                  "WHERE m2.count > 1)")
                 logger.error("Database corruption found. Cleaning up %d records.", delete.rowcount)
 
         except sqlalchemy.exc.DatabaseError:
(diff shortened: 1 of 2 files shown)


blitzmann:
1) I'm not sure this is actually an issue in the real world. It may be that due to development, you've fumbled your DB up a bit. And while your stated corruption sources are possibilities, I doubt it would happen - cloning a fit copies 1:1, subsystems shouldn't matter as swapping them out leaves the slots there technically and replaces it with red background letting you know it's illegal, and race conditions are pretty rare in a single threaded environment, of which most (all?) fit functions are.
2) Even if it does happen in the real world, the position really should only be used for sorting the modules IIRC. The error that you came across seems to have slipped through the cracks, and might be something we can deal with during that operation

I'm not partial to removing data. The only situation in which I'm comfortable doing that is when the item no longer exists in the EVE database - we remove it automatically upon fit load so as to prevent errors.

Let me take a look at the problem at hand first - haven't had a chance yet. I'll get back to you on where we want to go with this PR. Thanks

Ebag333:
>I'm not sure this is actually an issue in the real world. It may be that due to development, you've fumbled your DB up a bit.

Pretty sure this isn't due to dev and me fumbling the DB.  I was using the dev branch, no custom code, and I'd just cloned a Loki fit that _did not_ have this problem (so it's not a historical issue).  This was freshly cloned (as in minutes before I reported it).  Is it possible that _something_ that I've done to my DB _somewhere_ borked it?  Sure, but overall I really doubt it.

As for the possible sources of corruption, since neither of us have any idea of the source I'm throwing things out based on what I did before the corruption was noticed.  Namely:
1) Clone fit
2) Switch subs

I have no idea if these are the sources of corruption, but they're definitely things to look into as potential sources.

>I'm not partial to removing data. 

I'm not either.  But lets be honest, it's going to be a boatload of additional work to fix this any other way.

Probably the next easiest solution is when we load a fit the first time, we loop through the modules and check for more than one module in each slot.  When we hit one, we pop up a dialog and list the modules.  They select which one they want, hit save, and it drops the others from the database.

That's a lot more work this, but it _is_ a better solution.  We do have the potential issue that we're not in prefetch, so the DB being in use can make things more difficult.  If we catch this early before the fit actually opens, that should minimize that risk.

blitzmann:
> Pretty sure this isn't due to dev and me fumbling the DB. I was using the dev branch, no custom code, and I'd just cloned a Loki fit that did not have this problem (so it's not a historical issue). This was freshly cloned (as in minutes before I reported it). Is it possible that something that I've done to my DB somewhere borked it? Sure, but overall I really doubt it.

Does the source fit also exhibit this issue, or just the clone? Haven't looked at your database yet. 

I did take a quick look today, and indeed, position is only really used in two places: in the ordering of the modules (this is actually handled at a very low level - the sql alchemy collection class for modules), and this cloning thing. I wrote the cloning thing, and I believe I'm not using it correctly. The reason for the error is because we're trying to do something like fit.modules[position], and it's failing. Position should be the same thing as the index, but apparently it can get a bit mixed up...


> I'm not either. But lets be honest, it's going to be a boatload of additional work to fix this any other way.

I honestly doubt this. I will look at it a bit tonight, but I'm about 95% sure this can be fixed without touching the database. Your other solution is indeed a lot more work, but I don't think we'll need to go that far. We'll see after I take a peak.

Ebag333:
>Does the source fit also exhibit this issue, or just the clone?

Just the clone.

Ebag333:
Closing as @blitzmann has already submitted a fix. (Well done.)

Issue #517, opened 2016-02-10
User-selected price source


bvelasquez:
Is this in a configuration file somewhere?  I'm not sure what market (Jita maybe), but would like to change the market to my staging system/station.

thorr18:
Currently no setting for that. It's hardset in your service/price.py to system "30000142" (Jita)

bvelasquez:
Perhaps an option in settings that allows entering the ID (which is available via many methods) as a quick/dirty way of changing this.

blitzmann:
Yeah, I can do this. The issue that was preventing it from being implemented was that we didn't want to include solar system info in the database, but entering an ID could definitely work (even if it's not so user-friendly).

bvelasquez:
@blitzmann Great. Call it an "advanced user" feature and not worry about the user experience for now.  It still would be a step in the right direction.  Thanks.

ppfeufer:
How bout a "simple" select? Just for Jita, Amarr, Dodixie and Rens ....

That's 4 systems, I think that should be doable and no one has to do some obscure rituals to get a system ID.

blitzmann:
I was actually thinking about possibly using CREST or the EVE XML API to look up the system ID. Never made it that far though, got bogged down with something else.

blitzmann:
A simple implementation has been added via #934. Only does a few systems, but we may be able to expand if needed.

PR #934, opened 2017-01-04
Price system selection

petosorus:
I worked on the possibility to select the source system for prices, as suggested in #517

I only included the 5 main trade hubs in the selection. A service could be written to get all systems from CREST or ESI but I wasn't sure how I wanted this to work.

I don't know wx at all so the drop down isn't exactly to my taste regarding width and selection ordering.

I'm also refreshing the whole fit instead of only the price column in the main fit window and it feels a bit hacky but I couldn't see an easy way to update only one column without rewriting some of it for now.

### DIFF
--- a/config.py
+++ b/config.py
@@ -18,9 +18,9 @@ debug = False
 saveInRoot = False
 
 # Version data
-version = "1.23.3"
-tag = "git"
-expansionName = "YC 118.7"
+version = "1.24.0"
+tag = "Stable"
+expansionName = "YC 118.8"
 expansionVersion = "1.4"
 evemonMinVersion = "4081"
 
--- a/eos/effects/ecmburstjammer.py
+++ b/eos/effects/ecmburstjammer.py
@@ -1,3 +1,5 @@
+# ECMBurstJammer
+#
 # Used by:
 # Modules from group: Burst Jammer (11 of 11)
 type = "projected", "active"
--- a/eos/effects/iceharvestercapacitorneedmultiplier.py
+++ b/eos/effects/iceharvestercapacitorneedmultiplier.py
@@ -1,8 +1,4 @@
-# iceHarvesterCapacitorNeedMultiplier
-#
-# Used by:
-# Variations of ship: Procurer (2 of 2)
-# Variations of ship: Retriever (2 of 2)
+# Not used by any item
 type = "passive"
 def handler(fit, ship, context):
     fit.modules.filteredItemMultiply(lambda mod: mod.item.requiresSkill("Ice Harvesting"),
--- a/eos/effects/iceharvesterdurationmultiplier.py
+++ b/eos/effects/iceharvesterdurationmultiplier.py
@@ -1,8 +1,6 @@
 # iceHarvesterDurationMultiplier
 #
 # Used by:
-# Variations of ship: Procurer (2 of 2)
-# Variations of ship: Retriever (2 of 2)
 # Ship: Endurance
 type = "passive"
 def handler(fit, ship, context):
--- a/eos/effects/lightningweapon.py
+++ b/eos/effects/lightningweapon.py
@@ -1,3 +1,4 @@
+# Not used by any item
 type = 'active'
 def handler(fit, module, context):
     pass
--- a/eos/effects/miningyieldmultiplypassive.py
+++ b/eos/effects/miningyieldmultiplypassive.py
@@ -1,8 +1,6 @@
 # miningYieldMultiplyPassive
 #
 # Used by:
-# Variations of ship: Procurer (2 of 2)
-# Variations of ship: Retriever (2 of 2)
 # Variations of ship: Venture (3 of 3)
 type = "passive"
 def handler(fit, module, context):
--- a/eos/effects/pointdefense.py
+++ b/eos/effects/pointdefense.py
@@ -1,7 +1,4 @@
-# targetAttack
-#
-# Used by:
-# Citadel Point Defense
+# Not used by any item
 type = 'active'
 def handler(fit, module, context):
     pass
--- a/eos/effects/rolebonusiceoreminingdurationcap.py
+++ b/eos/effects/rolebonusiceoreminingdurationcap.py
@@ -1,3 +1,7 @@
+# roleBonusIceOreMiningDurationCap
+#
+# Used by:
+# Variations of ship: Covetor (2 of 2)
 type = "passive"
 def handler(fit, src, context):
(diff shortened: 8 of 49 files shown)


blitzmann:
Thanks for the PR!

This is how I envisioned it (I have a local branch with a half implemented version of this): The drop down resides in the preferences and is saved as a user setting as their default location. When pyfa starts, it uses this system. To quickly change, though, a contact menu can be added tot he price pane that can list the different systems. This would be a temporary override of the default selection, and last only until pyfa is restarted. I think I would prefer this mechanism instead of having a dropdown in the price pane, for various reasons.

As for refreshing the whole fit... honestly I don't know off the top of my head if there is a better way. I think there is (compiling the items to be sent to the price service). To handle the cache, you could pass a parameter to the price service to forceRefresh or something (IIRC, we actually have a function that clears the entire price cache, so that's a thing that can be used...). That being said, I don't mind the entire fit refreshing if needed at the moment. That would be a good starting point, and from there we can improve on it (not refreshing the fit, context menu switching, etc). I can also assist on things like wx is needed. :)

Ebag333:
This would be a good setting to put into the preferences pane.

I'd rather have them set the setting and keep it, most people have a home system that they want to search for (like I'd *always* use Amarr over Jita), so I don't know about resetting it each time you restart Pyfa.  That'd be rather annoying.

Can we see about getting the preference pane PR merged in, then I (or you) can assist with getting a setting added for the system to search in?

petosorus:
I agree that the preferences thing is a better way.

What is a contact menu? I just don't know the word :)

blitzmann:
Context menu? When you right click, it pops up a menu with selections based on where in the GUI it was clicked, so we can selectively show certain things when right clicking price pane. ie: the clear prices functionality is currently available under the price pane context menu. In this case, it would be nice to have somethings that adds another item, called "System", with a submenu giving a variety of different systems.

The context menu code can be a little confusing, so I would focus on migrating it to preferences and saving as a user setting first. We can the focus on improvements. :)

petosorus:
Got it. It would be nice as a further step yeah.

Would we prefer to get that in @Ebag333's new preference pane or do I starting working on putting that in the current preference window?

blitzmann:
Current window please. I haven't really looked at the PR for preference pane stuff yet, and this should be a quick and easy thing to move over if needed :)

Ebag333:
We don't really have a preference pane that fits this sort of thing, either currently or in the preference pane PR.  Network _kinda_ fits since it has the option to enable updates, but selecting the system to pull from doesn't really belong there.

Probably the best option is to create a new pane for API related things.  We'll need one eventually since we're planning (hoping!) on implementing some of the features from ESI, so we'll want a variety of selections (for example what level insurance to use).

When I mentioned the preference pane PR, I was thinking since I'd added a bunch of panes one might fit, but looking at it now they really don't.  No sense in mixing some stuff from this PR in that PR.

@blitzmann 3 minutes ago:
>I haven't really looked at the PR for preference pane stuff yet

@blitzmann on Nov 30, 2016:
>I'll take a look at this soon

You and I have very different definitions of soon. 😁

blitzmann:
I actually think it would be best to put it in general pane for now. Creating a page is a little overwhelming for one option, and more effort. We need to sit down ato some point and hammer out better organizations for it, but for now it can be a simple option in general :)

Ebag333:
The only potential issue with that is....the general pane is full. :D
https://puu.sh/ti6rS/9741b0d99a.png

We can double column it and make it wider though.

blitzmann:
Ehh... didn't realize that.

We could also make the window slightly higher. It's currently hardecoded at a certain height. I don't remember how the panes would react to that though, worth a shot as a quick fix

Ebag333:
We can adjust the size.  Just the height is already about the maximum of anything that we have set, and you'd expressed concerns about folks still on low resolutions.

Width wise we should have more room to play with, since traditional resolutions are wider than they are tall, and right now the preference pane window is very narrow.

blitzmann:
It's nowhere near maximum. Here is the preference pane overlayed onto pyfa at it minimum dimensions: 

image

15-20 px more isn't going to be a problem. I mean yeah we could do a column situation, but I'd rather not for the following reasons:
* Have to deal with text wrapping correctly, which is a bitch
* Have to deal with getting wx to lay the columns out correctly, which is a bitch
* Considering OP has stated that GUI is not his thing, the easiest solution is the best solution in this case.

@petosorus Go head and increase the height of the preference pane to allow for a new dropdown at the bottom, if required. This would be the simplest way.

petosorus:
I'll get on that this week or the next.

blitzmann:
@petosorus any news on this? I don't think we can make this into the release next week, but I'm hoping to get in into the one after. Let me know if you need additional support on it and I can potentially hop in. :)

petosorus:
Talk about a 'next week' task. 

Eve life and real life got very busy at the same time, and yeah. 

I have a lot of time tomorrow and I will check how the preferences pane works. I'll bother you if I need support, and thanks for the poke :)

petosorus:
I need a bit of assistance. 
I moved the system selection to the preferences pane, but I don't understand how to get the choice to persist.

blitzmann:
> I moved the system selection to the preferences pane, but I don't understand how to get the choice to persist.

You'll have to save it to the user's settings. I thing "General" settings are actually part of the fit service rather than the settings service (don't ask why, long story). See:

https://github.com/pyfa-org/Pyfa/blob/f7d7b3fe5e9cdbd1cb398767708df81991b56dad/service/fit.py#L65-L65

So you would add a new setting and give it a Jita system default. To change it in preferences, simply set it like all other settings there.

To retrieve it for use in the price fetch, `sFit.serviceFittingOptions["priceSystem"]` or whatever you choose to name the setting.

petosorus:
My git- fu is a bit weak and I got messy on today's commits but I think it works in the right place now

blitzmann:
Look pretty good. I tweaks a few things but overall looks solid. Thanks!

blitzmann:
@petosorus just FYI, the way you were setting the current system didn't work out. I didn't catch it when I merged, but it was an easy fix. The problem was that you weren't setting a default - when pyfa restarted, `currentSystemID` was not populated.

Regardless, the best way is to import the fit service and get the setting directly when trying to fetch prices. See 2dca9fc3f333054fab105a1f76f37cc028ac7f6b

Just fyi, alls good now. Just wanted to bring it to your  attention in case you were interested. Awesome addition :)

petosorus:
Oh right. I got a bit messy at the end but that's interesting. Thanks for explaining :)

PR #943, opened 2017-01-12
Add variations for: drones, fighters, boosters, and implants

Ebag333:
This adds to the context menu a variations option for:
*Drones
*Fighters
*Boosters
*Implants

Example:  https://puu.sh/tjyNK/0ccdbe2a48.gif

This isn't _quite_ where I want it.  It works beautifully, but there is an issue when we're searching for related items by group instead of parent.  Since CCP has dumped implants and boosters into one giant group, the list is very large.  For boosters it's rather managable, but implants are not.

I need to add some logic in to parse out the name and filter based on that, so that the context menu only shows related implants/boosters.

Essentially we have something like: `Strong Blue Pill`

We would strip `Strong` and then add to our where query `LIKE '%Blue Pill'`
(Or however SQLAlchemy handles wild cards....need to figure that out.)

Implants would strip the prefix like `High-Grade`.

There are some other potential...idiosyncracies.....with this.

If you have:
```
2x T1 Warriors
3x T2 Warriors
```
And swap the T1 for T2, it merges the stacks automatically.  This can put you over the stack limit of 5.  This is a fairly unusual case and there's other ways to get over the stack limit, since the user can then split the stack it's not a huge issue.

When you swap, the order of drone stacks sometimes changes.

Gecko shows as a valid choice for Dragonfly fighters.  This is a CCP thing, because Gecko is sometimes considered a fighter and sometimes a heavy drone.  If you choose Gecko, the stack gets removed but of course the Geckos aren't added (the add step is still ran, but it is not added and no stacktrace is generated).  Since this is a game data database issue, I don't consider this a bug but a CCP "feature."


Ebag333:
Added the extra bits I wanted to.

@blitzmann this is done, and can be merged.

So there are a couple little "weird" things that can happen with implants and boosters.  For implants/boosters that are special, we end up showing the whole list (alternative would be to show nothing).

You can also end up where you find an item one way but not exactly the other.  For example, variations of `Strong Blue Pill Booster` shows `Nugohuvi Synth Blue Pill Booster` (because we trim down to just `Blue Pill Booster` but if you try and show variations of  `Nugohuvi Synth Blue Pill Booster` you get the whole list, because we're not trimming off the faction name (we trim `Synth` but not `Nugohuvi Synth`).  We could add it to the list, but there's a slew of these one off scenarios and since we're doing this via text parsing it's not an exact science.

I considered trying to filter by slot, the issue there is that certain slots (for example slot 1 for drugs) have tons of items and then you'd end up with a giant list.

Probably the cleanest solution in the long run would be to modify the context menu to allow nested sub-menus, and then group by meta level (for items) or slot (for implants/boosters).  But that's a bigger change than what I was going for.

blitzmann:
I'll take a look at this in a bit, and at first glance looks fine.

But holy fuck, the gif. Does it really take 2+ seconds to swap the item to the new item?

Ebag333:
Pyfa is running super slow for me. This isn't normal even for me. Haven't troubleshooted why yet.

blitzmann:
Pyfa as a whole, or this feature?

Ebag333:
Pyfa as a whole.  It might just be the environment needs rebuilding, pyCharm was throwing some errors (for pyCharm, not Pyfa or stack traces).

However, I did notice that the context menu slows things down a lot.  It's not fetching the variants from the database, but seems to be actually generating the context menu.

blitzmann:
Is it this slow when running the release version (pyfa.exe)?

Ebag333:
Like I said, not Pyfa.

Seems to be due to a bad driver causing my CPU to get throttled at 50%.  Either that or Microsoft AV is choking it, it seems to love scanning Pyfa related files for some reason.

Ebag333:

blitzmann:
@Ebag333 it wasn't hard to update to dev. The reason for the slew of changes that was listed here was because you merged your `_development` branch, which our branch doesn't know about, so it thought all of it was new.

`git reset --hard HEAD~1` and merge, was easy.

Just tested this, looks good. I agree the name parsing is a bit hacky, but it works well and in fact we use it for the ammo changer. The only change I would like to see is the implant list sorted if possible, because I'm getting something like this:

image

blitzmann:
@ebag I went ahead and added sorting to these thing (as best as I could given CCP doesn't give us any decent way to group or rank these items).

Feel free to check it out, and let me know if there are any oddities. I'm going to merge now.

Ebag333:
Thanks. I trust you can manage a bit of sorting. :)

I'm still trying to remember what I did to fix the preference pane. :(

blitzmann:
For documentation, I sort boosters by the side effect chance (the higher the chance, the strong the booster). If we're showing a lot of boosters, just sort by name. Sort my name for implants as well.

blitzmann:
@Ebag333 question:
 just noticed this, but there are a lot of `type(selected_item).__name__ == 'Drone'`

Why did you choose to do that instead of `selected_item.__class__` or even a simple `isinstance(selected_item, Drone)`?

Ebag333:
>`isinstance(selected_item, Drone)`

That's the one I'd prefer.  It's quite a bit faster than `type(selected_item).__name__ == 'Drone'`.  @a-tal pointed that out to me on another thing, but I didn't  go back and retroactively change it.

There's a few other places where we used `type` that we could refactor to speed it up.

You mind making that change, since you caught it?

blitzmann:
> That's the one I'd prefer. 

Holy fuck a syntax we agree on!

I can clean it up. Just didn't know if there was a specific reason for it :)

Ebag333:
The specific reason is...I didn't know about the `isinstance` syntax.  Or rather, I knew the bits and pieces, but didn't connect them together.

You won't find any argument from me if there's a simpler, faster, and easier syntax.  This one checks all three boxes.

PR #948, opened 2017-01-17
Revert back to /s for mining. Adds /hour to tooltips.

Ebag333:
From #790 
>Linking https://www.reddit.com/r/Eve/comments/5o2npc/miner_qol_improvement_m3s_calculated_inclient_on/ where people disagree that m3/h is useful.
>Now that EVE has m3/s, it probably makes sense to revert this and stick with what EVE represents.

Since no one can agree whether /s or /h is better, why not have both?

Reverts the default display back to /s since that's what CCP has added and is shown in the client.

Moves /h into tooltips, so you can quickly see the /h rate if that's what you prefer.
If @blitzmann doesn't mind, can add a second line to the mining yield stats pane to add /h to be displayed below the /s.  So we'd effectively have both displayed in the GUI (without having to go to the tooltip).

Also, fixed the tooltips for the Mining Yield stats pane, so it shows something useful instead of just the same number again (now matches the misc tab tooltip).

Bit of housekeeping and cleanup (mining drones were their own `elif`, bit redundant).

### DIFF
--- a/gui/builtinStatsViews/miningyieldViewFull.py
+++ b/gui/builtinStatsViews/miningyieldViewFull.py
@@ -124,9 +124,9 @@ class MiningYieldViewFull(StatsView):
     def refreshPanel(self, fit):
         #If we did anything intresting, we'd update our labels to reflect the new fit's stats here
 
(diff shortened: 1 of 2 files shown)


blitzmann:
Thanks. This is fine for now. Sucks that CCP went the route of /s, but I think this is a good compromise

Issue #949, opened 2017-01-18
Damage Pattern Editor will not open


Cold-FuzZ:
Attempting to open the Damage Pattern Editor is Unsuccessful when using using Ctrl-D or clicking on it manually, as well as this I cannot use any Damage pattern profiles by right clicking on the incoming damage pattern - there is no drop down menu like there previously was.

Cold-FuzZ:
Just found other posts with the same issue, sounds like a  DB corruption, will send %userprofile%\.pyfa\saveddata.db file to sable.blitzmann@gmail.com, as reccomended previously.

blitzmann:
Thanks for the database, very helpful. The root cause of this is that you had a `NULL` value for the name of one of your profiles, and pyfa just freaked out not knowing how to handle that. 

This is what we should do to fix it:
* `NOT NULL` the name field
* Check to make sure the other fields can be NULL without breaking anything. If not, ensure that these are also `NOT NULL`
* We will have to make a migration script that modifies current user database to add the `NOT NULL`s required, and removes any offending records

This should fix that source of corruption

Ebag333:
The GUI already doesn't allow you to use nulls for names.  I'm guessing this is an artifact of something we're doing in code when manipulating these (adding/deleting/copying).

blitzmann:
> The GUI already doesn't allow you to use nulls for names. I'm guessing this is an artifact of something we're doing in code when manipulating these (adding/deleting/copying).

I know, this is something that may have slipped through the cracks. Not only that, I vaguely remember that there was the possibility of saving something without a name, and this might be a left over from that.

Regardless the cause, having the constraint on the database will ensure we don't have this happen again.

Cold-FuzZ:
Appreciate the response and the fix, it's now working as intended. Cheers !

PR #953, opened 2017-01-25
Catch scenario where Damage Profile and Target Resist names are missing

Ebag333:
See #949 for history.

Catches both null and empty names.  Sets name to `Unknown` if found.

### DIFF
--- a/eos/db/saveddata/databaseRepair.py
+++ b/eos/db/saveddata/databaseRepair.py
@@ -29,9 +29,6 @@ class DatabaseCleanup:
 
     @staticmethod
     def OrphanedCharacterSkills(saveddata_engine):
-        # Finds and fixes database corruption issues.
-        logger.debug("Start databsae validation and cleanup.")
-
         # Find orphaned character skills.
         # This solves an issue where the character doesn't exist, but skills for that character do.
         # See issue #917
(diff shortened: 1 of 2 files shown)


blitzmann:
This needs to be in a migration script, not one of our clean up scripts.

Ebag333:
This isn't migrating data, and it's not fixing the DB issues.  It's just getting users back on their feet just like the other cleanup scripts.

blitzmann:
> This isn't migrating data, and it's not fixing the DB issues. It's just getting users back on their feet just like the other cleanup scripts.

Which should happen in the migration script. I would rather have a migration script that fixes the database once and implements fixes on the database level to ensure it doesn't happen again than to have something the covers up the issue.

Ebag333:
The migration script can call this function to run a repair on the database.  Either way, this code has to be written.  Where we put it and what we call it doesn't really matter, the process and end result is the same.  There's no reason a migration script can't import this, there's a reason why it's called `databaseRepair` and why it's very standalone (has minimal imports).

The only reason we're calling the function in `prefetch` is because we need to run it at some point before the database is in use, it could have easily been added as an extension of the migration scripts rather than a stand alone module.  If we use it in a migration script, then we can yank the reference to it from `prefetch`.

blitzmann:
Okay, I can get behind that. 

> If we use it in a migration script, then we can yank the reference to it from prefetch.

Yes please, remove it from prefetch. I'll work on the migration maybe tonight :)

Ebag333:
>I'll work on the migration maybe tonight :)

IMHO the DB issues are more pervasive than just a single field.  There isn't a unified design behind the tables, there's tons of stuff that's half implemented, and the DB restraints aren't really there for most things.

I'd like for us to sit down and figure out a way to fix all of these issues, and create a single migration script.  It's not any more complex to migrate _all_ the tables instead of just a single table, so I'd rather do it in one fell swoop than doing all the work for one at a time.

To me the DB corruption is a separate (if related) issue, and since they can be handled separately might as well, especially since a proper DB migration will take significant effort and adding a check for this corruption was copy/pasta and tweak a query.

Anyway, I removed the reference out of `prefetch`.

FYI (and you probably noticed this already) but I abstracted out the actual query execution.  I wasn't thrilled with encapsulating that much code in a try/catch, and the bigger issues is that you don't know which of the queries failed (since we usually have 2 or 3).  So now the error handling is more granular, and it spits out exactly which query it failed on.

blitzmann:
> IMHO the DB issues are more pervasive than just a single field. There isn't a unified design behind the tables, there's tons of stuff that's half implemented, and the DB restraints aren't really there for most things.

Well, we can debate the structure of the database night and day. Fact remains that we're working off a codebase that was started 7-8 years ago, so yeah, it's gonna be funky. I've already identified a few things that will help corruption in the future (such as setting autoincrement for IDs), but I agree we should take a look and start adding more constraints where we need them, starting with the damage / target resist profiles. I am not a fan of doing a large batch of database changes at once though; I prefer to do these in iteration. Last thing I want is to fuck up a users database because one thing failed.

Ebag333:
>Well, we can debate the structure of the database night and day. Fact remains that we're working off a codebase that was started 7-8 years ago, so yeah, it's gonna be funky.

I don't think either of us disagrees about the state of things.  Or that they should get fixed.  Or even generally on how to do it.

Just the difference between tearing off the duct tape all at once, or very slooooowly. :)

blitzmann:
I changed it to a DELETE instead of an update. I don't know what multiples of the same name would do (if anything), and my thought it that users won't miss having it since they never had access to it in the first place. v0v

Ebag333:
Multiple names SHOULD be fine, but I see your point. I just hate deleting user data.

Issue #954, opened 2017-01-27
Mac deprecated version crashes on launch 


JoshuaNotowitz:
When I try to launch it it just bounces in the dock for a little while and then closes

Ebag333:
Can you please send me your saveddata database? ebagola@gmail.com 

If you see a log file there, see if there is anything logged.

JoshuaNotowitz:
Where do i find this file?

Ebag333:
https://github.com/pyfa-org/Pyfa/wiki/Database

Let me know if you have any trouble finding it.

JoshuaNotowitz:
Sent

Ebag333:
There is some database corruption.

```
2017-01-27 12:46:57,012 eos.db.saveddata.databaseRepair DEBUG    Running database cleanup for orphaned characters attached to fits.
2017-01-27 12:46:57,056 eos.db.saveddata.databaseRepair ERROR    Database corruption found. Cleaning up 5 records.
```

Next release of Pyfa will have some cleanup that it will perform.  Still looking over the log file, that has a ton of stuff in it that might be the cause for it failing to launch.

Also, dang.  I thought I had a lot of fits, your DB is twice my size.

JoshuaNotowitz:
yea that might contribute to why the non deprecated version seems to run so slow

Ebag333:
@blitzmann some of the relevant info from the logs.

All these tracebacks in the log are old (2015).  I want to still double check on them to make sure that we have a fix in place for them, since we've had some issues that go back that far that we still haven't been able to identify the cause of yet.  Skip past the tracebacks for the recent errors.

```
2015-11-04 11:20:34,592 STDERR                   ERROR    Traceback (most recent call last):
2015-11-04 11:20:34,593 STDERR                   ERROR      File "/Users/USER/Desktop/pyfa.app/Contents/Resources/lib/python2.7/wx/_core.py", line 14640, in <lambda>
2015-11-04 11:20:34,601 STDERR                   ERROR    lambda event: event.callable(*event.args, **event.kw) )
2015-11-04 11:20:34,601 STDERR                   ERROR      File "/Users/USER/Desktop/pyfa.app/Contents/Resources/gui/mainFrame.py", line 205, in ShowUpdateBox
2015-11-04 11:20:34,602 STDERR                   ERROR    dlg.Destroy()
[...]

PR #956, opened 2017-01-27
Wrap dlg.Destroy() in a Try/Except

Ebag333:
Seems we have on going issues with trying to destroy objects that don't exist.  See: #954

Simply went through `mainFrame.py` and wrapped all `dlg.Destroy()` in a try/except.

We only catch the exception if it's a dead object, if it throws a different error then it'll still stacktrace.

We also log when it happens, so hopefully sheds a bit more light into how often (and when) this happens.


blitzmann:
Merged the commit in question... github wants to merge a shit load more stuff. Closing.

PR #962, opened 2017-01-30
Logbook (Logging Improvements)

Ebag333:
This is a PR to explore improving logging in Pyfa, as currently our logging doesn't have very good coverage.

Logging in general isn't very friendly toward keeping the application fast and light weight.  Currently our logging solution is the built in `logging`, which is relatively slow as a solution.

To test this, I created a simple scenario where I looped 1 million times and spat out the time elapsed.  Essentially each test will write a message to sysout, or the log file, or both.

+ *Write to log*
1.9 seconds
+ *Printing to Console*
10.6 seconds
+ *logging* (prints to console and writes to log)
430 seconds
+ *logbook*  (prints to console and writes to log)
366.8 seconds

Clearly if we want speed our best solution is to write our own and keep it as simple as possible.  Both built in `logging` and `logbook` are very slow in comparison, even though `logbook` is about 15% faster.

However, @blitzmann made some good points about everything surrounding logging, and not reinventing wheels.

Here's a quick rundown of the advantages to `logbook` over `logging`.

You can configure logbook to completely ignore debug messages.  Even if debugging isn't turned out, with logging the debug statement is still ran.  This means debug messages will have a heavier impact in logging than logbook.

Logbook allows you to "bubble up" messages, so we can handle them at different levels.  Writing the log to memory is very light weight, writing to disk less so, and writing to the system console is very painful (see how printing took 10 times as long as a file write).  In addition to being able to handle different log levels differently, such as always sending debug to log while printing other levels, logbook has two handlers that we can explore. 

`TestHandler` will keep logs in memory (not write to disk).  This could be super useful if we wanted to write a GUI pane that exposed the log (without having to go to the actual log file).

`FingersCrossedHandler` is even more interesting, it wraps another handler and will log everything in memory until a certain level (action_level, defaults to ERROR) is exceeded. When that happens the fingers crossed handler will activate forever and log all buffered records as well as records yet to come into another handler which was passed to the constructor.  So for non-debugging use, we could operate in this mode, and then when we hit an error, all the debug messages get flushed to disk.

Between handling different logging levels differently (keeping debug from printing to sysout), and being able to cache to memory, we can _seriously_ lighten the load that logging currently is.

While I don't think that logging is a terribly heavy load on Pyfa, right now it's barely used.  We currently only importing logging in 33 files, which means that there's huge swaths of Pyfa that have no logging.

While I don't think we need to go to an extreme (logging wouldn't be particularly useful in most effects, for example), more through logging would give us a lot more info and help us debug scenarios where we have weirdness.  If we use the built in logging function, we're stuck with pretty much having it on or off, on with a performance hit or off and lose the data.

I think by using the `FingersCrossedHandler` we can have the best of both worlds. 

Right now I'm simply working on implementing like for like, so we don't have any of the extra benefits yet.  But the nice thing is once we have logging replaced, we can easily modify our logging setup in `pyfa.py` to point to the other handlers.

### DIFF
--- a/config.py
+++ b/config.py
@@ -8,6 +8,7 @@ logger = Logger(__name__)
 try:
     import configforced
 except ImportError:
+    logger.warning("Failed to import: configforced")
     configforced = None
 
 # Turns on debug mode
--- a/config.py
+++ b/config.py
@@ -1,10 +1,8 @@
 import os
 import sys
 
-# TODO: move all logging back to pyfa.py main loop
-# We moved it here just to avoid rebuilding windows skeleton for now (any change to pyfa.py needs it)
-import logging
-import logging.handlers
+from logbook import Logger
+logger = Logger(__name__)
 
 # Load variable overrides specific to distribution type
 try:
@@ -29,23 +27,6 @@ savePath = None
 saveDB = None
 gameDB = None
 
-
-class StreamToLogger(object):
-    """
-    Fake file-like stream object that redirects writes to a logger instance.
-    From: http://www.electricmonk.nl/log/2011/08/14/redirect-stdout-and-stderr-to-a-logger-in-python/
-    """
-
-    def __init__(self, logger, log_level=logging.INFO):
-        self.logger = logger
-        self.log_level = log_level
-        self.linebuf = ''
-
-    def write(self, buf):
-        for line in buf.rstrip().splitlines():
-            self.logger.log(self.log_level, line.rstrip())
-
-
 def isFrozen():
     if hasattr(sys, 'frozen'):
         return True
@@ -66,10 +47,7 @@ def defPaths(customSavePath):
     global gameDB
     global saveInRoot
 
-    if debug:
-        logLevel = logging.DEBUG
-    else:
-        logLevel = logging.WARN
+    logger.debug("Configuring Pyfa")
 
     # The main pyfa directory which contains run.py
     # Python 2.X uses ANSI by default, so we need to convert the character encoding
@@ -97,19 +75,9 @@ def defPaths(customSavePath):
         os.environ["REQUESTS_CA_BUNDLE"] = getPyfaPath(certName).encode('utf8')
         os.environ["SSL_CERT_FILE"] = getPyfaPath(certName).encode('utf8')
 
-    loggingFormat = '%(asctime)s %(name)-24s %(levelname)-8s %(message)s'
-    logging.basicConfig(format=loggingFormat, level=logLevel)
-    handler = logging.handlers.RotatingFileHandler(getSavePath("log.txt"), maxBytes=1000000, backupCount=3)
-    formatter = logging.Formatter(loggingFormat)
-    handler.setFormatter(formatter)
-    logging.getLogger('').addHandler(handler)
-
-    logging.info("Starting pyfa")
 
     if hasattr(sys, 'frozen'):
-        stdout_logger = logging.getLogger('STDOUT')
-        sl = StreamToLogger(stdout_logger, logging.INFO)
-        sys.stdout = sl
+        pass
 
         # This interferes with cx_Freeze's own handling of exceptions. Find a way to fix this.
         # stderr_logger = logging.getLogger('STDERR')
--- a/config.py
+++ b/config.py
@@ -2,13 +2,13 @@ import os
 import sys
 
 from logbook import Logger
-logger = Logger(__name__)
+logging = Logger(__name__)
 
 # Load variable overrides specific to distribution type
 try:
     import configforced
 except ImportError:
-    logger.warning("Failed to import: configforced")
+    logging.warning("Failed to import: configforced")
     configforced = None
 
 # Turns on debug mode
@@ -48,7 +48,7 @@ def defPaths(customSavePath):
     global gameDB
     global saveInRoot
 
-    logger.debug("Configuring Pyfa")
+    logging.debug("Configuring Pyfa")
 
     # The main pyfa directory which contains run.py
     # Python 2.X uses ANSI by default, so we need to convert the character encoding
--- a/eos/db/__init__.py
+++ b/eos/db/__init__.py
@@ -25,8 +25,8 @@ from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import pool
 from logbook import Logger
 
-log = Logger(__name__)
-log.debug("Initializing")
+logger = Logger(__name__)
+logger.debug("Initializing")
 
 import migration
 from eos import config
@@ -53,6 +53,7 @@ try:
         "SELECT `field_value` FROM `metadata` WHERE `field_name` LIKE 'client_build'"
     ).fetchone()[0]
 except:
+    logger.warning("Missing gamedata version.")
     config.gamedata_version = None
 
 saveddata_connectionstring = config.saveddata_connectionstring
--- a/eos/db/__init__.py
+++ b/eos/db/__init__.py
@@ -23,7 +23,10 @@ from sqlalchemy import MetaData, create_engine
 from sqlalchemy.orm import sessionmaker, scoped_session
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import pool
+from logbook import Logger
 
+log = Logger(__name__)
+log.debug("Initializing")
 
 import migration
 from eos import config
--- a/eos/db/__init__.py
+++ b/eos/db/__init__.py
@@ -25,8 +25,8 @@ from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import pool
 from logbook import Logger
 
-logger = Logger(__name__)
-logger.debug("Initializing")
+logging = Logger(__name__)
+logging.debug("Initializing")
 
 import migration
 from eos import config
@@ -53,7 +53,7 @@ try:
         "SELECT `field_value` FROM `metadata` WHERE `field_name` LIKE 'client_build'"
     ).fetchone()[0]
 except:
-    logger.warning("Missing gamedata version.")
+    logging.warning("Missing gamedata version.")
     config.gamedata_version = None
 
 saveddata_connectionstring = config.saveddata_connectionstring
--- a/eos/db/migration.py
+++ b/eos/db/migration.py
@@ -37,7 +37,7 @@ def update(saveddata_engine):
         for version in xrange(dbVersion, appVersion):
             func = migrations.updates[version + 1]
             if func:
-                logger.info("Applying database update: %d", version + 1)
+                logger.info("Applying database update: {0}", version + 1)
                 func(saveddata_engine)
 
         # when all is said and done, set version to current
--- a/eos/db/migration.py
+++ b/eos/db/migration.py
@@ -1,11 +1,11 @@
-import logging
+from logbook import Logger
+logger = Logger(__name__)
+
 import shutil
 import time
 
 import config
 import migrations
-
-logger = logging.getLogger(__name__)
 
 
 def getVersion(db):
--- a/eos/db/migration.py
+++ b/eos/db/migration.py
@@ -1,5 +1,5 @@
 from logbook import Logger
-logger = Logger(__name__)
+logging = Logger(__name__)
 
 import shutil
 import time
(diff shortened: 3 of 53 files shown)


Ebag333:
Okay, this is really cool.

We now handle debug mode.  When _not_ in debug mode, we write to `Pyfa.log` (yes the name is changed from `log.txt`...it bugged me that it didn't end in `.log` so my logging tools would just work).  We _only_ write to the log if we catch an error or higher, anything below error simply gets cached in memory.

When in debug mode, we act as we do now, write to a log (though we write to `Pyfa_debug.log) and output to the console.

We're also redirecting `stdout` as warning and `stderr` as critical.  Since traditionally we've just done Debug/Info/Error, this means that they will generally have their own level and we can handle them separately if need be.  It also means we'll catch all the prints scattered around, and devs can be lazy and just `print()` something for debugging.

If we want to add another switch, we could disable writing to the console when in debug  mode.  Or separate out debug mode from our logging mode.

This is really amazing though,  we'll now be able to get our users to give us debug level logs if something breaks without them having to do a thing.  And if there are no errors, it only gets written to memory so we have very little performance penalty.

Ebag333:
>We're also redirecting stdout as warning and stderr as critical.

So apparently we're not redirecting stderr due to cx.freeze, which I believe is what you were thinking about @biltzmann .  We need to fix this.

Since we're going to have to rebuild the skeleton anyway (I know I know, it's a PITA), moving it to `Pyfa.py` and running it at a higher level _might_ help.  We'll have to test....

blitzmann:
This all looks interesting. I am still interested in log rotating, so look into support for that.

>  We need to fix this.

How do you propose we fix it? I don't know how cx_freeze handles it. I never looked into it and am not too sure what exactly is happening there. But we need to ensure that this stuff absolutely works with a packages version.

I would recommend creating a small wx test app that utilizes logbook and and bundle it with cx_freeze so that you can have an easy test bench

Ebag333:
>I am still interested in log rotating, so look into support for that.

Already taken care of. ☺️

>How do you propose we fix it? I don't know how cx_freeze handles it. I never looked into it and am not too sure what exactly is happening there. But we need to ensure that this stuff absolutely works with a packages version.

Agreed. I know nothing about cx freexe, so that is a next step

Ebag333:
My evil plan to save ~the world~ Pyfa logging:

- [ ] Get Logbook working
- [ ] Get debug mode working
- [ ] Get  user mode working (writing to memory, dumping when errors hit)
- [ ] Replace all existing logging with logbook
- [ ] Add logging to all exceptions where uncaught errors might be found (so many....)
- [ ] Add logging to a bunch of key functional places in Pyfa, so we can start to get a better bigger picture of how it runs.
- [ ] Get it working in a release state

Ebag333:
One thing that this does _not_ help us with.

Logged messages inside threads are still blackholed. :(

Ebag333:
I don't know how much is from swapping out logging and how much is from all the _other_ improvements we made, but dev branch is running significantly snappier than it was (I suspect it's about 98% the other improvements, but logbook does feel snappier).

I've started populating some key points in the start up process to try and get a better idea for where we're spending out time.  This is running in full potato mode (full debugging on), and still only took 13 seconds to boot up.  That's about half of what it used to be.

Anyway, most of the start up time is still spent trying to get what categories we have fits for.  Below is what the log looks like if you're in full debug mode.

```
[2017-01-30 22:11:13.282000] INFO: __main__: Starting Pyfa
[2017-01-30 22:11:13.282000] DEBUG: __main__: Import wx
[2017-01-30 22:11:13.459000] DEBUG: __main__: Import eos.db
[2017-01-30 22:11:13.587000] DEBUG: eos.db: Initializing
[2017-01-30 22:11:14.127000] DEBUG: service.prefetch: Run database migration.
[2017-01-30 22:11:14.244000] DEBUG: service.prefetch: Import Required Database Values.
[2017-01-30 22:11:14.606000] DEBUG: service.prefetch: Starting database validation.
[2017-01-30 22:11:14.607000] DEBUG: eos.db.saveddata.databaseRepair: Start databsae validation and cleanup.
[2017-01-30 22:11:14.608000] DEBUG: eos.db.saveddata.databaseRepair: Running database cleanup for character skills.
[2017-01-30 22:11:14.615000] DEBUG: eos.db.saveddata.databaseRepair: Running database cleanup for orphaned characters attached to fits.
[2017-01-30 22:11:14.622000] DEBUG: eos.db.saveddata.databaseRepair: Running database cleanup for orphaned damage patterns attached to fits.
[2017-01-30 22:11:14.628000] DEBUG: service.prefetch: Completed database validation.
[2017-01-30 22:11:14.687000] INFO: gui.bitmapLoader: Using local image files instead of zip.
[2017-01-30 22:11:16.372000] DEBUG: __main__: Make sure the saveddata db exists
[2017-01-30 22:11:16.409000] DEBUG: __main__: Show GUI
[2017-01-30 22:11:16.410000] DEBUG: gui.mainFrame: Initialize MainFrame
[2017-01-30 22:11:16.423000] DEBUG: service.fit: Initialize Fit class
[2017-01-30 22:11:17.432000] DEBUG: gui.marketBrowser: Initialize marketBrowser
[2017-01-30 22:11:18.052000] DEBUG: gui.marketBrowser: Initialize marketTree
[2017-01-30 22:11:18.275000] WARNING: __main__: Missing icon file: C:\Users\Ebag333\Documents\GitHub\Ebag333\Pyfa\imgs\icons\.png
[2017-01-30 22:11:18.292000] DEBUG: gui.marketBrowser: Initialize ItemView
[2017-01-30 22:11:18.396000] DEBUG: gui.marketBrowser: Fill up recently used modules set
[2017-01-30 22:11:18.708000] DEBUG: gui.shipBrowser: Populate ship category list.
[2017-01-30 22:11:26.797000] DEBUG: __main__: Run MainLoop()
```

And here's the same file if you're a user:
```
 
```

No errors hit, so no log needs to be generated. 😁

Ebag333:
>How do you propose we fix it? I don't know how cx_freeze handles it. I never looked into it and am not too sure what exactly is happening there. But we need to ensure that this stuff absolutely works with a packages version.

This actually _might_ fix it.

Was having Pyfa crash without any errors because the log file it was trying to use was in use by another process (so it couldn't output anything).  With a try/except, we lose catching any stdout/stderr, but should keep trucking on.

```
        # Output all stdout (print) messages as warnings
        try:
            sys.stdout = LoggerWriter(logger.warning)
        except ValueError:
            logger.critical("Cannot access log file.  Continuing without writing stdout to log.")
        # Output all stderr (stacktrace) messages as critical
        try:
            sys.stderr = LoggerWriter(logger.critical)
        except ValueError:
            logger.critical("Cannot access log file.  Continuing without writing stderr to log.")
```

Need to test and see of course, but might possible be on to something there.

blitzmann:
Keep in mind: we cannot change the `mac-deprecated` build. We have been able to work around this for things like CREST by conditionally importing things, but since logging is going to be all over the place, this is unfeasible.

Granted, as stated before, `-deprecated` is consistently <1% of the total downloads (mac being 4-5% and windows being ~92%), so it's not that big of a deal. I want to focus on OS X stability as my next major push, so this might not be a problem for much longer... but regardless.

Worst case, I may be able to add the package in a dirty way - it's all just files.

Ebag333:
>Keep in mind: we cannot change the mac-deprecated build.

We  may just need to create a Mac depreciated fork, and basically put it on life support.  It's not fair to block 99% of the folks out there from new features because of the 1%.

There's a number of PRs that will require rebuilding the skeleton.  It's about time for that. :)

blitzmann:
Last resort we're going to nuke it. I don't want to maintain a fork of it. Fuck that.

But I will be looking into just stuffing the module in site_packages. Hell, it may even work for the windows build, which would help us avoid a rebuild for a long as possible

Ebag333:
>Last resort we're going to nuke it. I don't want to maintain a fork of it

The idea wasn't to maintain it, but provide a place where it existed so if someone wanted to run builds off it they could.  But given the low number of downloads, it's probably not necessary.

>But I will be looking into just stuffing the module in site_packages. Hell, it may even work for the windows build, which would help us avoid a rebuild for a long as possible

It's not just the module at play here.

We're anchoring the module in `pyfa.py`, which I do believe you stated would necessitate a rebuild of yonder skeleton.  By anchoring the current logging at a higher level in `config.py` we're definitely missing some stuff, and this might potentially help with some of the issues we were having with cx_freeze.

Plus, there's a number of `TODO`s in the code base referencing moving things into `pyfa.py`, this is an opportunity to do so.

I know you hate rebuilding the skeleton, but we might be able to enlist some help with that from some folks in devfleet who have some Python module building experience.  (That would certainly not be me.)

Ebag333:
@blitzmann overall  I'm rather happy with the state that this is in.  Would appreciate you giving it a spin and seeing if you can break it.

I've added boatloads of debug logging, we still only touch a _tiny_ amount of the code base.  The big one was getting logging added to exception handlers, I caught most but I'm sure that there is more to b e done.

The one thing we may want to consider is breaking logging levels away from debug mode.  We may also want a flag to skip logging to the console entirely, for any devs who want to just tail a file rather than watching output in the console.

I'd like to add even more logging, but I think at this point we've hit a lot of the high points, and that can be a future "as we go" improvement.

blitzmann:
Just FYI, pyfa.py used to have that restriction, no longer does
:)

StinGer-ShoGuN:
Hey guys, just read the discussion and I feel like there is something missing in the possible solutions in the beginning: write a C/C++ custom logger instead of a pure Python one. I won't guarantee it will improve performances significantly, but it could.

I know it looks terrible in the first place because you would have to generate the required *.pyd, *.so or *.dylib files to run on Windows, Linux and Mac, double time for the 32/64 bits arch. But this is all about build scripts. 👅 
More seriously, with a well defined API from the beginning, it could be a drop-in replacement for an existing pure pythonic implementation (like cElementTree was for ElementTree). And if put in a submodule or something, no need to regenerate the required files for every release.

Final words: I would not recommend tools such as SWIG or cython. They're good for big complex projects (to my mind). But for small pieces of software, I find they are more complicating things than easing them.

blitzmann:
That... just seems like a lot of work for negligible performance increase. I mean, I'm still of the opinion that the standard  logging library isn't particularly resource intensive. >_>

Ebag333:
>I feel like there is something missing in the possible solutions in the beginning: write a C/C++ custom logger instead of a pure Python one. I won't guarantee it will improve performances significantly, but it could.

>That... just seems like a lot of work for negligible performance increase. 

I appreciate the thought @StinGer-ShoGuN and I don't disagree, as I've shown above simply printing and writing is a tiny fraction the time that using Pythons built in logging takes (3%).  I'm sure a C/C++ logger would be even faster, though I don't think it'd as big of a difference as the one between logging and direct print/write.

I'm in line with @blitzmann on this one though, a custom logger could be easily written (or adapted) in Python for the majority of the savings, and a C/C++ one would mean a separate code base to maintain.

>I mean, I'm still of the opinion that the standard logging library isn't particularly resource intensive. >_>

Says you, who also doesn't take several minutes to launch Pyfa. :)

Not everyone is lucky enough to run on your magic box, @blitzmann, and while it's not a issue for you, for others who have resource constraints it very well can be.

I'm really not worried about resources with the _current_ logging setup.  But once we go through and start adding tons of additional logging, it very well _can_ become resource intensive, since you can easily generate thousands of logs a minute.  If you don't have a SSD and fast processor, then this will quickly become the choke point.

To me, there's two big reasons for using Logbook over Logging:
1. It's a bit of a faster solution out of the box, and we can tweak it to improve that significantly (disabling writes to console for example).  As we add more logging the difference between the two solutions will become larger and more noticeable, especially for users on spinning disks.
2. We can get debug level logs from users (including a backlog of events which lead up to the error), without the user having to do a single thing.  I'm absolutely in love the with `FingersCrossed` handler, as far as I'm concerned it's the best thing since sliced bread.  And since it's logging to memory, we're essentially eliminating entirely the resource cost.

To replicate all that functionality in C/C++ is certainly possible but....we're only going to be generating actual log files if something goes wrong, so in your standard use case logging is essentially disabled.  So the extra speed/time savings of a C/C++ log handler is mostly wasted.  It took me all of about 2-3 hours to actually convert logging to logbook, most of the time I've spent on this was customizing it, adding better handling for error scenarios, and adding a ton of logging.

blitzmann:
Gonna look into jacking this into the skels tonight hopefully, because the more i look at this the more I like it. Probably won't make it into the release next week, but if it works the way I think it will we won't have to wait until a binary rebuild.

Ebag333:
Closing as this is resubmitted in a rebased version in #986

PR #963, opened 2017-01-30
Skip character stuff if there is no fit.

Ebag333:
@blitzmann you'll want to review this one and make sure it's not breaking anything.

On a normal fit calc (with a ship, char, etc) `fitChanged` takes about 100-200 ms to run (in full debugging potato mode).  On an _empty_ fit, it takes 1000-1500 ms. (Why we run a full recalc against empty fits is another issue....)

Seems there's no sense in doing a bunch of character stuff for a fit that doesn't exist.

_Should_ be ready to be merged, if you don't see any issues with it.

### DIFF
--- a/gui/mainMenuBar.py
+++ b/gui/mainMenuBar.py
@@ -164,17 +164,19 @@ class MainMenuBar(wx.MenuBar):
 
     def fitChanged(self, event):
         enable = event.fitID is not None
-        self.Enable(wx.ID_SAVEAS, enable)
-        self.Enable(wx.ID_COPY, enable)
-        self.Enable(self.exportSkillsNeededId, enable)
 
-        sChar = Character.getInstance()
-        charID = self.mainFrame.charSelection.getActiveCharacter()
-        char = sChar.getCharacter(charID)
+        if enable:
+            self.Enable(wx.ID_SAVEAS, enable)
+            self.Enable(wx.ID_COPY, enable)
+            self.Enable(self.exportSkillsNeededId, enable)
 
-        # enable/disable character saving stuff
-        self.Enable(self.saveCharId, not char.ro and char.isDirty)
-        self.Enable(self.saveCharAsId, char.isDirty)
-        self.Enable(self.revertCharId, char.isDirty)
+            sChar = Character.getInstance()
+            charID = self.mainFrame.charSelection.getActiveCharacter()
+            char = sChar.getCharacter(charID)
+
+            # enable/disable character saving stuff
+            self.Enable(self.saveCharId, not char.ro and char.isDirty)
+            self.Enable(self.saveCharAsId, char.isDirty)
+            self.Enable(self.revertCharId, char.isDirty)
 
         event.Skip()


blitzmann:
>fitChanged takes about 100-200 ms to run (in full debugging potato mode). On an empty fit, it takes 1000-1500 ms. 

cAN YOU BE MORE DETAILED ABOUT WHY AN EMPTY FIT TAKES LONGER?

Ebag333:
>cAN YOU BE MORE DETAILED ABOUT WHY AN EMPTY FIT TAKES LONGER?

Not when you yell at me. :(

For this code block:
```
    def fitChanged(self, event):
        logger.debug("Fit Changed Triggered")
        enable = event.fitID is not None

        logger.debug("self.Enable x3")
        self.Enable(wx.ID_SAVEAS, enable)
        self.Enable(wx.ID_COPY, enable)
        self.Enable(self.exportSkillsNeededId, enable)

        logger.debug("Get Character Instance")
        sChar = Character.getInstance()
        logger.debug("Get Active Character ID")
        charID = self.mainFrame.charSelection.getActiveCharacter()
        logger.debug("Look up character from ID")
        char = sChar.getCharacter(charID)

        logger.debug("self.Enable x3")
        # enable/disable character saving stuff
        self.Enable(self.saveCharId, not char.ro and char.isDirty)
        self.Enable(self.saveCharAsId, char.isDirty)
        self.Enable(self.revertCharId, char.isDirty)

        logger.debug("event.Skip")
        event.Skip()
        logger.debug("Fit Changed Triggered - END")
```

You get these results:
```
[2017-01-31 01:47:17.998000] DEBUG: gui.mainMenuBar: Fit Changed Triggered
[2017-01-31 01:47:17.999000] DEBUG: gui.mainMenuBar: self.Enable x3
[2017-01-31 01:47:18.001000] DEBUG: gui.mainMenuBar: Get Character Instance
[2017-01-31 01:47:18.003000] DEBUG: gui.mainMenuBar: Get Active Character ID
[2017-01-31 01:47:18.006000] DEBUG: gui.mainMenuBar: Look up character from ID
[2017-01-31 01:47:18.011000] DEBUG: gui.mainMenuBar: self.Enable x3
[2017-01-31 01:47:18.449000] DEBUG: gui.mainMenuBar: event.Skip
[2017-01-31 01:47:18.466000] DEBUG: gui.mainMenuBar: Fit Changed Triggered - END
```

So this time it didn't take a full second, but "only" half a second.  Still long for an empty fit.

Looks like specifically it's the character save that's causing it to be slow, which makes sense (also makes sense why it's fairly variable since it's writing to disk).

In comparison, here's what happens when you  actually run it against a legitimate fit:
```
[2017-01-31 02:01:01.614000] DEBUG: gui.mainMenuBar: Fit Changed Triggered
[2017-01-31 02:01:01.615000] DEBUG: gui.mainMenuBar: self.Enable x3
[2017-01-31 02:01:01.615000] DEBUG: gui.mainMenuBar: Get Character Instance
[2017-01-31 02:01:01.616000] DEBUG: gui.mainMenuBar: Get Active Character ID
[2017-01-31 02:01:01.616000] DEBUG: gui.mainMenuBar: Look up character from ID
[2017-01-31 02:01:01.616000] DEBUG: gui.mainMenuBar: self.Enable x3
[2017-01-31 02:01:01.676000] DEBUG: gui.mainMenuBar: event.Skip
[2017-01-31 02:01:01.677000] DEBUG: gui.mainMenuBar: Fit Changed Triggered - END
```

Again saving it takes the longest, but now it's _much_ shorter.

blitzmann:
> Again saving it takes the longest, but now it's much shorter.

Saving what?

I'm looking at the code, and for me it takes 10-20ms regardless of if it's an empty fit or not (granted you and I always have vastly different performance metrics) I am not seeing where it's saving anything or writing anything to disk, hence why I'm unsure of the reasoning behind the discrepancy between empty / non-empty fit.

That being said, this pull request is broken in that we now only run the code if we change to a page with a fitting loaded. If we change to an empty tab, this code doesn't run and thus the menu items don't disable, which is the entire point to all that code.

So, I'm a bit confused??

blitzmann:
Also, when I asked for more details, I was more gearing towards information on which call seems to be taking longest. For you, it seems like it's this chunk: 

```
        self.Enable(self.saveCharId, not char.ro and char.isDirty)
        self.Enable(self.saveCharAsId, char.isDirty)
        self.Enable(self.revertCharId, char.isDirty)
```

Seems to take 0.5 s for you... but why? And why does it take that long on an empty fit vs non empty. Do there need to be performance tweaks for `isDirty` or `ro` properties? I was hoping you could investigate down that path (since I can't ever reproduce these performance issues)

thanks

Ebag333:
>That being said, this pull request is broken in that we now only run the code if we change to a page with a fitting loaded. If we change to an empty tab, this code doesn't run and thus the menu items don't disable, which is the entire point to all that code.

What exactly is broken when this code is disabled for an empty fit?  I didn't notice anything.

>Seems to take 0.5 s for you... but why? And why does it take that long on an empty fit vs non empty. Do there need to be performance tweaks for isDirty or ro properties? I was hoping you could investigate down that path (since I can't ever reproduce these performance issues)

To narrow it down down a  bit further, the specific line that is the kid eating paste in the back of the class is:
`self.Enable(self.saveCharId, not char.ro and char.isDirty)`

And more specifically if we run `test = char.ro` this one line will take at least 450 ms to run.

So that seems to be the culprit.

I'll poke into the `ro` property more when I get a chance.

>I'm looking at the code, and for me it takes 10-20ms regardless of if it's an empty fit or not (granted you and I always have vastly different performance metrics) 

I know it's Python, so were going to have widely different metrics for different users, but you're definitely on the fast end of things.  This doesn't seem to be an aberration for what I'm seeing, someone just posted on an issue with the same sort of experience I get.

Now if we can just figure out what makes you so special. :)

blitzmann:
> What exactly is broken when this code is disabled for an empty fit? I didn't notice anything

The menu items that need to be disabled aren't disable. So if there is an empty tab, export fitting would be enabled which males no sense and probably would cause errors.

That is the whole point to the code. The enabled variable determines if we switch those menu items on or off. If we never run that code, the menu items will remain in their previous do state

It would seem to me that you should look at ro property.  Don't take a sledgehammer to it when you need a scalpel,  etc

Ebag333:
Closing as requires more investigation

PR #967, opened 2017-02-03
Import Cleanup  III

Ebag333:
Cleans up some more imports where we're doing `import *` and packaging apps don't really understand them.

Also resubmits a few files because .gitattributes fixing line endings.

@blitzmann this is ready to be merged, and shouldn't be dependant on anything else.


Ebag333:
`moduleGlobalAmmoPicker`

This one seems to work fine.  Any idea of the history behind it @blitzmann and why it's not enabled?

blitzmann:
Just FYI, taking a look at the code, this PR will most likely be declined. `import *` is useful in the cases addressed in this PR and explicitly defining them is a hassle that should be unnecessary.

Furthermore... why did you get rid of the connection strings? These are helpful when I import `eos` via interpreter :(

Ebag333:
>Just FYI, taking a look at the code, this PR will most likely be declined. import * is useful in the cases addressed in this PR and explicitly defining them is a hassle that should be unnecessary.

I'd give the following reasons why we should avoid it.
1. Breaks pyCharm inspection (means it's harder to find errors, and when you do something bad, like shadow and overwrite an object, you have no idea you did that.  This happened a few times with Fits vs Fits vs Fits.)
2. Breaks compiling/build tools.  Programmatic references (such as  `eos.effects`) are one thing, but this is a different beast.  Build tools can't follow the references, so while Python might be able to figure out what to import with the `__all__`, not all other tools can.
3. Fails Tox.  Right now we have import checks disabled because of stuff like this. (Which isn't good, because imports are a quick and easy way to break lots of your code, especially with the dangers we have of shadowing existing classes.)
4. Little bit more difficult for new devs to understand what's going on.  Yes it's easier to write, but its also not explicit, and anyone who doesn't understand how `__all__` works won't understand the import statement.
5. It means that your import statement relies on multiple files.  Plus, if the `__all__` object doesn't exist in `init.py` doesn't exist then you have no idea what you're actually importing.

And the reasons for it:
1. It's faster to write? Debatable, since you have to modify two files initially, and when adding additional imports it's actually slightly more work to update the `__all__` than it is the import statement.

If you have other reasons why we _should_ do this, then I'd love to hear them.  But from everything I've read, there's a million and one reasons why we shouldn't do this, most of which boils down to `explicit is better than implicit`.

>Furthermore... why did you get rid of the connection strings? These are helpful when I import eos via interpreter :(

Because it's a redefinition of what's in `config.py` and not used anywhere.  If they belong in Eos then they should be in Eos, not in `config.py`.  (Or, hey, maybe `eos.config`! :) )

Shadowing is bad, and we have a ton of it.  It'll take a bunch of refactoring to actually remove it, but we should avoid it whenever possible.

Plus, the build choked on it.  I don't recall off hand what the error was, but it didn't like it.  Since it broke the build process, and since it's not used anywhere...... 🔪

blitzmann:
We've been through many of these points. 

>Breaks pyCharm inspection (means it's harder to find errors, and when you do something bad, like shadow and overwrite an object, you have no idea you did that. This happened a few times with Fits vs Fits vs Fits.)

I don't know why you keep thinking it breaks pyCharm inspection. pyCharm knows how to handle `import *`. What you're talking about is having different classes that are named the same. This has nothing to do with `import *`. Please do not assume "`from services import *` breaks because `service.Fit` has the same class name as `eos.saveddata.Fit` and breaks inspection, thus all `import *` is bad. I would like to see a detailed example of how `import *` is bad. 

> pyCharm also has a nifty new feature where you can ctrl-click on things

This is not a new feature, I'm aware of it and use it all the time. `import *` does not break this.

Case in point: `eos.db.__init__` has this command: 
```
from eos.db.gamedata.queries import *
from eos.db.saveddata.queries import *
```

So now all queries can be accessed via `eos.db.<queryname>`. I can now ctrl click `eos.db.getCharacter` and go to it's definition. See: https://gyazo.com/f54bbd67ff9dcb63ec7eda3db5aa59eb

> Breaks compiling/build tools. Programmatic references (such as eos.effects) are one thing, but this is a different beast. Build tools can't follow the references, so while Python might be able to figure out what to import with the __all__, not all other tools can. We can probably handle most of these by using hidden imports (as we did for effects), but then we have to add every single file to the hidden import list. So that's even more work. And there's no guarantee hidden imports will work. Worked fine for effects (gonna be fun maintaining that!), did not work at all for conversions.

As stated before, if they don't work by default, then there is probably some sort of configuration. I almost 100% certain that `import *` should work fine, but with the use of `__all__` I'm not sure. Then again, even if the packaging software doesn't understand `__all__`, it would fallback to it's default and include all files in that module. The `__all__` would then be handled by the python interpreter that the packager bundles.

That being said, wtf: https://github.com/Ebag333/Pyfa/blob/Packaging/pyfa.spec#L24 This is not what I meant when I suggested using this. `eos.effects` is a module, so simply adding `eos.effects` should work. It's like you're trying to make it harder on yourself... ;)

> Fails Tox. Right now we have import checks disabled because of stuff like this. (Which isn't good, because imports are a quick and easy way to break lots of your code, especially with the dangers we have of shadowing existing classes.)

Can't comment since I know next to nothing about tox, but if it can't handle `import *` then it sounds like it's a problem with them

> Little bit more difficult for new devs to understand what's going on. Yes it's easier to write, but its also not explicit, and anyone who doesn't understand how __all__ works won't understand the import statement.

Don't buy it. It took me about 10 minutes when first looking at the context menu code to see how it was being imported and registered. `__all__` is not a normal thing, but it's documented and can be figured out by a quick google search.

> It means that your import statement relies on multiple files. Plus, if the __all__ object doesn't exist in init.py doesn't exist then you have no idea what you're actually importing.

If `__all__` isn't there, then you import everything in that module. This is basic python dude.

Reasons for it:

There's no reason to change. The Pythonic Way. Keep It Simple, Stupid. Higher cost of maintaining it. I don't like it. etc

`import *` is there for a reason. There is no reason to explicitly import everything when you need to import literally everything.

Having said all that, if `__all__` truly is a problem, there are better ways to work around it (filename starts with `_` to remove it from 'public' scope, commenting out the line that registers the class with the parent, etc)

You're gonna be overruled on this one bud. 

> Because it's a redefinition of what's in config.py and not used anywhere. If they belong in Eos then they should be in Eos, not in config.py. (Or, hey, maybe eos.config! :) )

eos is designed to be more or less standalone. And it was in `eos.config`, not sure what you're getting at. pyfa, being higher level and utilizing eos, overwrites it's config so that they point to the correct location for pyfa's purposes. 

We have people who use pyfa specifically for eos - changing shit like that may very well break their apps, for no good reason.

Ebag333:
><thoughts>

You keep saying something along the lines of "if the tool doesn't work, then it's a problem with the tool not my code."

Sorry, I don't buy it. :)

>There's no reason to change. The Pythonic Way. Keep It Simple, Stupid. Higher cost of maintaining it. I don't like it. etc

The Pythonic way is simple, yes, but it's also explicit.  Good Python code should be near psuedocode, easy to read and understand.

Google `"explicit is better than implicit"` and you'll see that the first results are...Python. :)

Google `"from module import *"` and outside of docs/blogs telling you how it works, overwhelmingly the sentiment is "don't do this."

>That being said, wtf: https://github.com/Ebag333/Pyfa/blob/Packaging/pyfa.spec#L24 This is not what I meant when I suggested using this. eos.effects is a module, so simply adding eos.effects should work. It's like you're trying to make it harder on yourself... ;)

Do you think I didn't try that first?  It didn't work.  But hey, I guess it must be the tool that's broke so we should just not use it, eh?

We don't explicitly import effects (because they're programatically done), so this is the only way I can get them to import reliably.  If you can figure out another way, I'd love to see it.  Since the spec file is a python file, we can probably just os.walk the directory and handle it that way, but each file will still need to be added to hidden imports individually.  (Select all/copy/paste was faster to get it up and running.)

>eos is designed to be more or less standalone. And it was in eos.config, not sure what you're getting at. pyfa, being higher level and utilizing eos, overwrites it's config so that they point to the correct location for pyfa's purposes.

Sorry, I wasn't clear.  I meant move it entirely to `eos.config` if that's the level you want it at.

Either way, shadowing is bad and we shouldn't do it.  I don't care where it lives, but there shouldn't be an evil twin.

>We have people who use pyfa specifically for eos - changing shit like that may very well break their apps, for no good reason.

Who?  I asked around about this in Tweetfleet and found one person who used to.  They stopped using it, because the import mess in Eos made it more difficult to use than it was worth.

And shadowing is a good reason not to do it.

Ebag333:
>That being said, wtf: https://github.com/Ebag333/Pyfa/blob/Packaging/pyfa.spec#L24 This is not what I meant when I suggested using this. eos.effects is a module, so simply adding eos.effects should work. It's like you're trying to make it harder on yourself... ;)

Also, changed the .spec file to handle this programatically.  So no manual updating. \o/

Ebag333:
Closing this to handle via another PR.

PR #968, opened 2017-02-03
One Click Packaging

Ebag333:
So since I've heard rumors that rebuilding the skeleton is such a huge PITA, and I needed to try and troubleshoot some things surrounding logging and `stdout`/`stderr`, I went ahead and started playing around with freezing Pyfa.

After messing around with several, I  settled on pyinstaller.  This generates an executable file for Windows, Linux, and Mac, and the beautiful part of it is that it uses a spec file, so anyone can create a build of Pyfa with a simple command line ( assuming they have a proper build environment setup).

Advantages:
1. Faster/easier to make builds, lower bar of entry.
2. More automated build creation.
3. No more need for a skeleton, can more easily update referenced modules.
4. No need to monkeypatch things.
5. Theoretically faster, stated claims are about 10-15% faster than  simply freezing the code.

Disadvantages:
1. Can't monkeypatch live builds by replacing `.py` files in the zip.
2. Requires the base OS to make builds on (can't build for Mac on Windows, etc).

As this stands, Pyfa is mostly working (launches, loads fits, calculates them) but not every feature has been tested.

Anything that's a hidden import will need to be added to the `pyfa.spec`file manually (effects are a good example of this).

List of things left to be done:
- [x] Shows console window when executable is launched
- [x] Icon isn't set on the executable itself (can easily be manually changed after build is complete)
- [x] Need to include ,cer file (simple add to the spec)
- [ ] Need to build Mac and Linux spec files
- [ ] Conversions needs to be rewritten.  None of the package tools I've tried understand it (even when manually specifying the conversion files to import), pyCharm doesn't understand it, and it's generally meddlesome.  Currently the conversion code is commented out in this branch.

@blitzmann This PR is based on the logging PR #962 , so you'd want to merge that one first.  #967 is mostly based off this one, since a lot of the import statements mess up inspection (they needed to be cleaned up anyway), so might as well merge that one first.  #939 will also make builds more reliable by eliminating the third party imports, so....merge that one first. :)

There's less changes than you might think in this one, really just the last 4 commits.

___________________________________________________________________________

Some documentation for possible future Wikis:

**Linux**
_Setup/Configuration_
1. Install Linux (steps below use Ubuntu as a reference)
2. Escalate to su
`sudo su`
or prepend all commands below with
`sudo `
3. Update apt-get:
`apt-get update`
4. Install python if not already installed (2.7 recommended
5. Install pip
```
apt-get install python-pip python-dev build-essential 
pip install --upgrade pip 
pip install --upgrade virtualenv 
```
6. Install wxPython. (If 3.0 isn't available, you can fall back to 2.8, you do not need both.)
```
apt-get install python-wxgtk3.0

apt-get install python-wxgtk2.8
```
7. Create a directory to place the build files.  (I used `/tmp/Pyfa_build`)
8. Change directory to the one you created above.
9. Upload all files into the directory (can be done via git or simply copying the files)
10. Install Pyfa requirements.  
If you get any memory errors, use `--no-cache-dir`.
If you get build errors (`"python setup.py egg_info" failed`) try installing `libfreetype6-dev`.
```
pip install -r requirements.txt
pip install -r requirements_build_linux.txt
```
```
apt-get install libfreetype6-dev
pip --no-cache-dir install -r requirements.txt
```

_Build Process_
1. From within the same directory as `pyfa.py` and `pyfa.spec` run:
`pyinstaller --clean --noconfirm --windowed pyfa.spec`

Notes:
1. Because Linux doesn't bundle all the libraries in the package, it may not be backwards compatible.  It _should_ be forward compatible, but it's recommended to build on the oldest version you want to support.

.

**Windows**
_Setup/Configuration_
1. Install Windows (steps below use Windows 10 as a reference)
2. Install python if not already installed (2.7 recommended
3. Install wxPython (version 2.8)
4. Install Visual Studio (there is a free community edition)
5. Create a directory to place the build files.  (I used `/tmp/Pyfa_build`)
6. Change directory to the one you created above.
7. Upload all files into the directory (can be done via git or simply copying the files)
8. Install Pyfa requirements:
```
pip install -r requirements.txt
pip install -r requirements_build_windows.txt
```

_Build Process_
1. From within the same directory as `pyfa.py` and `pyfa.spec` run:
`pyinstaller.exe --clean --noconfirm --windowed --upx-dir=.\scripts\upx.exe pyfa.spec`


blitzmann:
Let me be completely honest: I am not a big fan of changing over packing solutions, and will not do so without good reason (for example, huge speed gains). Yes, this has the advantage to be super simple, but it comes at the high cost of easily creating new builds, which is currently one command for me in Linux and boom. The fact that we will now need to log into a OS X VM (which for me constantly break), for possible more than one OS X version (considering I want to look at why 10.11+ sucks so bad), this is a huge disadvantage to us as developers. The current 'build' process is clunky, but it works, and as a new developer someone should be able to figure it out by looking at the code (and running the commands found on the wiki for making new builds).

Speaking of linux, we most likely won't be doing installers for that platform as it's unnecessary (IMO) and I don't know what goes into it considering all the different distros. I would need more input from linux users / developers.

I will be asking folks to do A/B testing with a pyinstaller-produced package to see if there are any speed gains. If there are none, I am not sure if I want to hassle with changing packaging systems.

As an aside, the package that you sent me on Slack does not work, and also shows a console window with some sort bootloader message. I assume this can be turned off in production builds. Please send me a working example and I'll try to get some people on it to test.

Ebag333:
>As an aside, the package that you sent me on Slack does not work, and also shows a console window with some sort bootloader message. I assume this can be turned off in production builds. Please send me a working example and I'll try to get some people on it to test.
>EDIT: the reason it wasn't working is due to the alphaCloneID not being in my user database on this machine. It does in fact run. :)

HAH! So you yourself replicated that issue I put in.

I'm 98% sure that the branch has everything you added for Alpha clones, unless you snuck something else in.

Also, that build didn't have hidden imports fixed in it yet, so no effects for you! [/EffectNazi]

>I will be asking folks to do A/B testing with a pyinstaller-produced package to see if there are any speed gains. If there are none, I am not sure if I want to hassle with changing packaging systems.

That's the beautiful part, there's no real hassle.

Most of the changes here are stuff we should do _anyway_ (like fixing the imports).  I know you argue that `conversions` isn't broken, but the way it's implemented breaks a lot of things including pyCharm inspection, and we can super easily implement it in a way that inspection understands it _and_ is still easy to maintain.  Inspection to me is a super high priority, I'm 99.9% sure that a good number of long standing bugs got fixed when we did the import fixes PR, because inspection suddenly started pointing out problems.

The only thing specific to pyinstaller that this adds is `pyfa.spec`....which is basically done.  There might be a _bit_ more of a "hassle" finding other hidden imports (pyCrest might fall into that category for example), but the vast majority of the work is done.

> The fact that we will now need to log into a OS X VM (which for me constantly break), for possible more than one OS X version (considering I want to look at why 10.11+ sucks so bad), this is a huge disadvantage to us as developers.

See you're saying logging into VMs is a huge disadvantage.  I  say working around the limitations of the skeleton is a huge disadvantage.

I find your OSx VM breaking constantly as baffling as you find my slow times in Pyfa.  Should be able to snapshot it and have it permanently at a specific point in time.

You're already at 2 boxes (Windows and Linux), so to me adding another VM is less effort than jumping through the current hoops that we have, and dealing with all the potential issues that we're running into with the skeleton (I've heard you wonder more than once that you wonder what rebuilding the skeleton would do for OSx).

>Speaking of linux, we most likely won't be doing installers for that platform as it's unnecessary (IMO) and I don't know what goes into it considering all the different distros. I would need more input from linux users / developers.

I don't think that there should be _too_ much trouble with that.  Worst case scenario we simply do what most people do and release it based on the version we built it on.  If we want to go one step further and package it into an installer, then typically there's just two (.deb for Debian/Ubuntu and .rpm for RedHat/CentOS), though we probably don't need to go that far, just bundling it into a .tar.gz should be fine.

Ebag333:
>I will be asking folks to do A/B testing with a pyinstaller-produced package to see if there are any speed gains. If there are none, I am not sure if I want to hassle with changing packaging systems.

Fit calc times are highly variable so take this with a grain of salt but....

This is run off current EXE (pre-release build) and a quick build I made.  I loaded each (both in debug mode so I could get some meaningful log times) and ran through about 7 or 8 fits (same fits in same order on each one).  You can tell which is which by the timestamps (current build is on top). :)

Also, keep in mind that the new build is generating _much_ more in the way of logs, which all gets printed to the console (which is slow).  The current build generated 226 lines, the new build generated 5723.

2017-02-03 08:19:54,605 eos.saveddata.fit        DEBUG    Timer - Fit: 194, DPS - Done with fit calculation - 98.00ms (3317.00ms elapsed)
[2017-02-03 16:19:09.617000] DEBUG: eos.saveddata.fit: Timer - Fit: 194, DPS - Done with fit calculation - 368.00ms (1017.00ms elapsed)

2017-02-03 08:19:54,716 eos.saveddata.fit        DEBUG    Timer - Fit: 194, DPS - Done with fit calculation - 53.00ms (105.00ms elapsed)
[2017-02-03 16:19:10.277000] DEBUG: eos.saveddata.fit: Timer - Fit: 194, DPS - Done with fit calculation - 300.00ms (623.00ms elapsed)

2017-02-03 08:19:55,967 eos.saveddata.fit        DEBUG    Timer - Fit: 914, Retribution fit - Done with fit calculation - 0.00ms (27.00ms elapsed)
[2017-02-03 16:19:11.659000] DEBUG: eos.saveddata.fit: Timer - Fit: 914, Retribution fit - Done with fit calculation - 1.00ms (254.00ms elapsed)

2017-02-03 08:20:02,779 eos.saveddata.fit        DEBUG    Timer - Fit: 959, test copy - Done with fit calculation - 1.00ms (39.00ms elapsed)
[2017-02-03 16:19:20.474000] DEBUG: eos.saveddata.fit: Timer - Fit: 959, test copy - Done with fit calculation - 1.00ms (197.00ms elapsed)

It _seems_ like the new build is faster with complex fits (which the first one is), slower with empty/simple fits (the rest).  But it's not quite an apples to apples comparison since one is generating 25x the debug output.  Plus, fitting is highly variable, it'll jump hundreds of MS calculating the exact same fit on the exact same build running the exact same way.

I honestly couldn't tell you which one "feels" faster to me.

blitzmann:
https://github.com/pyfa-org/Pyfa/pull/968/commits/1a5ed3161c9776bb106c55de515d7d9189e76c14#diff-35e78883db663ee617e1604d8a81240e

We are not going to do this. These packaging softwares should work with dynamic imports. Please research this - pyinstaller probably has an option to force include directories to include all those files

blitzmann:
See:
https://pythonhosted.org/PyInstaller/when-things-go-wrong.html#listing-hidden-imports
https://pythonhosted.org/PyInstaller/hooks.html#hook-global-variables

you should be able to include them by adding 'eos.effects' or something. Absolutely no reason to include every friggen effect file in __init__ lol

Ebag333:
already done. 

I was seeing if it would understand `from eos.effects import *` with the all.... but no dice. 

Want the new build? :)

Ebag333:
https://drive.google.com/file/d/0B59pJ0pO7UMeUEhtRVlfclRxbmM/view?usp=drivesdk

Ebag333:
Well, got a bunch done.

>Shows console window when executable is launched

No longer shows. \o/

>Icon isn't set on the executable itself (can easily be manually changed after build is complete)

Now shows! \o/

>Need to include ,cer file (simple add to the spec)

Now shows up! \o/

Basically this is done for Windows.  So far everything I've thrown at it works fine, and `FingersCrossedHandler` works _beautifully_.  Plus we're redirecting `stdout` and `stderr` to the log file, so now if a user gets a traceback it'll get written to the log.  _AND_ as an added bonus we'll automagically include all the debug logs prior, as well as any logging that happens post.

It's really effing slick.

Lysarion:
> apt-get install python-wxgtk2.8
> E: Package 'python-wxgtk2.8' has no installation candidate

Just saying, current stable debian (jessie) does not include python-wxgtk2.8 in the repository (forcing the move to python3).

Ebag333:
>edit: Fixed by using python-wxgtk3.0

What I'm not sure about is if that's the Phoenix version or not. 

Phoenix isn't backwards compatible with the original wxPython. I looked into cont

Lysarion:
> apt-get install python-wxgtk3.0
did also install [python-wxversion](https://wxpython.org/docs/api/wxversion-module.html), if that may help you.

Ebag333:
The more I think about it, the more I'm sure that 3.0 is Phoenix.

All the documentation on  wxPython's site is for 2.8.

You can give it a go, but I don't think it'll work.  Please let me know if it does. :)

blitzmann:
@Ebag333 @Lysarion 

Let me try to clarify:

wxPython Pheonix is a re-write project that aims to support python 3 (and possibly be compatible with 2.7, though I'm not sure if this is a thing). It uses wxWidgets 3.0+ and currently (to my knowledge) doesn't have any stable builds available (although it is fairly stable last I heard)

Since pyfa is written for python 2, and the simple fact that it's 7+ years old, we use what's called wxPython "Classic". The version of wxPython is tied to the version of wxWidgets that it's compiled with (`python-wxgtk2.8` was built with wxWidget 2.8, while `python-wxgtk3.0` was built with wxWidgets 3.0). This doesn't have anything do do with the python version.

We added support for wxPython 3.0 over a year ago since repos like Ubuntu and Debian started including them as their defaults.

@Ebag333 protip: even though the Pheonix documentation is technically for Phoenix, which we don't use, I would say about 90% of it is applicable to classic as well. I find the new documentation easier to work with, so that's my primary goto unless I have trouble with it, then I fall back on the "proper" documentation.

Ebag333:
Thanks for the clarification.  It confuses the heck out of me since Phoenix uses a version number very similar to classic.

blitzmann:
> Phoenix uses a version number very similar to classic.

It kinda has to since it's tied to wxWidgets versioning. This is the whole reason they decided to name the rewrite Phoenix and retroactively label the old stuff "Classic" to differentiate the two codebases.

There is a Classic 3.0 and there is a Pheonix 3.0

I agree it's confusing when not understanding it, but it makes sense given the type of project. :)

Ebag333:
Closing as resubmitted with a cleaner PR.

PR #973, opened 2017-02-06
.gitAttributes strikes again!

Ebag333:
Just a couple more files that .gitAttributes is picking up.


Ebag333:
Closing as handle via another PR

PR #974, opened 2017-02-07
Merge mainMenuBar into mainFrame

Ebag333:
There is a hidden circular import reference between mainMenuBar and mainFrame (hidden because  of lazy imports).

Since the only thing to reference mainMenuBar is mainFrame.....merge them suckers together.

By eliminating importing the same things multiple times, shaved about 1-1.5 seconds off Pyfa loading time (in full debugging mode, with concurrancy graphs on).

Further improved loading times by changing several imports to be explicit instead of implicit/lazy (chromeTabs is particularly slow for just a single class).

Before all the import fixes, in full potato/concurrancy/dev mode Pyfa would take 25-30 seconds to load.

Master currently takes 14-15s to load.

This branch takes 11.5-12s to load.

Proof:
https://puu.sh/tRXMY/9f853d0a96.png

(Pyfa gui loads right after the CheckUpdate thread hits, it's the tiny orange blip right after the green line right before the 12s mark.)


Ebag333:
>Don't import a 1429 line file for a single function/defintion

There's a fair amount of time spent importing `chromeTabs`, probably because it's rather large (over 1400 lines).  We're _literally_ using either a single function and/or a single definition out of it.

The time savings is getting harder to tell as we drop, but it seems  to be good for about another .5 second speed up.

And yes, I made this import explicit: `from gui.builtinViews import fittingView, implantEditor` 

Not only is it implicit, it's buried in the middle of the imports at the top rather than at the bottom like the rest of these.  (Location doesn't really matter too much, though a the bottom is a common hack to get around circular dependencies.)

If you hate it, I'll put it back.

Ebag333:
>Break up graphFrame
>more graphframe stuff

Before you yell at me, take a look at this:
https://puu.sh/tS2N9/0476153ddf.png

_*10.5 seconds to load.*_

Also, the graphing window load seems to respond ever so slightly faster.

Old: 
https://puu.sh/tS3be/9fd0f3fb10.gif

New:
https://puu.sh/tS34b/fbaabf6cff.gif

Ebag333:
>Don't import all of eos.db just for a single query
Not including the actual code in `eos.db.__init__` (which is mostly just definitions), we were importing the following:
```
import threading
from sqlalchemy import MetaData, create_engine
from sqlalchemy.orm import sessionmaker, scoped_session
from sqlalchemy.ext.declarative import declarative_base
from sqlalchemy import pool
import migration
from eos import config
from eos.db.gamedata import *
from eos.db.saveddata import *
from eos.db.gamedata.queries import *
from eos.db.saveddata.queries import *
```

Why?  All that so we can get access to this:
```
from eos.db.saveddata.queries import getFit as db_getFit
```
(Aliased to db_getFit to prevent shadowing/confusion with `sFit.getFit`)

Ebag333:
>Make a bunch of imports explicit

Bunch more of "import all of eos.db so we can call 1-2 queries" type stuff.

>boatload of matplotlib fixes

matplotlib has been spitting errors for a while now.  Not only fixed the import and initialization errors for 2.0, but as long as we were in there fixed the legend for 2.0.
https://puu.sh/tSa3l/85704bdbe9.png

Careful viewers will notice that only the first 8 fits have unique colors.  Well, I couldn't be arsed to figure out (or look up) what the colors were past the first 8, so after that they don't get set to a color which means they go to that default blue.  And since if you have more than 8 fits on the graph it's virtually impossible to read anyway, I don't think it's much of an issue at all.

So, I think that resolves the issues Pyfa had with matplotlib 2.0.  We can upgrade to that in the skeleton, now. :)

blitzmann:
Few notes: please have one PR per issue. I don't want us to get into the habit of starting with one focus and then piling other stuff on top of it. This should fix the circular import, not do graph colors (which looks really awesome btw), change import statements outside of the scope of the focus, etc.

Also, I don't think I want MainMenuClass inside of MainFrame. It clutters it up a little too much. Taking a look at it, the only reason we import mainframe is to get a reference to the instance, which can easily be done by passing a reference to the class. i just tested this and decided to push to dev, take a look please and let me know if you see anything wonky. (dfc5627031e18eefefb8c1bfee098316e4a6fcfa)

If not, go ahead and close this and submit the graph colors to another PRs. That one should be a quick and easy merge. And please refrain from other import statement changes until we can get through the current backlog of import changes (unless they are actively proving to be a problem in __current__ builds - I'm not too concerned at the moment if pyInstaller chokes on them or if they are causing issues with other things that are currently in flux).

:)

Ebag333:
>Few notes: please have one PR per issue. I don't want us to get into the habit of starting with one focus and then piling other stuff on top of it. This should fix the circular import, not do graph colors (which looks really awesome btw), change import statements outside of the scope of the focus, etc.

First you complain about too many PRs, now you complain about too few. :P

>Also, I don't think I want MainMenuClass inside of MainFrame. It clutters it up a little too much. Taking a look at it, the only reason we import mainframe is to get a reference to the instance, which can easily be done by passing a reference to the class. i just tested this and decided to push to dev, take a look please and let me know if you see anything wonky. 

MainMenuClass uses a lot of the same imports as MainFrame.  Doing some profiling, the vast majority of our time is spend in our MainLoop(), but the second largest chunk of time (and by a significant amount over the third item in the list) is imports.

When we have an entire class (or even an entire module) that's used by a single other module, often there's not much reason to keep them apart.  And when those modules heavily share the same imports (and some of those imports are very expensive, like matplotlib), it makes a _lot_ of sense to put them together.

Regarding (most of) the other commits, these imports are fairly expensive.  `chromeTabs` is especially bad being _YUUUUGE_, when we're importing the whole file just to get 10 lines of code in a single function.  I don't even want to talk about `import eos.db` for a 3 line query.

All in all cleaning them up shaved another ~2s off my load time, so I'd say that nearly 20% speed gain is pretty significant.

I can spin the matplotlib graph fixes into it's own PR and leave this one import centric.  While it did grow a little in scope outside out just `mainFrame`, it's all still centered on the `gui` modules and the vast majority of the changes still focus on `mainFrame`.

There was another file merged into it's parents, btw.  Also due to a circular import.

I'll take a look at your commit here in a few minutes.

Ebag333:
So the solution you have looks good.

I still think that we should merge them, for one big reason.

Here are the imports from `mainMenuBar`:
```
import wx

import config
from service.character import Character
import gui.globalEvents as GE
from gui.bitmapLoader import BitmapLoader

if 'wxMac' not in wx.PlatformInfo or ('wxMac' in wx.PlatformInfo and wx.VERSION >= (3, 0)):
    from service.crest import Crest
    from service.crest import CrestModes
```

Out of the 7 imports done in `mainMenuBar`, all 7 are already done in `mainFrame`. Granted none of these  are very large imports (with the exception of `wx`, and `service.character` isn't very small either....okay so 2 our of 7 ain't bad!), but again the _only_  thing even referencing `mainMenuBar` is `mainFrame`.

TL:DR; your solution works fine, but it's still importing things multiple times which is slower. :)

blitzmann:
> First you complain about too many PRs, now you complain about too few. :P

My complaint with the PRs have always been they are too unwieldy. The quickest way to get something merged is if it's short and to the point. Now obviously with refactoring lots of files change and things need to be evaluated. But if you want something like graph colors merged, that should be in it's own PR.

>There was another file merged into it's parents, btw. Also due to a circular import.

Which file?

>  And when those modules heavily share the same imports (and some of those imports are very expensive, like matplotlib), it makes a lot of sense to put them together.

One thing about python imports is that you load them once and they are loaded for the whole program. So if mainframe and menu file import the same module, it shouldn't matter too much. I think a lot of the issues are from the circular imports  (main menu requires main frame, which imports a bunch of stuff, so it needs to finish importing those things before menu can be imported (from how I understand it)). You remove the circular import, and it should go much quicker because now you're not worried about all the imports in main frame. v0v My commit seems to be the cleanest solution.

blitzmann:
>TL:DR; your solution works fine, but it's still importing things multiple times which is slower. :) 

Well then lets just shove everything into pyfa.py. :)

Ebag333:
>Well then lets just shove everything into pyfa.py. :)

When 100% of the codebase between two files is shared by those files (as is the case between `mainFrame` and `mainMenuBar` then it makes sense to merge them.

When .0000001% of the codebase is shared between `pyfa.py` and `<insert random file here>` based on a chain of `x` functions and calls.......not so much.

blitzmann:
I can see that point. At the same time I don't like the solution of merging them, at least at the moment.

And the other file you're talking about is GraphFrame - I definitely want that to stay in it's own file. That is an entire window definition.

Ebag333:
>And the other file you're talking about is GraphFrame - I definitely want that to stay in it's own file. That is an entire window definition.

Only used by a single class. vOv

You'll need to figure out how to make it completely standalone then.  Plus it has duplicate imports of expensive import calls (matplotlib).  Merging that was a significant chunk of the 20% speed increase.

blitzmann:
> Only used by a single class. vOv

It is logically best to keep it separate since it's an entirely separate window. Target resists and Damage profile windows also only have mainframe as a parent, doesn't mean we should merge them to mainFrame.py. We need to keep a logical and physical structure that makes sense. 

>You'll need to figure out how to make it completely standalone then.

As stated, importing things multiple times isn't exactly a boon to performance since they are cached. 

I notice that this does not include the changes to the imports that I asked you to try last night, in which I moved all the matplotlib imports to make it more performant. Did you take a look? https://github.com/pyfa-org/Pyfa/blob/development/gui/graphFrame.py#L36 This should solve the expensive matplotlib import when importing GraphFrame, since it's not imported until GraphFrame is initialized (which I don't like to do in the class, but it works well). And GraphFrame isn't initialized until the user requests graphs, so it's a good alternative to having load it right from the get go.

Can we not spend multiple hours on this? This is ridiculous. Revert graph frame and main menu (after updating your dev branch ofc), branch out graph colors into new PR, and then the other import changes here seems fine and can remain until properly reviewed.  :)

Ebag333:
Closing as the circular reference was handled by @blitzmann matplotlib stuff was broken out, and am rolling this up into another PR.

PR #975, opened 2017-02-07
Git Attributes

Ebag333:
Bunch of files showing up needing line ending changes.


blitzmann:
These are the same files that I had to commit last night: https://github.com/pyfa-org/Pyfa/commit/c25a07c95b7036e164569c54546fb3e3b00716d7

I think we need to disable gitattributes until we figure out a solution...

Ebag333:
Just noticed that. Somehow my `development` branch got stuck (I think because I forked it and named the fork....`development`......oops.....)

We _could_ just set all files in the repo to the right line endings and get it over with. :)

PR #977, opened 2017-02-08
Matplotlib 2.0 support

Ebag333:
Moves import to top of file.

Adds better handling for 2.0.

@blitzmann I know you were playing around with this a bit, did you get the string issues sorted?

Also, I know you say that this takes forever to import, but at least with 2.0 this isn't an issue.  I didn't see a huge speed increase by not importing it.

With it setup this way, the graph window opens *VERY* fast.  In current build (prior to the last pre-release build) it opens painfully slow.  So I think a combination of the import fixes and 2.0 helps a lot with that.

This is _not_ ready to be merged (doesn't handle versions properly).

### DIFF
--- a/gui/graphFrame.py
+++ b/gui/graphFrame.py
@@ -53,7 +53,7 @@ class GraphFrame(wx.Frame):
 
         try:
             import matplotlib as mpl
-            if mpl.__version__[0] >= "2":
+            if int(mpl.__version__[0]) >= 2:
                 mpl.use('wxagg')
                 mplImported = True
                 from matplotlib.patches import Patch
@@ -88,7 +88,7 @@ class GraphFrame(wx.Frame):
             mpl.use('wxagg')
 
         graphFrame_enabled = True
-        if mpl.__version__[0] < "1":
+        if int(mpl.__version__[0]) < 1:
             print("pyfa: Found matplotlib version ", mpl.__version__, " - activating OVER9000 workarounds")
             print("pyfa: Recommended minimum matplotlib version is 1.0.0")
             self.legendFix = True
--- a/gui/graphFrame.py
+++ b/gui/graphFrame.py
@@ -19,7 +19,6 @@
 
 import os
 import logging
-import imp
 
 import wx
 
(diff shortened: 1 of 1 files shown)


Ebag333:
This is failing because of things *NOT* changed by this PR.

Another PR will be handling cleaning those up, so don't expect this PR to pass Travis.

Ebag333:
>Which string issues were that? I kinda gave up on working on the colors - I'm sure there's a way to make it work with matplotlib < 2, but I'm not worried about it at the moment.

Where version is a string and not a number.  So `if version > 1` doesn't work the way you would expect.  Sounds like you didn't, so I'll work on that. :)

>I believe, since matplotlib is only used in the GraphFrame, importing it inside the graph frame constructor makes most sense. 

Putting it (roughly) back to the way it was originally, and running some prints on it, I get this:
```
('graphFrame importing matplotlib:', 6.999000072479248)
('mainFrame importing graphFrame:', 7.010999917984009)
```

So, roughly speaking, it takes about 7 seconds to import matplotlib, which matches up with your eyeball.

What I _don't_ like about that, is that then when they open the graph window it'll be a very noticeable delay.

So, here's an idea.  What if we created a simple thread to try and load matplotlib, and kick it off in prefetch?  It'll be threaded, so no impact to the user.  We can wrap it in a try/except so it'll never error (and handle errors as we do now in the module).  So we do what you're suggesting (so matplotlib doesn't get loaded when `graphFrame.py` is initialized), but then there _shouldn't_ be a delay when you open the graph window.

It's a crazy idea. So crazy it just might work.

blitzmann:
> Where version is a string and not a number. So if version > 1 doesn't work the way you would expect. Sounds like you didn't, so I'll work on that. :)

Oh yeah, that's easy. `int(mpl_version) > 1`. Just cast it to the correct type.

>open the graph window it'll be a very noticeable delay.

I think the benefit of most outweigh the dissatisfaction of the few. It's a delay for a subset of users rather than a delay for all users.

> So, here's an idea. What if we created a simple thread to try and load matplotlib, and kick it off in prefetch? 

Could work. And you need to handle thread communications to let mainframe know if it was imported and, if so, to enable the graph support, etc. But sounds like a hell of a lot more work than just sticking the import back into the class (especially considering the code is already written for that). 

For now as an easy win please put it back, and we can discuss threaded imports when we start to discuss the thread handler you want.

Ebag333:
>Could work. And you need to handle thread communications to let mainframe know if it was imported and, if so, to enable the graph support, etc. But sounds like a hell of a lot more work than just sticking the import back into the class (especially considering the code is already written for that).

I think you misunderstand, I was saying to do both. :)

And no handling needed.  If it's imported, we can import it as normal in GridFrame.  If it's not imported...we import it as normal.

>we can discuss threaded imports when we start to discuss the thread handler you want

Fair enough.

PR #980, opened 2017-02-09
Code cleanup

Ebag333:
I know you hate big PRs @blitzmann and for that I am sorry.

While there is a number of things in here that you could argue doesn't _have_ to be done, cleaning up all that fluff exposed some flaws.  I did find and fix a number of bugs along the way.  And for that I'm not sorry. (Hence the branch name. :) )

Here's how the inspection report looks in `pyfa-org/Development` currently:
https://puu.sh/tVD0B/880543c79f.png

And here's the inspection report for this PR:
https://puu.sh/tVDdY/7707e0e9c2.png

Note that these do _not_ include `effects` or `scripts`.  Since there's almost 2k files in effects, and we do some things that causes inspection to be a little weird, those should be handled separately.

Essentially, here's what this PR does.

1. Fixes the remaining circular import loops that caused stacktraces.  There are still issues with circular imports (most of which are probably hidden by `import eos.db`), but now we should be down to explicit and lazy imports.
2. Adds line or function exceptions to disable inspections on code that will never pass (primarily wxPython references).
3. Fixes a bunch of obscure bugs called out by inspection. This is stuff rarely/never hit, for example double right clicking on an effect while in debug mode and not on Windows/Mac.  Or we try and access something without validating it exists first.
4. General cleanup (suprisingly small amount, thanks to previous PRs doing the bulk of that work).

In regards to the inspections left, there are a few I'd like to finish cleaning up before this gets merged.
* `Method signature does not match signature of overriden method` I think we have a combination of things.  1) Shadowing (fix) 2) pyCharm not following wxPython correctly (exception) 3) Actually throwing out the base class paramaters and doing our own...in which case...why? (fix)
* `Shadowing` Change to not shadow, simple enough.
* `Type checker` Most of these work fine because of duck typing, but we still should pass the right types in.  Also, for some we're getting `None` instead of the expected type, so we may need to create a default so pyCharm understands it (especially if we're just pulling from the database).
* `Unresolved references`  This is mostly where we access something that will exist at the point of time when we access it (probably...) but it doesn't exist when the class/function is created.  We _should_ do defaults for these...but since many we get from the database, it gets a bit tricky.
* `PEP8` There's really only 3 things in here that we're checking (or are left to check, after all the various cleanups).  First, we use some lambdas when we should have a def, should be pretty easy to convert.  Second, we have a handful of imports and classes that don't match our general naming conventions (two of the few things where we actually are fairly standardized).  Both easy fixes.

I suspect if we go through the reported issues we''ll find a number of issues that are actually legitimate bugs (if obscure).  And now that we're only looking at a couple issues per file and not dozens (or hundreds), should be easy to separate the actual bugs from not-bugs.

I will also add the inspection profile and coding style to this PR.  If we're going to actually be checking this (or rather, Travis doing most of the checking for us), then we should have something that is easily imported.  The profile should match what we use in Travis/Tox.


eos/saveddata/fit.py
```diff
@@ -1257,16 +1256,16 @@ def fits(self):
...
         return True
 
-    def __deepcopy__(self, memo):
-        copy = Fit()
+    def __deepcopy__(self, memo=None):
+        copy_ship = Fit()
         # Character and owner are not copied
-        copy.character = self.__character
-        copy.owner = self.owner
-        copy.ship = deepcopy(self.ship, memo)
```
blitzmann: Why are we deleting memos here? I'll be honest and say I'm not exactly sure what a memo is or does, but to just delete them gives me pause.

gui/itemStats.py
```diff
@@ -856,7 +856,7 @@ def OnRightClick(self, event):
         elif 'wxMac' in wx.PlatformInfo:
             os.system("open " + file_)
         else:
-            subprocess.call(["xdg-open", file_])
+            subprocess.call({"xdg-open": file_})
```
blitzmann: I'm looking at the docs and they're telling me this is supposed to be a list, not a dict

https://docs.python.org/2/library/subprocess.html

Reason for this change?

service/eveapi.py
```diff
@@ -346,19 +346,6 @@ def __getattr__(self, this):
...
-            for k, v in self.parameters.iteritems():
-                if k not in kw:
-                    kw[k] = v
-        else:
-            # no keywords provided, just update with contextual ones.
-            kw.update(self.parameters)
-
-        # now let the root context handle it further
-        return self._root(self._path, **kw)
-
```
blitzmann: This is removing code from a mostly third party module. Have we tested to make sure all API functions still function correctly? I'm not sure why we're trying to remove this bit of code

gui/utils/compat.py
```diff
@@ -1,262 +0,0 @@
-# Backport of OrderedDict() class that runs on Python 2.4, 2.5, 2.6, 2.7 and pypy.
```
blitzmann: Honestly we may be able to get rid of this crap altogether. This is a class that is used to mimic OrderedDict support for running in py2.6 and below. Since we only support 2.7 now, we might be able to nuke this.

gui/builtinViews/fittingView.py
```diff
@@ -408,6 +408,9 @@ def swapItems(self, x, y, srcIdx):
             if mod1.slot != mod2.slot:
                 return
 
+            if not getattr(mod2, "modPosition"):
+                self.mods.modPosition = mod2.modPosition = dstRow
+
```
blitzmann: This isn't making any sense to me, please explain what led up to this addition.

gui/boosterView.py
```diff
@@ -27,7 +27,8 @@
 
 
 class BoosterViewDrop(wx.PyDropTarget):
-    def __init__(self, dropFn):
+    def __init__(self, dropFn, *args, **kwargs):
+        super(BoosterViewDrop, self).__init__(*args, **kwargs)
```
Ebag333: As I understand it, and admittedly this is an area in Python where I'm still weak, when have classes that inherit the `super` call simply makes it explicit.  Yes it's inheriting it, no it's not explicitly doing so, and `explicit is better than implicit`.

The way we're using `super` is the standard basic call, you don't need to  pass `self` because  that's handled for you with the `super(BoosterViewDrop, self)`.

The `__init__` change for the line above the `super` is just to make sure that we match what the parent class is.

If I'm misunderstanding what you're getting at let me know.

eos/saveddata/fit.py
Ebag333: I agree and `deepcopy` is weird.  If you google it (which I did extensively) you find people complaining about all sorts of weird issues trying to use memos (including several versions of Python where they're broken entirely).

pyCharm clearly doesn't recognize memo as a valid parameter  (even though if you go to the function it's calling, it clearly has `Memo=None` as a parameter).

Ultimately though, we're not passing in a memo when using deepcopy.  I also validated that the object before/after the change looked the same, it copies fine.  vOv

gui/itemStats.py
Ebag333: Yeah, I think pyCharm is wrong here.

Well, not exactly wrong.  It was complaining because  there were situations where we could pass in a `None`, and because `subprocess.call` supports `**kvargs` it throws it off a bit.  So I suspect that we could actually do it either way.

I went ahead and reverted it though.

service/eveapi.py
Ebag333: Yeah, looking it over again I don't think we actually should.  It's a little weird the way they're making it callable, but I reverted it.

This is one reason why I don't like including third party modules as direct parts of Pyfa, it exposes all the things they do that are weird/wrong. :)

gui/utils/compat.py
Ebag333: There's a few bits of code that are geared for 2.6 and older (we actually try and import some 2.6 stuff and only except back to the 2.7 version...which is...weird....).  If you're okay with nuking it, I'm okay with nuking it. :)

gui/boosterView.py
blitzmann: > The __init__ change for the line above the super is just to make sure that we match what the parent class is.

We don't pass anything to the parent class. If we ever need to we can add that. But having something in code that is not in use is somewhat confusing. I would remove that - looking at the definition of the class it's not immediately clear why we support that (and I'm not even sure if `wx.PyDropTarget` even has parameters that it uses - Im having trouble finding documentation online). Having the `super()` is find though, thanks for the explanation.

And why did you keep the original `wx.PyDropTarget.__init__(self)` call after using the `super()`?

gui/builtinViews/fittingView.py
Ebag333: Long story short there are times when we can have `modPosition` be empty.  This is just a check to see if it is, and if so we explicitly set it.  Looking at it, though, I think there's a better way of handling it.

It's a good example of what we have scattered throughout the code base, which is a larger issue.

```
if clone and mod2.isEmpty:
    sFit.cloneModule(self.mainFrame.getActiveFit(), srcIdx, mod2.modPosition)
```

We actually commit the same sin twice here, though one is worse than the other.  If `isEmpty` didn't exist, then we'd skip it, so that's not huge problem (though there's lots of times when we are checking for a value in the `if` and that would stacktrace).

The one that will stacktrace is `mod2.modPosition`.

We do this a _lot_ in the code base, we simply use variables assuming that they exist without any validation or defaults.

I'm going to rewrite this one to use `getattr` directly, I don't know why I didn't go that route in the first place.

Honestly these really should be rewritten to use `getattr` and a default value, but since that's littered _everywhere_ in the code I don't think we should tackle that directly.  Mostly just make a habit of _not_ accessing attributes directly, but using `getattr` and a default.

eos/saveddata/fit.py
blitzmann: Fair enough. Looking it up it seems to keep a ledger of the original objects that can be accessed outside of the deepcopy. We don't use it, so this is probably fine.

service/eveapi.py
blitzmann: > It's a little weird the way they're making it callable, but I reverted it.

How so? Checking on it, it's a method that is definitely used. I believe it's used so that you can make a call to `api.characters` and it automatically knows where to go without having to code in support for the Characters api. What was pyCharm complaining about?

gui/utils/compat.py
blitzmann: I mean, pyfa doesn't run correctly in 2.6 anyway. 2.7 is nearly 7 years old now. Go for it.

gui/builtinViews/fittingView.py
blitzmann: @Ebag333 missed this comment?

blitzmann: > Long story short there are times when we can have modPosition be empty. This is just a check to see if it is

When will `modPosition` be empty? If we're in the `swapItems` methods, then that means we have two modules to swap and they will both have this property simply due to the fact that they are active modules. It's okay to assume we return something here, unless you noted it actually breaking during normal use. In which case, if they don't have it, I'd like for it to break so we know about it rather than trying to deduce which position the module is supposed to be in.

Furthermore, we shouldn't be setting `modPosition`. It doesn't have a setter and shouldn't have one - this is supposed to be a read-only property that returns the modules index in the fit's module list. I think this is being done all wrong.

Additionally, I'm curious on two things:
* Why `getattr(mod2, "modPosition")` instead of simply `mod2.modPosition`?
* Why are you setting `self.mods.modPosition`? `self.mods` is a list of modules...

Ebag333:
>SQLAlchemy handles it all. The properties are defined via DB schema, thus pyCharm probably doesn't know about them. They are always there though by virtue of ORM.
>As you saw when you start defining them things cvan get weird. I ran into this problem a long time ago when I first started.

For most of them, adding a default seem fine.  I'd still like to tackle this, because then:
1) Inspection and various tools can actually follow the chain.
2) We would have it explicitly defined that these values come from SQLAlchemy (granted through a comment above the definitions, but then future devs would know).

I suspect that something along the lines of:
```
if not ID:
     ID = None
````
would work, but given the complexity of it, and the relatively low impact, and how pervasive this is,  I don't really want to tackle it in this PR.

File: gui/builtinViews/fittingView.py (hunk 5 above)
Comment (Ebag333) [inline gui/builtinViews/fittingView.py]: >When will modPosition be empty? If we're in the swapItems methods, then that means we have two modules to swap and they will both have this property simply due to the fact that they are active modules. It's okay to assume we return something here, unless you noted it actually breaking during normal use. In which case, if they don't have it, I'd like for it to break so we know about it rather than trying to deduce which position the module is supposed to be in.

It's more of a "this shouldn't happen but it could" scenario.

Honestly, I can't remember if I actually ran into this breaking.

>It's okay to assume we return something here

Defensive programming all the way! :)

>Furthermore, we shouldn't be setting modPosition. It doesn't have a setter and shouldn't have one - this is supposed to be a read-only property that returns the modules index in the fit's module list. I think this is being done all wrong.
>Additionally, I'm curious on two things:
>Why are you setting self.mods.modPosition? self.mods is a list of modules...

Dammit, I can't hide stupid code on git. :(

>Why getattr(mod2, "modPosition") instead of simply mod2.modPosition?

Now this one is legitimate and half right.  Because in Python `None` = `False` = `0`.

Anyway, it's that last one that will get you, since often lists and positions start with 0.  Rather than assuming that `modPosition` can never have a position of 0, we use `getattr` which will return nothing (since no default is set), and not run into problems if it's legitimately set to 0.

Of course in this case if `modPosition` were actually 0....ummm....so yeah, I ended doing the same thing by trying to avoid it.

We can avoid it entirely by using `is not None` and `is not False` because `is` is explicit.  But I think what I changed it to avoids this entirely.

File: gui/boosterView.py (hunk 6 above)
Comment (blitzmann) [inline gui/boosterView.py]: > If we're not using what the parent sets, then often it's a really good indication that we shouldn't be using that as a parent.

This, in general, is not true. If we inherit something and the parent __requires__ it as part of it's `__init__`, then absolutely the child class should also either have it in it's own `__init__` definition or call the parent class with some known initial values. There's no other way around it, considering it's required. But if it's an optional parameter, then it's unnecessary and, in my view, clutters up the code. `*args, **kwargs` is by definition an optional parameter.

Take all of our `wxWindow` definitions. We inherit `wx.Panel`, `wx.Frame`, etc, but we don't define the same things the parent defines "just in case we need them in the future". All we do (usually) is define the parent window, and then call the parent with the required parameters (along with our implementation of optional parameters depending on what we need for this Window). ie:

```
class FighterView(wx.Panel):
    def __init__(self, parent):  # note lack of parameters
        wx.Panel.__init__(self, parent, id=wx.ID_ANY, pos=wx.DefaultPosition, style=wx.TAB_TRAVERSAL)
```

I would rather keep things explicit (as you keep advocating for) - if our class implementation doesn't require us to pass in arguments to the parent, then don't. Otherwise we can have situations where I or some other contributor is trying to figure out why we have our implementation defined with  `*args, **kwargs` but we aren't using them anywhere. And I definitely don't want us to get into the habit of littering our wx definitions with `*args, **kwargs`.

Please change this and the extra call to the parent, and I believe we should be good to go with this PR. :)

Ebag333: >All we do (usually) is define the parent window, and then call the parent with the required parameters (along with our implementation of optional parameters depending on what we need for this Window). 

And in this case, I would agree with the argument made here:
http://stackoverflow.com/questions/2294200/issue-with-python-class-hierarchy/2295784#2295784

If we're not doing what the parent is doing, then why are we inheriting the parent?  If we need to wrap it to do something else, we can do that then call the parent.

If we are doing what the parent is doing....then why do we have different parameters?

>Take all of our wxWindow definitions. We inherit wx.Panel, wx.Frame, etc, but we don't define the same things the parent defines "just in case we need them in the future". 

Especially in the case where we're extending (and not replacing) a default class/function, we can get ourselves into trouble.  There's a rather good explanation here:
http://stackoverflow.com/questions/2398666/python-can-subclasses-overload-inherited-methods

Ultimately,  I think our future selves will thank us for this.  It all boils down to two main arguments.

Because inspection tells us to.  Now, inspection isn't perfect (it has a hard time following `wx` for example) but it's pretty damn good.  We currently have 49 warnings where the signatures don't match.  If we don't clean those up, then we have a few  choices.
1) Exception out the check entirely.  This I don't like, because it's definitely a good smell test that something might not be right.
2) Exception out individual classes/methods.  This is better than option 1,  but then the temptation becomes "just add the exception  to everything".
3) Leave it as is.  Then we just start ignoring the warnings, and might as well have done option 1.

The second argument is a little simpler.  Because  our future selves (and future devs) won't go hunting for monkeypatches when they are looking to implement something.  When I'm looking at docs, I go back to the Python/wx/whatever docs.  If something then doesn't behave the way I expect it to, I will first  assume it's my code or lack of understanding, not a monkeypatch that causes it to not be handled it the way the docs say it will.  That means I'm wasting time/effort and is _incredibly_ frustrating.

Yes, this means that we'll have definitions with `*args` and `**kwargs`, but there's no way around that unless you  want to entirely just throw out that check or handle the reference back to it differently (explicitly call it instead of inheriting it).

And as far as "littering" goes, we're really only talking about about a handful of lines here,  and primarily will just be in the `__init__`.  This isn't a change of 2000 lines, probably more like 40-50.

There  _are_  a few  functions that we might need wider changes. `populatePanel` is pretty inconsistent, for example, and for those we should probably tackle them outside this PR, since that might entail wider code changes.  But for just explicitly inheriting base classes, that's not a big deal. 

>I believe we should be good to go with this PR. :)

I'm going to go through another pass of the warnings and make sure there's nothing  else we missed.  Inspection caught a missing import in `fittingView`, so I don't think we're _quite_ there yet (but damn close).

gui/utils/compat.py
Ebag333: Mmmm, I'm going to leave this one for now.  Taking a quick look at it, `OrderedDict` is fairly intertwined, and I don't want to make this PR any more complex.

I did nuke the import of the 2.6 `urllib` though. :)

Ebag333:
@blitzmann I think that we've managed to hit everything. I'm not finding any more missed imports or references, or if they're still there they are not being flagged.

One last change I do want to make is the `method signature` one, but only for ones where it's a simple add of base parameters (no refactoring needed).

If we are _not_  going to do that, then I will disable that check entirely.

gui/boosterView.py
blitzmann: > If we're not doing what the parent is doing, then why are we inheriting the parent?

We are doing what the parent does. That's the point of the `super`. The parent will initialize and do whatever it needs to do, but doesn't require any parameters to work. We inherit from it because that is what we need to do to get drag and drop support. This is basic inheritance. 

> If we are doing what the parent is doing....then why do we have different parameters?

Because __our__ implementation of it does not require additional parameters. 

I'm contacting you on Slack to find out more information on this. If we do go ahead with this, I would prefer to only have `**kwargs` instead of that and `*args`. From what I read on the interwebs it's the preferred method.

PR #986, opened 2017-02-11
Logbook V2

Ebag333:
Rolling up all changes from #962 as the base has been massively changed by code cleanups.  

Few minor differences between this version and the previous one, primarily rename our logging instance to a universal `pyfalog` to avoid shadowing either Logging or Logbook (turns out `logging` and `logger` are both used, go figure).

There are a few things left todo.

The timer is disabled currently, as we were passing the logger instance into it. Just need to figure out how to pass that cleanly (or perhaps we don't even need to do that).

The other thing I'd like to tackle is to see about getting logging output back out from our threads (we don't currently).  We *might* be able to simply pass the logbook handler instance in.  Alternative is more complex, handle writing output back out.

Other than those two things, this PR should be pretty much done.


Ebag333:
>The other thing I'd like to tackle is to see about getting logging output back out from our threads (we don't currently). We might be able to simply pass the logbook handler instance in. Alternative is more complex, handle writing output back out.

Probably should just handle this as part of a larger threading refactoring.

Ebag333:
@blitzmann I think this is ready to merge now.

You may want to add several different logging level configurations in pycharm.  The two I use primarily are:
`--debug --logginglevel Debug`
and
`--debug --logginglevel Info`

Debug lets me see everything. Info filters out most of the noise (if you're just looking for stacktraces and such).

blitzmann:
I took a look at the skels last night, and I think jacking this dependency in would would work as a stop gap. Gonna try to test this theory tonight. If it works well, I'll try to prioritize this PR as the next one :)

blitzmann:
@Ebag333 I've confirmed that simply moving the `logbook` package from site packages to the respective areas in the skeletons actually works, so this should be googd to go. Some outstanding questions:

* You mentioned a timer, what do you mean?
* Is there a flag or a way currently to dump output to a file? I know we';re currently logging to memory and dumping on error, but is there a way to just dump to file by default? (I want to use this to better understand the unicode path issues we're having)

thanks!

blitzmann:
Think you already answered my question from the other PR:

> When in debug mode, we act as we do now, write to a log (though we write to Pyfa_debug.log) and output to the console.

Haven't actually tersted this yet, but will do so

Ebag333:
>You mentioned a timer, what do you mean?

Something I disabled because it tied into logging. Don't worry, I fixed it. :) 

>Is there a flag or a way currently to dump output to a file? I know we';re currently logging to memory and dumping on error, but is there a way to just dump to file by default? (I want to use this to better understand the unicode path issues we're having)

I think that I set it up so if it's flagged as frozen and debug mode is on, it'll throw a critical "error". It's just to trip it so that it will write everything to the log.

Won't really help us for that problem though. We are running into the problem getting the path, so if we don't have a path how can we write a log there? 

As much as I want this PR merged, it's not going to help I don't think....

blitzmann:
> Won't really help us for that problem though. We are running into the problem getting the path, so if we don't have a path how can we write a log there?

I'm aware, I was hoping to use this more for my dev builds as I look into the issue, can hardcode to a known good path in that case

blitzmann:
Gonna try to start setting milestones on things so that we know what the focus is. :D

blitzmann:
What I've tested seems to check out

PR #992, opened 2017-02-15
Merging Drone Stacks

### LINKED ISSUE #728: Bug: Stacking active drones on inactive results in mixed stack
If you grab a stack of inactive drones (in-bay, red x) and drag them on top of a stack of active drones (in-space, green check) of the same type, the resulting stack is inactive: this is ok, if not what I would have expected. However, if you grab a stack of active drones and drag them on top of a stack of inactive drones, the result is a stack which shows as active, but actually contains (as evidenced by the drone damage display) the mix of active and inactive drones. In this state, clicking to deactivate the stack actually activates the inactive drones and leaves the stack marked active.

Reproduce by:
1. Open a Federation Navy Comet.
2. Put a Hobgoblin I (we'll call it HA) in the drone bay.
3. Mark the drone active. Note the drone DPS.
4. Put another Hobgoblin I (we'll call it HB) in the drone bay, by double-clicking to create a new stack.
5. Drag HA atop HB. Note that the stack is marked active, but drone DPS doesn't change.
6. Click the HA+HB stack. Note that the drone DPS doubles, and the stack remains active.

Ebag333:
Check if we have active drones when merging, if we do make them all active to avoid getting weird scenarios (like deactivating stacks, or only activating parts of stacks).  Fixes #728

### DIFF
--- a/gui/droneView.py
+++ b/gui/droneView.py
@@ -148,11 +148,6 @@ class DroneView(Display):
         sFit = Fit.getInstance()
         fitID = self.mainFrame.getActiveFit()
 
-        # If either stack has active drones, make them all active. Fixes #728
-        if (getattr(self.drones[src], "amountActive", 0) + getattr(self.drones[dst], "amountActive", 0)) > 0:
-            self.drones[src].amountActive = getattr(self.drones[src], "amount", 0)
-            self.drones[dst].amountActive = getattr(self.drones[dst], "amount", 0)
-
         if sFit.mergeDrones(fitID, self.drones[src], self.drones[dst]):
             wx.PostEvent(self.mainFrame, GE.FitChanged(fitID=fitID))
 
--- a/gui/droneView.py
+++ b/gui/droneView.py
@@ -147,6 +147,12 @@ class DroneView(Display):
     def _merge(self, src, dst):
         sFit = Fit.getInstance()
         fitID = self.mainFrame.getActiveFit()
+
+        # If either stack has active drones, make them all active. Fixes #728
+        if (getattr(self.drones[src], "amountActive", 0) + getattr(self.drones[dst], "amountActive", 0)) > 0:
+            self.drones[src].amountActive = getattr(self.drones[src], "amount", 0)
+            self.drones[dst].amountActive = getattr(self.drones[dst], "amount", 0)
+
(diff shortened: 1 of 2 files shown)


Ebag333:
@blitzmann this is ready to merge.  This is failing Travis because of other build issues outside of the scope of this.

blitzmann:
I would like to have this take the source stack and apply its 'state' to the resulting stack (so if you drag an inactive to active, result is inactive. Active to inactive = active). We would have to check if `d1.amountActive > 1` I believe.

Additionally, I'll write a migration script to fix this in the database. We currently don't have any need to have `amount != amountActive`

Also, whats up with all the `getattr()` litter? `amount` and `amountActive` will always be available (doubly so since they're NOT NULL in the database, yay for foresight). We can't get into the habit of always checking to make sure these things exist, otherwise that would incorporate half of the code :P

Ebag333:
>I would like to have this take the source stack and apply its 'state' to the resulting stack (so if you drag an inactive to active, result is inactive.

I don't think that's intuitive.  The OP even  stated as much:

>If you grab a stack of inactive drones (in-bay, red x) and drag them on top of a stack of active drones (in-space, green check) of the same type, the resulting stack is inactive: this is ok, if not what I would have expected.

So I think it's more intuitive that if you're merging two stacks, and one is active, the result is active.

>Also, whats up with all the getattr() litter? amount and amountActive will always be available (doubly so since they're NOT NULL in the database, yay for foresight). We can't get into the habit of always checking to make sure these things exist, otherwise that would incorporate half of the code :P

I don't consider it litter (yay for defensive programming!) but you are right, we don't need it twice (just the first time we set the attribute).  And "half of the code" is a bit of an exaggeration, since we're only adding a handful more characters that we'd have without it. :)

blitzmann:
> the resulting stack is inactive: this is ok, if not what I would have expected.

>  this is ok, if not what I would have expected

ergo, it's expected.

But, 

>one is active, the result is active

I missed this part. So if at least one is active, it's active, otherwise it's inactive? I can get behind that.

> we don't need it twice

We don't need it the first time. This is a trend that I see in a lot of your recent code, and it's wholly unnecessary and very clean or readable code (I know, pyfa doesn't really fit that bill so well). There literally is no reason to check if it exists, as it always will. Please remove. <3

PR #993, opened 2017-02-15
Settings: Bug fixes and better default value handling

Ebag333:
In current branches (Dev/Master), there's a bug that requires us to have default values. But sometimes you don't want to have to have defaults for settings, it might not make sense to set them, or they might be variable, or the defaults might be set elsewhere (think column widths).

So, lets not force you to have defaults. :)

Also, we don't need to try and inject defaults 3 times, that's silly. Lets make that a little simpler.

Additionally, if you added a default down the road (after the settings file was created), it wouldn't get added to the settings.  So lets go ahead and allow for that scenario.

@blitzmann this should be ready to merge.

### DIFF
--- a/service/settings.py
+++ b/service/settings.py
@@ -66,6 +66,8 @@ class SettingsProvider(object):
             if item not in s.info:
                 s.info[item] = defaults[item]
 
+        self.settings[area] = s
+                
         return s
 
     def saveAll(self):
--- a/service/settings.py
+++ b/service/settings.py
@@ -50,12 +50,15 @@ class SettingsProvider(object):
             p = config.parsePath(self.BASE_PATH, area)
 
             if os.path.exists(p):
+                # noinspection PyBroadException
                 try:
                     f = open(p, "rb")
                     info = cPickle.load(f)
                 except:
                     info = {}
                     # TODO: Add logging message that we failed to open the file
+            else:
+                info = {}
 
             s = Settings(p, info)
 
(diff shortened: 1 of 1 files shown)


blitzmann:
> there's a bug that requires us to have default values. 

What bug this is?

Maybe I'm not understanding, but settings should require some sort of default (otherwise whatever uses that setting doesn't know what to use, ofc). And the code in this PR doesn't really touch defaults from what I can see (but I see what you mean about the redundant for loops, nice catch)

Ebag333:
>Maybe I'm not understanding, but settings should require some sort of default (otherwise whatever uses that setting doesn't know what to use, ofc). And the code in this PR doesn't really touch defaults from what I can see (but I see what you mean about the redundant for loops, nice catch)

Currently if you try and do settings, but do not give defaults, even if you set it when the code is running it won't stick properly.  You have mentioned this before, that settings must have a default to work properly (or maybe I'm reading more into what you mentioned off hand than was there).

Anyway, this is silly. You shouldn't *HAVE* to have a default, since lots of things we might want settings for get their defaults from various locations (database, wx, etc), and might as well let those be handled unless the user explicitly changes it.

Setting code responsible for that behavior is below:
```
def getSettings(self, area, defaults=None):

    s = self.settings.get(area)
    if s is None:
        p = config.parsePath(self.BASE_PATH, area)

        if not os.path.exists(p):
            # SNIPPED FOR LENGTH
        else:
            try:
                f = open(p, "rb")
                info = cPickle.load(f)
                for item in defaults:
                    if item not in info:
                        info[item] = defaults[item]

            except:
                info = {}
                if defaults:
                    for item in defaults:
                        info[item] = defaults[item]

        self.settings[area] = s = Settings(p, info)

    return s
```

The parts to pay attention to are:
`defaults=None` (from the function definition)
and
`for item in defaults:` (under the `else` inside the `try`)

This is why if you pass in defaults it works, because it has something to iterate through.  When you _don't_, we define Default as `None`....which is not iterable.  So we stacktrace on `for item in defaults`.

But wait, we have a valid list of items in our `info{}` that we loaded from the file.  We just failed to inject any missing defaults.  Well, we get bumped down to the `except` and the first line there is `info = {}` which completely nukes all the settings we loaded from the file.  Here we actually check if we have defaults, which we don't, so we skip it.

So basically the steps are:
1) Load settings from file (works fine)
2) Nuke all settings
3) Create an empty setting object

This _really_ showcases two reasons why I'm pushing hard for this cleanup.

1) We have an exception with no logging, which means our stacktrace was swallowed up.
2) pyCharm inspection throws a warning about `defaults`, because it knows that `None` is not iterable.

While we don't have any settings currently that would benefit from this, there are a number of settings that I'd like to explore that _would_, such as remembering user set column widths.

blitzmann:
So it boils down to the fact that we currently must explicitly define a setting in order for it to be a thing, whereas you want to be able to dynamically define a setting...?

Ebag333:
>So it boils down to the fact that we currently must explicitly define a setting in order for it to be a thing, whereas you want to be able to dynamically define a setting...?

No?

It boils down to we stacktrace, and then hide the stacktrace instead of exposing it or properly handling it, and then throw away the things we did correctly and replace them with an empty hollow shell.

Sorry, was the explanation above too long? :D

Here's the TL;DR:
The way that I have it now is the way it was _supposed_ to be originally.  Yes this means that settings can be defined dynamically and they do not have to be a hard 1:1, nor do they need a default.

blitzmann:
meh, if you say so. I'll take a peak and merge into dev.

Issue #235, opened 2014-12-27
Add Shield/Armor/Capacitor transfer per second for Logi ships


Iiridayn:
I'd like to see what my shield, armor, and capacitor transfer per second are when designing logistics ships. This might work under firepower as another option to the button that changes the display to mining amounts. I don't particularly care about structure transfer per second.

lunedis:
I'm currently trying to implement this by building a 3rd StatsView besides Firepower and Mining.

I copied the MiningYieldViewFull and made a new class called OutgoingViewFull, changed some stuff there (which should be syntactically correct) and also called OutgoingViewFull.register() at the end.
I inserted "outgoingViewFull" in the __init__.py.

Pyfa still doesn't find the View if trying to load it by changing the switch button in FirepowerView or making it a default view. As far as I understand pyfa the following line in StatsView should load all views?
```python
from gui.builtinStatsViews import *
```

Should I commit my code so you can see my commit or do you have any ideas what my mistake might be?

blitzmann:
Yeah, go ahead and commit the code to you repo on another branch off master. I can take a look at it. IIRC the switching between panels was pretty hackish. The contributor that did it did a great job, but the panel system just wasn't designed with the idea of turning panels on and off. It's probably something to do with that.

lunedis:
![image](https://cloud.githubusercontent.com/assets/9085102/8557142/acbc25c4-24fb-11e5-964d-4af11990942e.png)

I did something.

https://github.com/lunedis/Pyfa/commit/d033e4a8d919664d3ae210098291e38df71d3256

droppedoncaprica:
Nice dude!

blitzmann:
I will take a look at it as I can. We just release v1.13.0, so we until the next release to polish this up if needed. =)

blitzmann:
Sorry about the delay, but I was working on some other stuff. Took a look at your patch, and overall I like it. There is an issue tho:

![capture](https://cloud.githubusercontent.com/assets/3904767/8890693/0a390a48-32d8-11e5-9c6f-fa2004ec21ab.PNG)

Too much text will overflow and obscure the button, which cannot happen as there is no way to make the stats area wider, and the user may get stuck on a display if they have large enough text there. There seems to be a space between the icon and the text - not sure if that's just how it's laid out, or if there is an issue with the gui grid.

[...]

PR #994, opened 2017-02-15
Stats Pane - Remote Reps

Ebag333:
Going through some old issues and turned up #235

It took a bit of work, and unfortunately I didn't see a PR or branch I could easily work off (largely in part due to the major changes over the last few months), but I took what @lunedis had and tweaked it a bit to make it fit.

Adds a very small (so hardly any space lost @blitmann) pane for remote shield, armor, hull, and capacitor.
https://puu.sh/u4So7/8d210d9e56.png

Should be fairly difficult to overflow the numbers since I rearranged it to free up some room.

This would be _extremely_ useful for Logi pilots.

@blitmann I need to add some tooltips, then this will be ready to merge.

### DIFF
--- a/eos/saveddata/fit.py
+++ b/eos/saveddata/fit.py
@@ -18,6 +18,7 @@
 # ===============================================================================
 
 import time
+import math
 from copy import deepcopy
 from itertools import chain
 from math import sqrt, log, asinh
@@ -127,6 +128,12 @@ class Fit(object):
         self.__capUsed = None
         self.__capRecharge = None
         self.__calculatedTargets = []
+        self.__remoteReps = {
+            "Armor": 0,
+            "Shield": 0,
+            "Hull": 0,
+            "Capacitor": 0,
+        }
         self.factorReload = False
         self.boostsFits = set()
         self.gangBoosts = None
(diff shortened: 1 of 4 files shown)


blitzmann:
As we've talked about on a number of occasions, I'm hesitant to increase the amount of panes. Not to mention that, for someone who flies DPS, this is just wasted space.

The thing with the firepower / mining / RR panes is that a ship will only ever really do one of those things at a time (with some cross contamination ofc). I think the original implementation would have worked great if the overflow was fixed. 

That being said... I think I would be okay with adding this panel as it's information not currently available, and I agree logis would benefit a great deal.

But, the code...

Yours seems way more complex that it should be in comparison to https://github.com/lunedis/Pyfa/commit/d033e4a8d919664d3ae210098291e38df71d3256#diff-12319c39afbe21ce25350ff3158c7d24R938. Why all the complexity?

It also breaks a lot of current conventions:
1) Doesn't use `clear()` to clear the values, which is what we use when a fit is recalculated. Instead it just clears them in the function. 
2) Doesn't use properties like all other "calculated" things. This would help if we were ever needing just the shield HP, for example (instead of running the calculate code again, which is one of the reasons why the properties are set up the way they are to run the code if it currently doesn't exist)
3) `module_group in (41, 1697)` Please use the actual group name. I'm aware there is the chance that CCP changes things and it breaks, but it is fairly rare and using the names is so much more readable.
4) `getattr(module.item, "groupID", 0)` As with the other PR I just commented on... there is no reason to use these. If an item doesn't have a group ID, then there's more fucked up with it than a stat calculation. And this is
5) `module.getModifiedItemAttr("reloadTime", 0)` oh god more defaults... going overboard with defaults. :)

I like his code much better - it's simple, readable, and understandable. But I like the way your panel looks a lot more. However, you forgot to include it in the commit... <_<

What I would suggest: cherry picking @lunedis commit and tweaking it to get it working with current pyfa if needed, and then using your panel.

Ebag333:
>As we've talked about on a number of occasions, I'm hesitant to increase the amount of panes. Not to mention that, for someone who flies DPS, this is just wasted space.
>The thing with the firepower / mining / RR panes is that a ship will only ever really do one of those things at a time (with some cross contamination ofc). I think the original implementation would have worked great if the overflow was fixed.

Don't tell my Porpoise that, since it does reps, DPS, bursts, _and_ mines all at the same time. :)

Also, once (if?) we merge the preference pane PR, we will have more control over it.

>Yours seems way more complex that it should be in comparison to lunedis@d033e4a#diff-12319c39afbe21ce25350ff3158c7d24R938. Why all the complexity?

Mostly to properly handle ancil remote reppers, which weren't a thing when that was initially created. And without all the hard coding that's found elsewhere.

>Doesn't use clear() to clear the values, which is what we use when a fit is recalculated. Instead it just clears them in the function.

Can change that.

>Doesn't use properties like all other "calculated" things. This would help if we were ever needing just the shield HP, for example (instead of running the calculate code again, which is one of the reasons why the properties are set up the way they are to run the code if it currently doesn't exist)

Not sure what you mean by this.  It's a property, so it'll be there.  And because  we define a default, you can always use `self.remoteReps['Shield']`.

There a re a number of other similar properties, such as effective HP, etc that  function in the same way.  So.....

>module_group in (41, 1697) Please use the actual group name. I'm aware there is the chance that CCP changes things and it breaks, but it is fairly rare and using the names is so much more readable.

Mmkay.

>getattr(module.item, "groupID", 0) As with the other PR I just commented on... there is no reason to use these. If an item doesn't have a group ID, then there's more fucked up with it than a stat calculation. 

It also means the difference between stacktracing and just skipping the broken module.  Considering that we've had issues with database corruption (though granted, on the saveddata side rather than the EVE db side)  and CCP has released corrupted data in the past, it's not outside the realm of possibility.

blitzmann:
>Not sure what you mean by this. It's a property, so it'll be there.

Sorry, wasn't clear. Yes, it a property, but it's a property that calculates this stuff every time it's accessed, regardless if the fit was recalculated or not. This is why a lot of the the other properties look like this:

```
    @property
     def neut(self):
         if self.__neut is None:
             self.calculateOutgoingStats()
 
         return self.__neut
```

So that as soon as one of these four "attributes" is accessed, it's calculated, but from there on out it's not calculated, it uses what was calculated before. Doesn't matter too much if we're only calling your function once in the entirety of pyfa (stats pane) but this is against current convention and, like I said, if we want to use these properties else where, it will be calculated every time it's called with the current implementation. 

> It also means the difference between stacktracing and just skipping the broken module. Considering that we've had issues with database corruption (though granted, on the saveddata side rather than the EVE db side) and CCP has released corrupted data in the past, it's not outside the realm of possibility.

I'm not really interested in testing for every single edge case. A meteor shouldn't strike tomorrow, but I'm not going to be defensive and burrow underground today because of the possibility. If CCP breaks shit, then we adapt. 

In the years that I've worked on pyfa, surprisingly very little has changed in the structure of CCPs data (at least for existing mechanics - new mechanics such as fighters and command boosts obv require some research to implement). If existing things do change, we usually know about it via the update / diff procedure, and can react.

The only reason, in my view, to have `getattr()` or `getModifiedItemAttr('attr', 0)` is if we're dealing with a variety of unknown modules or types that may or may not have these attributes. But in this case, we're explicitly saying "if this is a remote shield rep, do this" and we can assume the remote shield rep will have a certain set of attributes. If it doesn't, then a lot more needs to change in the code _anyway_ to support that change (other properties, misc column, effects, etc), so there's no point in safeguarding these.

>Mmkay.

<3 My eyes and brain thank you

> Mostly to properly handle ancil remote reppers, which weren't a thing when that was initially created.

We should add ancil repper support, but at the same time it's more complex than it has to be. Case in point: reload stuff is already handled by the `cycleTime` property, but you seem to be reinventing what we already have. Which is why I suggested taking your panel code and @lunedis implementation code and merging them. It would take very little to bring his code up to date. If you want, you could just include the panel here and I can bring his commit it and update it, shouldn't be too difficult.

Ebag333:
>Case in point: reload stuff is already handled by the cycleTime property, but you seem to be reinventing what we already have. 

Please, point to me where we already have this. Because we do it several ways, and most of them are wrong. :)

(We hardcode various values in effects, or used to, I forget if that got fixed or not. We hardcode the reload time for capsim and tank calcs.  Etc)

Anyway, I'll close this and open a new one with the cherry picked commit.

blitzmann:
> Please, point to me where we already have this. Because we do it several ways, and most of them are wrong. :)

> (We hardcode various values in effects, or used to, I forget if that got fixed or not. We hardcode the reload time for capsim and tank calcs. Etc)

As stated, we have a `cycleTime` property for many of our item wrappers. If we're using different conventions elsewhere, we should probably fix that. However, in the case of effects, I'm sure there are certain effects that don't utilize it because we either overlooked this property or we require a slightly different calculation for whatever reason. I can't comment on that without looking at specific examples. All I know is the `Module.cycleTime` incorporates a lot of other properties, such as factoring in reload time, module reactivation delay, etc.

I haven't looked at the other PR yet, will let you know. Thanks! :)

PR #995, opened 2017-02-16
Remote Repair Pane v2

Ebag333:
Finally (maybe) implements #235

See #994 for some history.


Ebag333:
@blitzmann got the tooltips added.

eos/saveddata/fit.py
```diff
@@ -392,6 +399,9 @@ def clear(self, projected=False):
         self.ecmProjectedStr = 1
         self.commandBonuses = {}
 
+        for remoterep_type in ("Armor", "Shield", "Hull", "Capacitor"):
+            self.__remoteReps[remoterep_type] = 0
+
```
blitzmann:
```
for remoterep_type in self.__remoteReps:
    self.__remoteReps[remoterep_type] = 0
```

blitzmann:
For the property thing, I can give you a better example of why I request it specifically once I get home. :)

eos/saveddata/fit.py
blitzmann: This is not that big of a deal since it's all hardcoded, but didn't know if you knew about this syntax (and if you use it we wouldn't have to update it here if the original ever changes)

Ebag333:
>Still not using this construct which I requested (taken from cherry picked commit):

Well no, we are not. And since we're not doing neuts here at all, and neuts are handled entirely differently (we already have a construct for them, btw), and since neuts are (hopefully, eventually) going to be completely redone with the capsim fix, I didn't see a reason for including a property we wouldn't be using anyway.

>When I asked you to cherry pick it, I didn't mean cherry pick it only so that the other contribute gets credit. I literally meant to use his code and fix it up to work with what we now have (ancil repairers, etc). It's much cleaner. =/

I disagree (especially on the GUI bit).

I'll look this over. Again.

blitzmann:
Please see my commit as to what I was talking about with respect of the properties. The previous way calculated it during every access of the property:

```
Python 2.7.10 (default, May 23 2015, 09:40:32) [MSC v.1500 32 bit (Intel)] on win32
In[2]: import eos
In[3]: import eos.db
In[4]: fit = eos.db.getFit(330)
In[5]: fit
Out[5]: Fit(ID=330, ship=Vexor, name=Vexor fit) at 0x4549b90
In[6]: fit = eos.db.getFit(2806)
In[7]: fit
Out[7]: Fit(ID=2806, ship=Guardian, name=Guardian fit) at 0x34b28f0
In[8]: fit.calculateModifiedAttributes()
In[9]: fit.remoteReps
doing calcs
Out[9]: {'Armor': 341.3333333333333, 'Capacitor': 0, 'Hull': 0, 'Shield': 0}
In[10]: fit.remoteReps['Armor']
doing calcs
Out[10]: 682.6666666666666
In[11]: fit.remoteReps['Shield']
doing calcs
Out[11]: 0

```

Where as this way (the way used throughout other calculated properties) calculates only once until the results are cleared:


```
In[2]: import eos
In[3]: import eos.db
In[4]: fit = eos.db.getFit(2806)
In[5]: fit
Out[5]: Fit(ID=2806, ship=Guardian, name=Guardian fit) at 0x452e2f0
In[6]: fit.calculateModifiedAttributes()
In[7]: fit.armorRR
do calcs
Out[7]: 341.3333333333333
In[8]: fit.shieldRR
Out[8]: 0
In[9]: fit.hullRR
Out[9]: 0
In[10]: fit.calculateModifiedAttributes()
In[11]: fit.shieldRR
Out[11]: 0
In[12]: fit.clear()
In[13]: fit.calculateModifiedAttributes()
In[14]: fit.shieldRR
do calcs
Out[14]: 0
In[15]: fit.hullRR
Out[15]: 0```

Let me know if there are any questions about it. I'll take a look at the `cycleTime` thing we talked about on Slack - once that's figured out, I would still like for this to utilize what we have instead of pulling all these attributes again.

**Comment by Ebag333** [comment]
>Where as this way (the way used throughout other calculated properties) calculates only once until the results are cleared

I'm not convinced that this is significantly faster (the majority of our properties are updated on access), but we'll give it a shot. :)

I cleaned it and streamlined the logic quite a bit.  <3 named dicts.

I also fixed `cycleTime`, fixed `rawCycleTime`, killed the use of that hacky `floatfloor` method (no need since we're not 2.6 anymore).

Because I was editing it while you were messing with it, not sure how many of your changes made it through the merge @blitzmann . vOv

**Comment by blitzmann** [comment]
> I'm not convinced that this is significantly faster

I'm not interested in convincing anyone of anything. It's not significantly faster, no, since these are simple calculations. But it is more efficient, it's the way it's done elsewhere, and it's the way I want it to continue to be done. No reason to run the same calculation 4 times in a row. :)

> One change I did revert was to go back to a named dict

It seems you go out of your way to do it the way you specifically envision rather than the convention that is used literally everywhere else in the code, but whatever. This should work and I'm not going to continue a back and forth on syntax anymore.

I will say this looks cleaner and more polished than the original, so thanks for that. I will evaluate the changes to the additional properties and merge in if they look good.

**Comment by Ebag333** [comment]
>It seems you go out of your way to do it the way you specifically envision rather than the convention that is used literally everywhere else in the code, but whatever.

There's a number of other properties that are set this way, including the tanking properties, neuts, extra_attributes, and a few others.  So I think you mean figuratively. :)

**Comment by blitzmann** [comment]
was just about to merge this, but found this peculiar bug.

```
[Scimitar, Thingy]

Mark I Compact Power Diagnostic System
Mark I Compact Power Diagnostic System
Mark I Compact Power Diagnostic System
IFFA Compact Damage Control

50MN Y-T8 Compact Microwarpdrive
X-Large Ancillary Shield Booster, Navy Cap Booster 400
Adaptive Invulnerability Field II
Adaptive Invulnerability Field II
Medium Capacitor Booster II, Navy Cap Booster 800

Medium Murky Compact Remote Shield Booster
Medium Murky Compact Remote Shield Booster
Large Murky Compact Remote Shield Booster
Large S95a Scoped Remote Shield Booster

Medium Ancillary Current Router I
Medium Ancillary Current Router I


Light Shield Maintenance Bot I x1
Medium Shield Maintenance Bot I x4


Nanite Repair Paste x500
Navy Cap Booster 400 x18
Navy Cap Booster 800 x9
```

When this fit is first calculated, it returns this:

image


It seems the stats updates twice to get to that point. If you change anything on the fit (state of a module, it recalculates to

image

This seems to be double what it's supposed to be (225 hp/s). Haven't had a chance to look any deeper than that

**Comment by Ebag333** [comment]
Oh. I bet I know why. One of them is `None` so it recalcs all of them, and we always add (because we're looping, so we can't just set it).

**Comment by Ebag333** [comment]
Okay, so this is weird.

Apparently when we rerun the calcs multiple times we aren't running the clear.

Put a breakpoint on the clear, and a breakpoint where the module gets calced.  Runs in this order.

clear()
calc  property
calc  property

The reason why we run into the issue here vs other properties, is that most properties explicitly set the value.  This one can't (because we run it for each module), so we're adding all the runs together (both the ones we expect to, and the ones that  are because  of the multiple recalc issue).

What this means is that at a base level, clear isn't always being run between the recalcs, which is disturbing enough.  (At least this is how it looks to me, I  haven't dug super deep into this yet, so it could be that a few quick and dirty breakpoints aren't telling the whole story.)

Anyway, for this specific property I have a fix for it that'll avoid this or similar issues.  We were running the recalcs if any values were None, now if we run it we'll explicitly set to 0 to clear the None's.  If you didn't have all 4 types loaded, it wouldn't clear the ones you didn't have.  Easy enough fix, we can dig deeper into the clear later.

**Comment by blitzmann** [comment]
> What this means is that at a base level, clear isn't always being run between the recalcs, which is disturbing enough

I doubt this is the case. Recalculation is defined as:

```
    def recalc(self, fit, withBoosters=True):
        pyfalog.info("=" * 10 + "recalc" + "=" * 10)
        if fit.factorReload is not self.serviceFittingOptions["useGlobalForceReload"]:
            fit.factorReload = self.serviceFittingOptions["useGlobalForceReload"]
        fit.clear()

        fit.calculateModifiedAttributes(withBoosters=False)
```

So `fit.clear()` is always run before recalculating. If this wasn't the case, pyfa would be completely broken. In this situation, your fix seems proper in that we were "resetting" in the wrong place in the logic.

PR #996, opened 2017-02-16
Preferences Pane V4

Ebag333:
Replaces #868

This rolls up everything in that PR, catching up with the massive changes we've had over the last few months.

Additionally, added a new preference for being able to control what shows in the context menus.  Currently _all_ context menus are setup so they can be disabled via preference, but we are only currently allowing users to disable specifc ones.  This should make a big difference for folk performance wise if they want to limit what shows, as well as removing some clutter if you don't find a feature useful.

We can easily add in additional context menus to be disabled, just need to expose the preference in the preferences pane.
https://puu.sh/u61ww/46145d9067.gif

(Ignore that the text isn't fully displayed, I'll be adding a commit shortly to made the preferences window a bit wider.)

@blitzmann A lot of future functionality depends on this PR, there are a *TON* of settings that I'd like to expose to let users better control how Pyfa functions.  Until this is merged, I'm at a roadblock there, so I'd like for you to make this a priority after logging.  (Since this doesn't use any external packages, it might be easier to merge in.)


Ebag333:
@blitzmann this is ready to be merged

Ebag333:
Ran some timing checks on this.

If we run the context menu against a AAR with the full menu, it takes about .2 to .25 seconds to run.

If we disable skills, charges (all), and variations, that drops to:
`0.00600004196167`

Doing a little more profiling, the charges sub-menu takes about 0.005 seconds to generate. (Granted that's with a single charge in the submenu.)

Skills take about 0.022 seconds to generate.

So ouch, on those variations.  Those are what's killing the context menu. :(

Ebag333:
>I think pyfa is awesome. But I was wondering if there is a way to turn off the ship attributes that pop up when you hover the mouse over a fitting. 
>See this screen shot. http://imgur.com/a/7MaVl 
>I've looked under preferences but I don't see an option. Am I missing something?

See about adding a preference to turn this off, also disable if debugging is on.

blitzmann:
@Ebag333 I saw that comment on the EVE forums... why someone want to disable it I have no idea. I think it's an awesome feature. But to each their own...

Looking at the code, I don't see anything funky. I agree that we should only allow a select number of context menus to be togglable (the optionsal ones, lest people toggle them off and forget and complain to use that their update is broekn...). Gonna test on my end then merge.

An idea: I don't particularly like the way these things are laid out in the preferneces... It might be interesting to provide the user with a dropdown that, when selected, shows a radio button with the options. This way we have a bit more room, and can even write a small blurp about that particular selection.

More of an iteration change though.

Ebag333:
> I saw that comment on the EVE forums... why someone want to disable it I have no idea. I think it's an awesome feature. But to each their own...

I want to turn it off because it is SUPER annoying when you breakpoint. Other than that, it's fine (though I don't use it really). 

>An idea: I don't particularly like the way these things are laid out in the preferneces... It might be interesting to provide the user with a dropdown that, when selected, shows a radio button with the options. This way we have a bit more room, and can even write a small blurp about that particular selection.

I'd like to, at some point, redo prefs layout entirely. Instead of hard coding them, make them dynamic and have the same layout. This works though, and redoing them will take a lot of wx work, which you know how much I love that.

blitzmann:
- The Logging page - is there any need for this to at the moment? I assume eventually we would want to have the option here to force dump our logs or something, but I assume end users are always going to have Debug Logging Enabled deselected. If it's just going to tell us something we already know as developers (that we are running in debug), then we should probably disable it for now.

Two things that need to be fixed:

This is master:
master

this is this branch:

branch

- Note the odd collapsible icon in the first panel. 
- Note lack of Misc and Targeting

I assume the Misc and Targeting is there, but is not being set to it's sizer correctly in wx. This can cause it to float up to it's parent (hence the ghosting icon). Just a guess, but should be fixed before this is merged in.

Ebag333:
>The Logging page - is there any need for this to at the moment? I assume eventually we would want to have the option here to force dump our logs or something, but I assume end users are always going to have Debug Logging Enabled deselected. If it's just going to tell us something we already know as developers (that we are running in debug), then we should probably disable it for now.

Force dump is a good idea, I didn't think of that but I like it.

The main reason was so we could expose the log path, but I didn't want to add that until we sorted out logbook.  That way, we can direct users to the preferences and they can see what log file is being used.

I think we should keep it for now, because adding a force log dump button is a really good idea, and it'll make it easier for users to find the logs.  The debug flag is nice as a visual confirmation to users if we're  trying to get them to run in debug mode (for testing purposes), since they won't have access to the console and won't see the output there.

>I assume the Misc and Targeting is there, but is not being set to it's sizer correctly in wx. This can cause it to float up to it's parent (hence the ghosting icon). Just a guess, but should be fixed before this is merged in.

Bah, I fixed this once before.  Yeah, the misc targeting is getting pushed to the top for some reason. I'll fix it.

Ebag333:
@blitzmann I had to force push this one, I think I got the wx stuff you changed merged in properly, but can you double check?

FYI, since there were conflicts with development (because of you merging in the remote reps pane) I went ahead and resolved those, so you might pick up some errors because it's trying to load a pane that doesn't exist (yet).

blitzmann:
> I had to force push this one, I think I got the wx stuff you changed merged in properly, but can you double check?

Sounds and looks like you didn't pull from your repo (which I made a commit to) so it asked to force push. You could have brought the change down which would have merged my change with your change and should have been kosher.

I'll take a peak

Ebag333:
>Sounds and looks like you didn't pull from your repo (which I made a commit to) so it asked to force push. You could have brought the change down which would have merged my change with your change and should have been kosher.

I usually use fetch, as from what I understand you should avoid pull when possible.

I know force push isn't kosher.  It might have had to do with me trying (and failing) a rebase earlier too.  But didn't want to lose the changes that I'd made, and I think this worked out in the end....

blitzmann:
@Ebag333 actually, which commit addresses the fixes? I can probably just cherry pick it and merge in with dev. It looks like you made new commits for adding the pref pane stuff which will I believe will cause havok in the history.

Dunno why you tried to rebase...? I normally use merge - don't remember the last time I rebased

blitzmann:
> I usually use fetch, as from what I understand you should avoid pull when possible.

Nope, fetch just tells git "hey, these are what remote has". it's updating what your local git knows about your remote git. Pull does a fetch (updates information on the remote) and then pulls the changes that the remote has into your branch, which is what you wanted.

Ebag333:
>Dunno why you tried to rebase...? I normally use merge - don't remember the last time I rebased

https://www.derekgourlay.com/blog/git-when-to-merge-vs-when-to-rebase/

>Rule of thumb:
>When pulling changes from origin/develop onto your local develop use rebase.
>When finishing a feature branch merge the changes back to develop.

vOv

>@Ebag333 actually, which commit addresses the fixes? I can probably just cherry pick it and merge in with dev. It looks like you made new commits for adding the pref pane stuff which will I believe will cause havok in the history.

It's a little weird, yeah.  If you want I can create a new clean branch and replay the commits so that the history is clean.

I'd rather not try and cherry pick the commits straight into dev, that sounds even messier.

Ebag333:
>It's a little weird, yeah. If you want I can create a new clean branch and replay the commits so that the history is clean.

Actually, let me do that. Then I can properly add the new states pane to this, and we won't have to mess with it again.

PR #999, opened 2017-02-17
Rewrite handler for non-unicode paths.

Ebag333:
Rewrite handler for non-unicode paths.  Add option to force use of a particular codec. If the codec set via parameters fails, will auto fall back to trying every available codec.

@blitzmann can you test this and see about doing a new build?  Please note that this PR is pointed at Master and *NOT* Development.

Also, I did a thing! \o/
https://github.com/pyfa-org/Pyfa/wiki/Command-Line-Parameters-(Startup-Options)

### DIFF
--- a/config.py
+++ b/config.py
@@ -203,14 +203,20 @@ def parsePath(root, Append=None, Create=False):
             if codec_return and Create:
                 path_exists = parsePathCreateDir(codec_return)
 
-            if path_exists and os.path.exists(path_exists):
+            if path_exists and (os.path.exists(path_exists) or os.path.isfile(path_exists)):
                 return codec_return
             else:
                 continue
 
-        return
+        return None
     else:
-        return root_path
+        if root_path and Create:
+            path_exists = parsePathCreateDir(root_path)
+
+        if root_path and (os.path.exists(root_path) or os.path.isfile(root_path)):
+            return root_path
+        else:
+            return None
 
 
 def parsePathApplyCodec(root_path, apply_codec):
--- a/config.py
+++ b/config.py
@@ -16,6 +16,9 @@ except ImportError:
 debug = False
 # Defines if our saveddata will be in pyfa root or not
 saveInRoot = False
+# Defines if we use a hard set codec or not
+# See https://docs.python.org/2/library/codecs.html#standard-encodings
+codec = None
 
 # Version data
 version = "1.27.0"
@@ -53,11 +56,6 @@ def isFrozen():
         return False
 
 
-def __createDirs(path):
-    if not os.path.exists(path):
-        os.makedirs(path)
-
-
 def defPaths(customSavePath):
     global debug
     global pyfaPath
@@ -72,16 +70,15 @@ def defPaths(customSavePath):
         logLevel = logging.WARN
 
     # The main pyfa directory which contains run.py
-    # Python 2.X uses ANSI by default, so we need to convert the character encoding
     pyfaPath = getattr(configforced, "pyfaPath", pyfaPath)
     if pyfaPath is None:
-        pyfaPath = getPyfaPath()
+        pyfaPath = getPyfaPath(None, True)
 
     # Where we store the saved fits etc, default is the current users home directory
     if saveInRoot is True:
         savePath = getattr(configforced, "savePath", None)
         if savePath is None:
-            savePath = getPyfaPath("saveddata")
+            savePath = getPyfaPath("saveddata", True)
     else:
         savePath = getattr(configforced, "savePath", None)
         if savePath is None:
@@ -90,7 +87,7 @@ def defPaths(customSavePath):
             else:
                 savePath = customSavePath
 
-    __createDirs(savePath)
+    savePath = getSavePath(None, True)
 
     if isFrozen():
         certName = "cacert.pem"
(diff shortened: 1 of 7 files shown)


blitzmann:
I don't want to test ever codec known to man as I stated in slack. Give me a bit to research the issue. A good thing to figure out: why was it working I the previous builds but not in this one? It would seem the path.join did fine with this codex.

Try to research into that while I look into better ways of doing this.

blitzmann:
Also, setting anot optional codex is a rather interesting idea. I would rather just give user the option. To do that than to do a loop

blitzmann:
(Mobile typing sucks)

Ebag333:
>I don't want to test ever codec known to man as I stated in slack. 

>Also, setting anot optional codex is a rather interesting idea. I would rather just give user the option. To do that than to do a loop

The issue with not doing the loop is that you can easily have multiple codecs in use, especially in cases of things such as being on Linux but using a Windows hard drive.

The reason why we still do the loop is because if they have multiple codecs in use, then it'll fail.

In testing this, I had multiple codecs, `cp1251` for my saveddata path and `cp1252` for my pyfa directory.

The loop still needs to exist because the forced codec can fail.  If I set it to `cp1251`, and didn't have the loop, it'd fail when it hits the path typed as `cp1252`.

Order of operations is this:
1) Test if unicode.
2) Try the forced codec.
3) Loop through the codec dict. As soon as we find a codec that looks like it works, bail.

3 is the tricky part.  We sort it by the more common codecs, and if you want to potentially shorten the loop we can drop the special ones (though that might impact Japanese and Korean users, as they can use ISO instead of UTF or CP).

The other tricky part is that we might use the _wrong_ codec, but it provides what appears to be a valid path.  I don't see a way around this, but the risk is that we'll end up creating a directory structure somewhere we didn't expect to.  (This is actually why I added the launch option to let the user explicitly set the codec, so it'd try that one first.)

For example, we might want to point to:
`C:\Users\Гоша Егорян\.pyfa`
and end up with:
`C:\Users\昨夜のコンサートは最高でした\.pyfa`

What we can do to help with that is to always test the path.....in fact I'm going to do exactly that. (Currently we only test if we're creating it.)

The other benefit is that if it's allowed by the OS, we'll simply always use that path so it's not the end of the world.  (Pyfa writing to an unexpected place and working is better than Pyfa trying to write to the right place and not working.)  Plus if we ever get the preferences pane PR merged, that'll expose that to the user, so it won't be hard to find.

That's also why I put the less common codecs toward the back, and it's a dict and not a set (so it stays ordered).  We can also pull other codecs we want to try to the top, so they get tried before the less common ones.

>A good thing to figure out: why was it working I the previous builds but not in this one? It would seem the path.join did fine with this codex.

Probably because of duck typing.  On my very vanilla Windows system, a standard (no special characters) path actually fails the unicode check (shows as a string) because it's actually typed as `cp1252` (AKA `windows-1252`).  So if you do strict typing it fails, but because it doesn't have any non-unicode chars in it, duck typing means it works even though it's not actually unicode.

Remember, we initially went down this road to help people who couldn't use Pyfa because duck typing didn't work in their case. So yes, we broke a few folks because now we're strictly typing unicode, and they _aren't_ one of the three codec types we were handling (unicode, utf8, cp1252).  (In some of those cases, I'm amazed it worked.  Windows has some built in OS level functionality that I suspect was kicking in to help us find directories when they weren't quite right.)

We also didn't handle all of our checks properly, so that's also been fixed.

>Give me a bit to research the issue.

Feel free to research the issue.  I've looked into it pretty extensively, but wouldn't mind a second set of eyes on it.

Previously I couldn't get certain character sets to work on Windows, it was Windows stripping them out (and replacing it with `?`) before it ever hit pyCharm.  Now that I figured out that it's the locale setting, I can test ones I couldn't before (like Russian for example).  So that helps a lot. :)

I think that this is a solid solution.  We really don't hit these functions very often.  We hit `getPyfaPath` 8 times (only 3 outside of `config.py`) and `getSavePath` 5 times (only 2 outside of `config.py`).  So yes this will slow down our startup _slightly_, but the impact is going to be darn near imperceptible (even if you hit one of the less used codecs).

I  ran concurrency diagram for this, and found that my startup time wasn't any noticeably slower than before (sub 14s for the empty database I was testing, and a empty DB takes longer to start than a full DB).

Ebag333:
>Speaking of testing....TESTS! In Russian!

We now have _five_ tests!

blitzmann:
Closing this as I've decided to take a different approach when it comes to these locale issues, namely to revert back to how we were doing it and address the issues as they come up (see #1010). The reasoning for this is because, aside from the odd Hebrew one (which I'm still bewildered about, but need confirmation on), the other locale issue we were dealing with was easily fixed with a simple `.encode('utf-8')` to the path, and I personally can't ever get unicode stuff to break for me after that fix ;_;

I appreciate the work and research you put into this - if we need to revisit this we can always revert the revert and reopen this. But I would prefer to take these issues as the come up and develop a fix specifically for it -- obv if this means we have 10 different fixes, we'll have to consider a better, more universal solution.

Thanks!

Ebag333:
@blitzmann please revert and reopen..  Your commit does not fix the issue.

I went back to Master, fetched your commits, and then cherry picked my locale test PR.  I then ran the tests, it passed because of course we have in the test to use `cp1251` for Russian (which my OS is currently set to).

I changed it to `utf8` and the test failed.
`UnicodeDecodeError: 'utf8' codec can't decode byte 0xe7 in position 18: invalid continuation byte`

This test is a simple.  Get the actual folder based on our path string (same exact method used by Pyfa), read the `testcodec` file, and verify it has `True` in it.

I ran the latest master, and it worked.  Yay!

That's thanks to something you added (which is really a good idea):
`root = unicode(root, sys.getfilesystemencoding())`

Because Windows is explicitly set to Russian, Python can use that to convert.

All fixed, right?

Afraid not.  The problem is that this works _as long as_ I have  the OS set to Russian.  If I flip it back to English (or any other language) then I get this traceback:
```
  File "C:\Program Files (x86)\JetBrains\PyCharm 2016.3.2\helpers\pydev\pydevd.py", line 1596, in <module>
    globals = debugger.run(setup['file'], None, None, is_module)
  File "C:\Program Files (x86)\JetBrains\PyCharm 2016.3.2\helpers\pydev\pydevd.py", line 974, in run
    pydev_imports.execfile(file, globals, locals)  # execute the script
IOError: [Errno 2] No such file or directory: 'C:/Users/Ebag333/Documents/GitHub/Ebag333/Pyfa/tests/test_locale/????/Pyfa/pyfa.py'
```

`sys.getfilesystemencoding()` no longer works because the default system encoding is English.  _BUT_ the catch is, that doesn't stop me from using Russian (or any other language I have configured) at all.

Here's some random text in Russian (while the locale is set to English): рш еруку Ш вщ тще лтщц црфе ершы ьуфты

The magic setting in Windows that makes `sys.getfilesystemencoding()` work is exactly the equivalent to the flag I implemented in this PR.  It's just a notification to apps that "hey, might want to use this encoding."  It's _not_ actually changing anything as far as OS supported languages go.

Additionally, `sys.getfilesystemencoding()` on Windows will only support either 1 (English) or 2 (English and 1 selected encoding).  We will have users that have more than 2 languages used, for example one user who posted on the issues was in Ukraine, but explicitly set to Russian, so could very possibly have 3 (since Ukrainian is it's own).



So while your fix will fix it _some_ (or perhaps even most) of the time, it won't fix it _all_ of the time.  You have to  go to the lengths of this PR in order to do this.

If we do not go this route, then we'll have to tell users that have issues they have the following options:
1) Change the reported locale (not be possible for users without admin rights, or too cumbersome for other users since it requires a reboot every time you change it).
2) Change all paths to be English only (may not be possible for users without admin rights, and really?)

Those will be the only two options they have.

blitzmann:
Two things:

* I've reverted simply so that we go back to the way things were so that I can get a release out that includes fixes. I have no doubt more work is needed to include support with as many locales as we can.
* As stated, I'm not sure the tests are or ever will be 100% accurate. As you've stated, if we change the system locale, then it changes the results of the tests, and we can't programatically change the system locale.

There is the concern that using Russian characters in the path with an English system locale will break, but to that I ask: why the fuck are you using a Russian path when you have your locale set to English? I think it's safe to assume folks that have their locale set to a specific thing are also using paths that match that.

I am sure there are certain edge cases, of that I'm sure, but are they worth a complete upheaval of how our path handling works, where we go to great lengths to address <1% of the users and potentially introduce bugs? I don't think so, but we'll see. With EFT as cold as a white walker, I'm sure we will continue to pick up users of all nationalities going forward that can be real-world test cases.

This will remain closed for now. My goal here is to hit the reset button and start anew with possible locale issues, and then focus on why it happens and fix the issue specifically. If it turns out we need to reopen this, we will, but I don't want to chase issues that will almost never happen in the real world. I would like users to actually report the issue if they have a problem (and they'll know once I get the GUI notification window going). 

btw, I did find a possible solution to this whole thing:

```
import sys
reload(sys)
sys.setdefaultencoding("utf-8")
```

Makes py2 act like py3 in that all string defaults are in utf-8. However, this is a discouraged practice (since third party modules may assume a different default, and equality checking might be done differently). I don't think I want to implement this, but it's good to know it's a thing.

Ebag333:
>As stated, I'm not sure the tests are or ever will be 100% accurate. As you've stated, if we change the system locale, then it changes the results of the tests, and we can't programatically change the system locale.

Tests will never be 100%.  I had two goals with that specific PR.
1) get them to the place where they either work without error or they don't run at all (ala Windows), but they can be ran on demand to test specific scenarios.
2) To set up an environment where we can test very specifically and abstractly, something that the Pyfa codebase is _not_.  We're about halfway there. Keep in mind that not all of these tests will be automated, for example a huge part of the reason for doing this is so you can quickly copy/pasta Pyfa's .py files into a language directory, and then run it.  (Which I did, which worked wonderfully.)

>There is the concern that using Russian characters in the path with an English system locale will break, but to that I ask: why the fuck are you using a Russian path when you have your locale set to English? I think it's safe to assume folks that have their locale set to a specific thing are also using paths that match that.

I can't speak to every scenario, but it's not uncommon.  I used to work for a multinational, and that's exactly how they were setup most countries (English everything default, install charset for the local language so they can hot switch between the two).  Granted that's a business environment so not quite the same, but the point still stands.

English is the Windows default, so if you aren't using a regional build (or you live in Europe, since the Windows N build also defaults to English) you will end up with the standard `cp1252`.

Let me list all the places you can change regional settings in Windows:
- Date & Time
- Region and Language
- Speech
- Input Methods
- Region
- Additional Settings (under Region)
- Location
- Date/Time/Number Formats
- Language Preferences
- Input Methods
- Handwriting
- Language Advanced Settings
- Administrative (under Region) <- This is the one we need set

And there's probably a few that I missed in there.  So are you  really suggesting that out of 17 panes that can be changed, you'd *REALLY* be surprised that _one_ could get missed?  Especially since it _only_ would impact applications that explicitly do not handle the unicode conversions (meaning that the vast majority of the applications out there aren't affected by it)?  And guess what? With this PR, we wouldn't be affected by it either. :)

>I am sure there are certain edge cases, of that I'm sure, but are they worth a complete upheaval of how our path handling works, where we go to great lengths to address <1% of the users and potentially introduce bugs?

"Complete upheaval" is a wee bit melodramatic.  It's literally only called a double handful of places in the codebase, so this change is tiny.  Yes, it's a rather important bit of code, but all code is equally important.  (Except you, graphs code. Go sit in the corner and think about how terrible you are.)

We're handling paths the exact same way that we were before, just centralizing the handling of it (rather than handling it multiple ways), adding error handling, and adding handling for multiple codecs. It is indeed _extending_ the functionality, but 

>With EFT as cold as a white walker, I'm sure we will continue to pick up users of all nationalities going forward that can be real-world test cases.

Realistically, no, we won't.  We have enough trouble finding a person running Linux or Mac to test for us, trying to add a layer on top of that for each language?

>I don't want to chase issues that will almost never happen in the real world. 

Except that we were able to replicate some of these issues.  And there's ones we haven't been able to replicate that still occur.

You are right in that we should step back a bit, but I think we need to create some test scenarios to test specific functionality outside of Pyfa, address the issues we have from those *INSIDE* Pyfa, and that will solve a lot of our problems.

By reverting back we now handle each file/folder looking in it's own way, and it's not centralized at all. So we can spend a lot of time reinventing wheels instead of fixing it once across the board.

blitzmann:
> Except that we were able to replicate some of these issues. And there's ones we haven't been able to replicate that still occur.

The path fix resolved two issues that were known: 
- The French path, which is resolved with another method 
- The Hebrew path, the status of which is now unknown until the user gets back

And it introduced the Russian path issue. This PR does a brute force method which I feel is unnecessary, especially now that I have made the decision to step back deal with issues as they come up. I don't want to spend a great deal of time worrying about Russians that use English system and using French paths or whatever - that effort is more valuable elsewhere. 

Basically, I do not want a heavy handed solution to this. I doubt that pyfa is the only project that has ever needed to support unicode paths, and during my research I have never seen anyone suggest doing a brute force codec test (which also will not work 100% of the time). 

Thanks :)

PR #1006, opened 2017-02-21
Tests for Pyfa

Ebag333:
Just a quick little PR to add a few tests for Pyfa.

Also, found a small bug in `floorFloat`.  We were handling it a bit hacky for 2.6 compataibility, and this meant we would round the wrong way for negative values (-1.1 rounded to -1, when it should round to -2).  Didn't see any use cases when we'd hit that, but rewrote it anyway to use `floor` since we're  not supporting 2.6 anymore, and it's MUCH simpler.

### DIFF
--- a/eos/mathUtils.py
+++ b/eos/mathUtils.py
@@ -17,15 +17,9 @@
 # along with eos.  If not, see <http://www.gnu.org/licenses/>.
 # ===============================================================================
 
-import decimal
+from math import floor
 
 
 def floorFloat(value):
-    """Round float down to integer"""
-    # We have to convert float to str to keep compatibility with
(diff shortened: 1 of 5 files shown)


Ebag333:
All these commits are because Dev and Master are out of sync. :(

blitzmann:
zomg 4.49% coverage!

Ebag333:
>zomg 4.49% coverage!

And a test against `attributes.py` is actually quite valuable.

It's a fairly hacky test, we're just testing a single attribute and making sure that the returned values/types are what we expect, but that's a good thing in case  someone makes low level attribute changes!

The other tests? Meh, mostly to just get some coverage.

PR #1008, opened 2017-02-24
Locale Testing

Ebag333:
This isn't so much a test our our locale handler per say (though we can switch to pointing at that) as it is a way to have a set of directories pre-setup to make it easier to test locales in different environments.

Eventually would like to have a good number of locales in here.

@blitzmann this is not ready to merge, but feel free to poke at my branch and experiment with this.  As an example, when I set the Russian test to use `UTF8` it fails on Windows (have to use `CP1251`).  Hoping this might help your testing some.


Ebag333:
```
tests/test_locale/знаф/test_codec_russian.py                                                                                                 18      4    78%
tests/test_locale/פטכש/test_codec_hebrew.py                                                                                                  18      4    78%
tests/test_locale/测试/test_codec_chinese_simplified.py                                                                                        18      4    78%
```

For some reason Hebrew and Chinese aren't running for me in Windows.  But they're both working in Linux. 
 I think I'm missing something from Windows itself.

blitzmann:
This could be nice, but isn't part of the problem with locales the system setting? For example, I changed the pyfa path to include `Téléchargements`, and it works (with 6e643d7) on English locale, but failed with Russian locale (granted, this was a cxfreeze failure, not a pyfa failure, so might not be related). 

So, unless we change the system locale, the tests could prove to be unreliable? Thoughts?

blitzmann:
> For some reason Hebrew and Chinese aren't running for me in Windows. But they're both working in Linux.
> I think I'm missing something from Windows itself.

See above - probably have to change the system locale for it to work. But question: I assume travis-ci runs in linux, so the tests should work normally there, yeah?

Also haven't looked to see what the code does yet <_<

Ebag333:
>This could be nice, but isn't part of the problem with locales the system setting? For example, I changed the pyfa path to include Téléchargements, and it works (with 6e643d7) on English locale, but failed with Russian locale (granted, this was a cxfreeze failure, not a pyfa failure, so might not be related).

>So, unless we change the system locale, the tests could prove to be unreliable? Thoughts?

For Windows, it only runs English (since Windows always has English) and the locale you've selected.  It can't get into the other folders, `os.walk` returns `???` (of varying lengths) for the non English and selected languages.  So testing is a very manual process as you have to switch between the languages, but you *CAN* test, which is the whole idea.

No idea the behavior on OSx, but it doesn't matter since even if the test fails the theoretical dev can know that it'll fail and ignore it. I'm not even sure which locales are supposed to be used for OSx, to be honest.

>See above - probably have to change the system locale for it to work. But question: I assume travis-ci runs in linux, so the tests should work normally there, yeah?

Travis-CI is indeed Linux (Debian I believe), so everything is UTF-8.  Which is why they all ran, and why they all passed. That was easy!

Ebag333:
>Linux (Debian I believe), so everything is UTF-8.

Just a note, but this actually isn't true.

Everything by default is UTF-8 on Linux, but you can still hit other encodings (such as mounting a Windows drive).

Ebag333:
Need to  figure out why the latest test fails on Linux but works on Windows.....

Ebag333:
>Need to figure out why the latest test fails on Linux

You might want to revert this merge. One test fails on Linux for some reason (so Travis).

blitzmann:
eh... fuck

@Ebag333 do you have access to change the labels on your PRs?

blitzmann:
(reverted btw, thanks for the notification)

PR #1012, opened 2017-02-25
Add Properties Pane

Ebag333:
Just a quick little dev QoL.

Adds a pane that exposes properties on an item, so we don't have to set breakpoints just to see what that one @property from `modules.py` set.

Need to check where we're  pulling this from, I think there's a better spot.

https://puu.sh/uiAOR/a316d5a05f.png

### DIFF
--- a/gui/itemStats.py
+++ b/gui/itemStats.py
@@ -1365,8 +1365,6 @@ class ItemProperties(wx.Panel):
         self.paramList.SetColumnWidth(0, 110)
         self.paramList.SetColumnWidth(1, 1500)
         self.paramList.setResizeColumn(0)
-        #self.imageList = wx.ImageList(16, 16)
-        #self.paramList.SetImageList(self.imageList, wx.IMAGE_LIST_SMALL)
 
         if self.stuff:
             names = dir(self.stuff)
@@ -1374,7 +1372,6 @@ class ItemProperties(wx.Panel):
             names = dir(self.item)
 
         names = [a for a in names if not (a.startswith('__') and a.endswith('__'))]
-
 
         idNameMap = {}
         idCount = 0
@@ -1389,6 +1386,7 @@ class ItemProperties(wx.Panel):
             except Exception as e:
                 # TODO: Add logging to this.
                 # We couldn't get a property for some reason. Skip it for now.
+                print(e)
                 continue
 
             index = self.paramList.InsertStringItem(sys.maxint, attrName)
--- a/gui/itemStats.py
+++ b/gui/itemStats.py
@@ -1357,26 +1357,39 @@ class ItemProperties(wx.Panel):
         self.paramList.ClearAll()
         self.PopulateList()
         self.Thaw()
-        self.paramList.resizeLastColumn(100)
+        self.paramList.resizeLastColumn(1500)
 
     def PopulateList(self):
         self.paramList.InsertColumn(0, "Attribute")
         self.paramList.InsertColumn(1, "Current Value")
         self.paramList.SetColumnWidth(0, 110)
-        self.paramList.SetColumnWidth(1, 400)
+        self.paramList.SetColumnWidth(1, 1500)
         self.paramList.setResizeColumn(0)
-        self.imageList = wx.ImageList(16, 16)
-        self.paramList.SetImageList(self.imageList, wx.IMAGE_LIST_SMALL)
+        #self.imageList = wx.ImageList(16, 16)
+        #self.paramList.SetImageList(self.imageList, wx.IMAGE_LIST_SMALL)
 
-        names = dir(self.item)
+        if self.stuff:
+            names = dir(self.stuff)
+        else:
+            names = dir(self.item)
+
         names = [a for a in names if not (a.startswith('__') and a.endswith('__'))]
-        # names = names.sort()
+
 
         idNameMap = {}
         idCount = 0
         for name in names:
-            attrName = name.title()
-            value = getattr(self.item, name)
+            try:
+                if self.stuff:
+                    attrName = name.title()
+                    value = getattr(self.stuff, name)
+                else:
+                    attrName = name.title()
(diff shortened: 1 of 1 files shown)


blitzmann:
Well that was quick.

Looks cool. Is it possible to click and ctrl+c a cell to copy the information (for those long ones?). I've been meaning to look into that to apply to the other grids in pyfa anyway, don't know if you came across it

blitzmann:
> Need to check where we're pulling this from

Also, check what now?

Ebag333:
It was a super quick add.

I was just thinking of double clicking to "drill down" but dunno if I want to mess with it beyond what we got right now.

>Also, check what now?

`self.item` is where I was pulling from initially.  I switched it to `self.stuff` which is more useful (if it exists, otherwise we fall back to item).

blitzmann:
I'm not seeing things like the `cycleTime` property... am I missing something?

Ebag333:
>I'm not seeing things like the cycleTime property... am I missing something?

No you're not.  We might have to fetch the fit from the fit service to get that. `self.stuff` seems to not really contain everything we actually use.

Yeah, actually looking into it, we're getting it really early in the process and just pulling the item from the market.  I'm not sure why.

Here's where it starts:
https://github.com/pyfa-org/Pyfa/blob/master/gui/itemStats.py#L86-L93

So since it's not the actual fit object, it won't have most of the attributes.  It's a little odd to me though that we're passing the market instance down the line.

blitzmann:
No, `stuff` should be the module instance. Otherwise we wouldn't be able to get the modified attribute values in the attributes tab. If it was just the `Item` instead of `Module` then it would show the default values.

you sure `dir()` brings back our defined properties?

blitzmann:
Oh heres why:

`names = dir(self.item)` you aren't using stuff, you're using the item

blitzmann:
image

Ebag333:
>No, stuff should be the module instance. Otherwise we wouldn't be able to get the modified attribute values in the attributes tab. If it was just the Item instead of Module then it would show the default values.

Yeah, it only gets the actual market item if it's missing (which  makes sense).

>you sure dir() brings back our defined properties?

Yup. Slick little way of walking through the properties.

>Oh heres why:
>names = dir(self.item) you aren't using stuff, you're using the item

Are you looking at the latest version? :)

```
        if self.stuff:
            names = dir(self.stuff)
        else:
            names = dir(self.item)
```


This all comes from  the `victim` code which is a bit...complicated. 

It seems to be working.....just we scatter stuff so far around it doesn't necessarily show what you might expect.

Ebag333:
Remember when I said I thought that there were stacktraces deep in the @properties, and you were all like "No way dude!"

```
  File ".\gui\contextMenu.py", line 138, in handler
    menuHandler.activate(context, selection, i)
  File ".\gui\builtinContextMenus\itemStats.py", line 65, in activate
    ItemStatsDialog(stuff, fullContext)
  File ".\gui\itemStats.py", line 111, in __init__
    self.container = ItemStatsContainer(self, victim, item, itmContext)
  File ".\gui\itemStats.py", line 201, in __init__
    self.properties = ItemProperties(self.nbContainer, stuff, item, context)
  File ".\gui\itemStats.py", line 1333, in __init__
    self.PopulateList()
  File ".\gui\itemStats.py", line 1384, in PopulateList
    value = getattr(self.stuff, name)
  File ".\eos\saveddata\module.py", line 670, in cycleTime
    speed = self.rawCycleTime + reactivation
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```

There are stacktraces deep in the properties. :)

That one is coming from the `cycleTime` code.  But right now a `Heat Sink II` is failing to open the window because of it.

Hmm. I wonder if the rewrite I just did of cycleTime fixes this..... 🤔

blitzmann:
>Remember when I said I thought that there were stacktraces deep in the @properties, and you were all like "No way dude!"

I never said they didn't happen. What I said that pyCharm shows them happening because pyCharm is accessing those properties during debugging in the middle of execution when all the data may or may not be there.

This seems like a case where the heat sink, being a passive module, doesn't have a cycle time. It doesn't break in the rest of pyfa because we never have to access that property for calculations or display purposes for the Heat Sink.

A simple wrapper around the get attr should suffice :)

Ebag333:
>I never said they didn't happen. What I said that pyCharm shows them happening because pyCharm is accessing those properties during debugging in the middle of execution when all the data may or may not be there.

Just a little humor.  These are hard to find, though, because if you pause execution the actual stacktraces are hidden by the "just can't run right now" ones.  I know you'll hate it, but I wouldn't mind wrapping the @properties in try/except simply so  we can log when they do fail (since they are such a base level and buried deep, the stacktraces don't bubble up like they normally would).

But after logbook gets merged. I don't want to add a bunch of logging messages then have to convert it over.

>This seems like a case where the heat sink, being a passive module, doesn't have a cycle time. It doesn't break in the rest of pyfa because we never have to access that property for calculations or display purposes for the Heat Sink.

Either that or a reactivation delay, which it also doesn't have.  Either way, it shouldn't be stacktracing, but there's a fair chance that's already fixed so I'm not touching it here.

>A simple try wrapper around the get attr should suffice :)

Way ahead of you boss. :)

Ebag333:
>Hmm. I wonder if the rewrite I just did of cycleTime fixes this..... 🤔

It does! A nice neat cycleTime of 0 instead of a stacktrace.  So #995 fixes that at least.

blitzmann:
We are not going to wrap all our properties in try/catch. You need to get out of the mindset that everything must be error handled simply because it has the possibility of failing. Same thing with default values and `getattr()`. Those will clutter things up very fast with the way we have things laid out currently. :)

This is one of those things where returning a default `0` *would* be preferable. Right now `rawCycleTime` is defined as `attr or otherAttr` when it can be defined as `attr or otherAttr or 0`

PR #1021, opened 2017-02-27
Add more functionality to preferences window.

Ebag333:
Adds a few functionalities we've talked about.

1) For RAH: If uniform damage profile, set RAH to uniform (old behavior). This defaults to off (defaults to new behavior).
2) Add path for log to the logging window preference pane, so users can find it easier.
3) Add a "dump all logs" button to the preference pane. If a users pushes this button, it will write a critical log which will trip the fingers crossed handler. All logs in memory and all future logs will then be written to the file
4) Ability to disable the ship browser traits popup.  This one is annoying for some folks, and super annoying when debugging with breakpoints.  If you disable it after popups have been generated, those popups will continue to show but new ones will not.


eos/effects/adaptivearmorhardener.py
```diff
@@ -13,6 +14,13 @@
 def handler(fit, module, context):
     damagePattern = fit.damagePattern
 
+    sFit = Fit.getInstance()
```
blitzmann: eh... problem here. We can't import services from inside eos.

I mean, obviously we can, but we shouldn't (separation of concerns, etc).

To bypass that limitation, we can set the setting (and any other "engine" settings in the future) on `eos.config`, and then import and use that in the effects that need it. It will take a bit more engineering, but that would be the way to do it to keep eos completely separate.

Or you could set the setting on the fit itself (since we have the fit object in the effects). This is how we do the factor reload setting IIRC. Might be easier and quicker that way.

eos/effects/adaptivearmorhardener.py
```diff
@@ -13,6 +14,13 @@
...
+    static_adaptive_behavior = sFit.serviceFittingOptions["useStaticAdaptiveArmorHardener"]
+
+    if (damagePattern.emAmount == damagePattern.thermalAmount == damagePattern.kineticAmount == damagePattern.explosiveAmount) and static_adaptive_behavior:
+        pyfalog.debug("Setting adaptivearmorhardener resists to uniform profile.")
+        return
```
blitzmann: I think this is an excellent compromise for this particular problem from before. :)

Ebag333:
>eh... problem here. We can't import services from inside eos.

Yeah, settings is something I'd like to talk about.

We have too many settings that cross boundaries, especially once we start adding in more advanced functionality (like the new capsim).

Way I see it, we have a few options.
1) Jump through a ton of hoops to try and keep them separate even when they aren't.
2) Move all settings to Eos and just say "screw it" and let GUI settings live there.
3) Break settings out into a discrete service that can be used be Eos or the GUI.

Personally, I think option 3 is the best. We have too much overlap to have a clear bright line.  Yes, that means Eos isn't standalone (you need Eos *AND* settings) but let's be realistic here, Eos isn't a stand alone package to begin with (ala new Eos).  And trying to fully separate the two (Eos and GUI) to have each run without the other being part of the same package would be difficult to say the least (mostly from the GUI side, but plenty of gotchas from the Eos side).

>Or you could set the setting on the fit itself (since we have the fit object in the effects). This is how we do the factor reload setting IIRC. Might be easier and quicker that way.

It might be quicker and easier, but it's also rather hacky, and it blurs the line.

Ebag333:
Pictures!

https://puu.sh/umCNQ/f237ba66fd.png
https://puu.sh/umCMF/63a9c44a86.png

Ebag333:
>Ability to disable the ship browser traits popup. This one is annoying for some folks, and super annoying when debugging with breakpoints. If you disable it after popups have been generated, those popups will continue to show but new ones will not.

Added another asked for preference.  I can't find the issue that mentioned this, but it's a good idea.

blitzmann:
I still want as much separation as possible between eos and other pyfa things. The way I see it: pyfa uses eos, not the other way around. I know of a few folks thbat have, over the years, used eos as "standalone" (with probably a few hacks) for their own personal projects, and I want to maintain that option.

Yes, a better solution for infusing settings into eos should be available in the long term. I have an idea and I don't think it would be too difficult to implement (need to review the settings code again). But for now please separate them out. :)

blitzmann:
> There are multiple comments in the code about moving sFit sittings, and you've lamented before about them. It's also been a source of confusion for more than one dev.

Completely agree, and we should get it out of the service and into it's own thing.

> Share with the class?

Since the settings are accessed like dictionaries, my thought was to simply have a dictionary in eos.config and replace it with a settings class. To call it, import settings in effects / type classes and call with `settings['<setting>']`. It should work with both the basic dictionary or the settings class from pyfa.

I've got a new commit on dev that demonstrates this, seems to work well (haven't actually built out the ability to set anything on it, so that would need to be tested)

blitzmann:
@Ebag333 gonna implement those setting changes and push to your branch for review / additional testing. if everything looks fine there, we'll merge this.

Ebag333:
I've been super crazy busy lately so haven't been able to get back to this. I'll be back on a "normal" schedule next week. 

Looking forward to looking at this. Your solution sounds like a good idea.

blitzmann:
No worries, haven't heard from you lately so I figured you were busy or dead. <_<

Ebag333:
Okay, so looking this over....

First off, I don't really see this as removing the dependency between Eos and the settings service.  It abstracts it through `config.py`, but it's still there.

If we're going to go with having Eos dependent on root level files, why not just bump up settings out of services and to the root?  Then we wouldn't have to migrate settings from one setting group to another, and deal with that refactoring (not terribly complicated, but still extra work).

https://puu.sh/uHxO3/513debdf06.png

I think that illustrates the way you have it with your most recent commit (on the left) vs moving settings up a level (on the right).

What about this (as a thought).  It'd be a bit more work, but what if we moved all the settings into the database, and let it live entirely in Eos.  Then the GUI can call/set it via Eos.

blitzmann:
> First off, I don't really see this as removing the dependency between Eos and the settings service. It abstracts it through config.py, but it's still there.

Yeah... that abstracts it out so that we're not dependent on pyfa-specific modules. If someone were to utilize eos, they can change the config file or write their own implementation of the settings. Not sure of the confusion, and that flowchart doesn't help clear it up lol

> What about this (as a thought). It'd be a bit more work, but what if we moved all the settings into the database, and let it live entirely in Eos. Then the GUI can call/set it via Eos.

Most settings are pyfa-specific, so having settings defined in the database only makes sense for pyfa. I'm aware this is me being picky, but there it is (it would also be a right pain in the ass to migrate to database-backed settings for very little gain)

blitzmann:
Oh wait, are you talking about `pyfa/config.py`? This doesn't (read: shouldn't, I got stuck on something else and haven't actually implemented this fully yet) go out to the root pyfa config.py, instead it uses the eos config.py. Maybe that clears up the confusion.

Ebag333:
>The only thing the pyfa config.py does is initialize the eos config with values from the pyfa settings manager. :)

Right. So it means that there is a a hard dependency on \config.py from \eos\config.py to \service\settings.py.  In other words, if you remove the link in \config.py, it doesn't work.

The flowchart was supposed to be levels.  The config.py there is the one at the root (not the one under \eos\....good gravy our naming is confusing).

>Most settings are pyfa-specific, so having settings defined in the database only makes sense for pyfa.  I'm aware this is me being picky, but there it is (it would also be a right pain in the ass to migrate to database-backed settings for very little gain)

Well...not really.

It'd be fairly trivial to migrate the settings over to the DB.  And Eos would just handle the connection to the DB, it can be agnostic to what's stored there.  And since Eos already handles all the DB connections, that would make a lot of sense from that perspective.


Point is, unless you're rejiggering your approach quite a bit from what you have currently, there's going to be a hard dependency from Eos back to Settings which is what you wanted to avoid.  (Yes, only a single line, but that's all it takes!)

Personally, I vote for just saying screw it and not worrying about it.  If someone wants to use Eos "stand alone" it's pretty trivial to add Eos + services.  There's still no GUI requirement, and it's not really much different than what it is today.  (Plus I question whether anyone's actually use Eos on it's own.  I could only find one person who had, and they quit doing so years ago.)

If you're dead set against that, then let's just bump settings up a level. It's easy, it's simple, and could be done quickly without any significant changes.

blitzmann:
> So it means that there is a a hard dependency on \config.py from \eos\config.py to \service\settings.py

No, there's not. eos will still work standalone and has no dependencies to pyfa things. Now, to apply the pyfa settings to eos, yes, you'll have to initialize the settings manager and inject them into eos.config, but that's not the same thing as saying eos is dependent on SettingsManager or anything pyfa-specific.

That is assuming it works the way I hope it does. Preliminary testing shows it does, but I want to fully integrate pref page with the new EOSSettings class. We shall see.

Ebag333:
> Now, to apply the pyfa settings to eos, yes, you'll have to initialize the settings manager and inject them into eos.config,

Which will mean we have two sets of settings again, like we do now. :)

Plus migrating things like the setting for including reload time.

It's basically going to be the exact same as what we have now, only instead of being attached to the fit class it'll be it's own class.  I'm just not seeing much improvement here.

blitzmann:
The point here is that the SettingsProvider will act as a proxy for the eos settings, simple as that. Instead of completely redoing our settings infrastructure to support this for no real benefit, this is a quick and easy way of doing it which a few quick tests have proved.

> It's basically going to be the exact same as what we have now, only instead of being attached to the fit class it'll be it's own class.

Not quite. All other setting "realms" have their own class for accessing and applying these settings, this shouldn't be any different when applied to pyfa. Of course, if you're running pure eos, you simply get a dictionary for the settings that you can modify yourself, which is ideal. 

Before (well, technically currently), the setting was saved in the fitting service, and then simply applied to the eos Fit class so that eos was aware of it and could take it into consideration during calcs. I've always hated how that worked. My solution designates a specific place for eos settings, which is handled within pyfa by the SettingsProvider without the fitting service being involved or having some sort of janky application. Along with the added benefit that eos is agnostic to everything else.

Also, the fact is that I aim to keep eos a self-contained entity, so it's going to be special in that it contains it's own settings. When developing new settings for the engine, just go to the eos config instead of the settings service. I don't see the problem with this - at least the setting definition isn't tucked away in an odd service somewhere, and we don't have two definitions floating around.

PR #1023, opened 2017-02-28
Quick Eos cleanup and logging fix

Ebag333:
Fixes case that would generate a exception on the first fit calc (would trigger logging dumps for users).  Also  clean up a bunch of unnecessary try/exceptions, and some logging improvements.

@blitzmann this is ready to merge.


blitzmann:
#yolo

Issue #1027, opened 2017-03-05
Unable to drag and drop module into first low slot


bjjerry:
Nothing happens when you attempt to drag and drop a module into the first low slot. To replicate this bug, fit a low slot module to the ship. Drag it from the first low slot to any other, then attempt to drag it back.

This happens because the first low slot's modPosition is 0, so the if block beginning on fittingView.py:409 is not correctly entered when it is the drag and drop target. Checking if the modPosition is not None seems to fix this issue.

I'd love any feedback on this issue, commit, pull request, or anything else you can offer as this is my first contribution to Pyfa. Thanks in advance.

Ebag333:
>I'd love any feedback on this issue, commit, pull request, or anything else you can offer as this is my first contribution to Pyfa. Thanks in advance.

Did you submit a PR? Don't see one.... 

Sounds legit. Would like to see your fix. :)

blitzmann:
Been in touch with @bjjerry on Slack, he's going to submit a PR. But indeed, this sounds like an issue of not checking for None explicitly. Nice catch!

bjjerry:
Sorry for the delay, went to grab some food and needed some lessons that @blitzmann was kind enough to provide. Hoping everything is now in order.

PR #1029, opened 2017-03-06
Fix for first low slot drag and drop

bjjerry:
Fixes issue #1027


blitzmann:
Excellent find and fix, thanks!

PR #1030, opened 2017-03-06
Add option to open fit in new tab

Will-W:
Just a quick usability update suggested by a corpmate.

Works from context menu and on middle click - like a web browser

### DIFF
--- a/gui/shipBrowser.py
+++ b/gui/shipBrowser.py
@@ -1552,6 +1552,10 @@ class FitItem(SFItem.SFBrowserItem):
         self.selTimer.Start(100)
 
         self.Bind(wx.EVT_RIGHT_UP, self.OnContextMenu)
+        self.Bind(wx.EVT_MIDDLE_UP, self.OpenNewTab)
+
+    def OpenNewTab(self, evt):
+        self.selectFit(newTab=True);
 
     def OnToggleBooster(self, event):
         sFit = Fit.getInstance()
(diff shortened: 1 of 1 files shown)


blitzmann:
I had thought middle click already worked. Nope, it was CTRL Click.

The only thing that I'm iffy on is this bit:

`wx.PostEvent(self.mainFrame, FitSelected(fitID=self.fitID, startup=2))`

The `startup` parameter really should only be used during startup, not during normal program execution. Normally I would recommended looking here:

https://github.com/pyfa-org/Pyfa/blob/master/gui/builtinViews/fittingView.py#L72

And trying to extract the middle click from the GetMouseState(), and then handling it there.

That being said, I've been meaning to look into how these events fundamentally work throughout pyfa and if they are named appropriately. The page handling is done in the FitSpawner class but, IMO, a lot of the things that it does should be handled via attributes on the events themselves (like open in new tab, etc).

I'll test this out tonight and merge ASAP. The review of the events will have to be done at a later date. :) Thanks for the contribution!

Will-W:
It just felt like a better separation of concerns that the thing handling the click event made the distinction rather than thing spawning fits. An example of the issue is handling of paste: if I do it from the menu, it opens the new fit in the same tab. If I do it using ctrl+v, because ctrl happens to be down, you get a new tab.

Using a parameter called 'startup' is admittedly a bit hacky, but the naming could be changed and it given more meaningful IDs (e.g. NORMAL, NEW_TAB_NO_CALCULATE, NEW_TAB).

That said, it is of course your call and I did depart from the established convention. Happy to rework it if you want.

blitzmann:
Nope, we're both in agreement with everything stated. It's not current convention, but in this case current convention is kinda hacky and I've been meaning to change it. No need for tweaking, this is fine as is as soon as I can test it. :)

Ebag333:
Code looks good? Check.
Nice little QoL improvement? Check.
Passes Travis? Check.
Slightly hacky? Check.

Looks like a good pyfa PR to me. 😁

PR #1031, opened 2017-03-07
Tooltip for capacitor neut resistance

Ebag333:
Just a small add to have a tooltip for capacitor neut resistance.  A very useful stat, but one not exposed at all effectively.

https://puu.sh/uykft/0076290861.gif

Also prevents exceptions from modifiedAttributeDict when getting an attribute that doesn't exist (returns default).  (Ran into this when passing in an object that doesn't exist because we're on a blank tab.)

Bit of Tox cleanup.

### DIFF
--- a/eos/config.py
+++ b/eos/config.py
@@ -16,4 +16,3 @@ settings = {
 
 # Autodetect path, only change if the autodetection bugs out.
 path = dirname(unicode(__file__, sys.getfilesystemencoding()))
-
--- a/eos/effects/tractorbeamcan.py
+++ b/eos/effects/tractorbeamcan.py
@@ -2,8 +2,9 @@
 #
 # Used by:
 # Modules from group: Tractor Beam (4 of 4)
-type = "active"
 from eos.config import settings
+type = "active"
+
 
 def handler(fit, module, context):
     print settings['setting1']
--- a/eos/modifiedAttributeDict.py
+++ b/eos/modifiedAttributeDict.py
@@ -26,17 +26,23 @@ cappingAttrKeyCache = {}
 
 class ItemAttrShortcut(object):
     def getModifiedItemAttr(self, key, default=None):
-        if key in self.itemModifiedAttributes:
-            return self.itemModifiedAttributes[key]
-        else:
+        try:
+            if key in self.itemModifiedAttributes:
+                return self.itemModifiedAttributes[key]
+            else:
+                return default
+        except AttributeError:
             return default
 
 
 class ChargeAttrShortcut(object):
     def getModifiedChargeAttr(self, key, default=None):
-        if key in self.chargeModifiedAttributes:
-            return self.chargeModifiedAttributes[key]
-        else:
+        try:
+            if key in self.chargeModifiedAttributes:
+                return self.chargeModifiedAttributes[key]
+            else:
+                return default
+        except AttributeError:
             return default
 
 
--- a/eos/modifiedAttributeDict.py
+++ b/eos/modifiedAttributeDict.py
@@ -283,7 +283,7 @@ class ModifiedAttributeDict(collections.MutableMapping):
         return skill.level
 
     def getAfflictions(self, key):
-        return self.__affectedBy[key] if key in self.__affectedBy else {}
+        return self.__affectedBy.get(key, {})
 
     def iterAfflictions(self):
         return self.__affectedBy.__iter__()
--- a/eos/modifiedAttributeDict.py
+++ b/eos/modifiedAttributeDict.py
@@ -131,7 +131,7 @@ class ModifiedAttributeDict(collections.MutableMapping):
 
     def getOriginal(self, key, default=None):
 
-        if self.overrides:
+        if self.overrides_enabled and self.overrides:
             val = self.overrides.get(key, None)
         else:
             val = None
--- a/eos/modifiedAttributeDict.py
+++ b/eos/modifiedAttributeDict.py
@@ -19,6 +19,7 @@
 
 import collections
 from math import exp
+from types import NoneType
 
 defaultValuesCache = {}
 cappingAttrKeyCache = {}
@@ -26,24 +27,22 @@ cappingAttrKeyCache = {}
 
 class ItemAttrShortcut(object):
     def getModifiedItemAttr(self, key, default=None):
-        try:
-            if key in self.itemModifiedAttributes:
-                return self.itemModifiedAttributes[key]
-            else:
-                return default
-        except AttributeError:
-            return default
+        return_value = self.itemModifiedAttributes.get(key)
+
+        if isinstance(return_value, NoneType) and not isinstance(default, NoneType):
+            return_value = default
+
+        return return_value
 
 
 class ChargeAttrShortcut(object):
     def getModifiedChargeAttr(self, key, default=None):
-        try:
-            if key in self.chargeModifiedAttributes:
-                return self.chargeModifiedAttributes[key]
-            else:
-                return default
-        except AttributeError:
-            return default
+        return_value = self.chargeModifiedAttributes.get(key)
+
+        if isinstance(return_value, NoneType) and not isinstance(default, NoneType):
+            return_value = default
+
+        return return_value
 
 
 class ModifiedAttributeDict(collections.MutableMapping):
@@ -102,17 +101,47 @@ class ModifiedAttributeDict(collections.MutableMapping):
     def overrides(self, val):
         self.__overrides = val
 
+    @property
+    def modified(self):
+        return self.__modified
+
+    @modified.setter
+    def modified(self, val):
+        self.__modified = val
+
     def __getitem__(self, key):
+        '''
         # Check if we have final calculated value
-        if key in self.__modified:
-            if self.__modified[key] == self.CalculationPlaceholder:
-                self.__modified[key] = self.__calculateValue(key)
-            return self.__modified[key]
+        if key in self.modified:
+            if self.modified[key] == self.CalculationPlaceholder:
+                self.modified[key] = self.__calculateValue(key)
+            return self.modified[key]
         # Then in values which are not yet calculated
         elif key in self.__intermediary:
             return self.__intermediary[key]
         # Original value is the least priority
         else:
+            return self.getOriginal(key)
+        '''
+
+        # Check if we have final calculated value
+        key_value = self.modified.get(key)
+        if key_value is self.CalculationPlaceholder:
+            key_value = self.modified[key] = self.__calculateValue(key)
+
+        if not isinstance(key_value, NoneType):
+            return key_value
+
+        # Then in values which are not yet calculated
+        if self.__intermediary:
+            val = self.__intermediary.get(key)
+        else:
+            val = None
+
+        if not isinstance(val, NoneType):
+            return val
+        else:
+            # Original value is the least priority
             return self.getOriginal(key)
 
     def __delitem__(self, key):
@@ -122,11 +151,18 @@ class ModifiedAttributeDict(collections.MutableMapping):
             del self.__intermediary[key]
 
     def getOriginal(self, key):
-        if self.OVERRIDES and key in self.__overrides:
-            return self.__overrides.get(key).value
-        val = self.__original.get(key)
-        if val is None:
-            return None
+        val = None
+
+        if self.__overrides:
+            val = self.__overrides.get(key, None)
+        elif self.overrides:
+            val = self.overrides.get(key, None)
+
+        if isinstance(val, NoneType):
+            if self.__original:
+                val = self.__original.get(key, None)
+            elif self.original:
+                val = self.original.get(key, None)
 
         return val.value if hasattr(val, "value") else val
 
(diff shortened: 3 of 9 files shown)


blitzmann:
Looks good

> Also prevents exceptions from modifiedAttributeDict when getting an attribute that doesn't exist (returns default). (Ran into this when passing in an object that doesn't exist because we're on a blank tab.)

How can I reproduce? I don't get an exception when removing these try/catchs and loading up a blank fit. Provide an example? (I'm assuming this is caught when `itemModifiedAttributes` doesn't exist on `self`, which should never really happen IIRC)

I thing we've discussed this particular code before - I'd rather not do a try catch here - instead I would want whatever is calling it to be fixed as it shouldn't call it. But I can't reproduce the situation you had... :(

Ebag333:
>I thing we've discussed this particular code before - I'd rather not do a try catch here - instead I would want whatever is calling it to be fixed as it shouldn't call it. But I can't reproduce the situation you had... :(

We did for the _other_ set of get attributes, and we did indeed end up including the try/catch.

I was fiddling around with it, and was getting it when trying to call it directly.  Of course the _best_ solution is to do what  I ended up doing:
```
        if fit is None:
            # Set default if fit is empty
            neut_resist = 0
         else:
```
and look before we leap, because not only is there no `getModifiedItemAttr`, there's no `ship` at all.

We are only passing back the default when we  hit an exception of `AttributeError` which is going to be a missing attribute.  We won't hit it in any of our normal code, because we're properly gating if fit exists (since we have to have a ship), so yes this is a case of "lets be cautious".

We're not going to be hitting this with any of our current code, because all the attributes we're trying to get _should_ always be there.  But it is something I'd like to do as it makes it more Pythonic, it's how `getattr` is handled, and it'd mean we could rewrite a lot of the `(fit.ship.getModifiedItemAttr("agility") or 0)` statements as `fit.ship.getModifiedItemAttr("agility", 0)` which is both easier to read and follows `getattr` standards.

It's mostly for protection from a future "we could be trying to get a stat that doesn't always exist" scenario, which is entirely plausible but not something we'd hit currently.

blitzmann:
I guess I'm getting confused by the terminology... when you say attribute, you mean an EVE attribute or a python object attribute?

So instead of `energyWarfareResistance` you accidentally typed in `energyWarfareResistanceWhoopsThisDoesntExist`, and hit an exception? Is this not covered with the `if key in self.itemModifiedAttributes:` conditionals? This is why I asked for an example to reproduce, because I'm not sure what the `except AttributeError` is specifically designed for (I am assuming this is actually coming from the `ModifiedAttributeDict.__getitem__`)

>  But it is something I'd like to do as it makes it more Pythonic, it's how getattr is handled, and it'd mean we could rewrite a lot of the (fit.ship.getModifiedItemAttr("agility") or 0) statements as fit.ship.getModifiedItemAttr("agility", 0) which is both easier to read and follows getattr standards.

I'm not arguing against the default, I'm just not understanding what the the try catch is trying to achieve other than a catch all in case we try to call an attribute that doesn't exist (which I'm on the fence of including - we shouldn't be trying to call an attribute that doesn't exist, and even if we were it should be caught bu the `key in itemModifiedAttirbutes` bit)

I would like to reproduce what you were reproducing to understand it a bit more. Thanks :)

Ebag333:
>I'm not sure what the except AttributeError is specifically designed for (I am assuming this is actually coming from the ModifiedAttributeDict.__getitem__)

It comes from `builtins`.

```
exception AttributeError
Raised when an attribute reference (see Attribute references) or assignment fails. (When an object does not support attribute references or attribute assignments at all, TypeError is raised.)
```

Looking at this, though, I think there's a better/easier way.

Here's a  little chart with how this behaves.

Current Code:
Valid Key, No Default    - Returns Value
Valid Key, Default          - Returns Value
Invalid Key, No Default - Returns None
Invalid Key, Default       - Returns Default

`.get(key, default)`
Valid Key, No Default    - Returns Value
Valid Key, Default          - Returns Value
Invalid Key, No Default - Returns None
Invalid Key, Default       - Returns None

Now, I don't know why it's returning None instead of the default.  But we can easily handle that....

So we can turn this block of code:
```
if key in self.itemModifiedAttributes:
    return self.itemModifiedAttributes[key]
else:
    return default
```

into this block:
```
return_value = self.itemModifiedAttributes.get(key, default)
if return_value != default:
     return_value = default

return return_value
```

Okay, so lines of code wise that doesn't look that impressive, but if I can figure out why `get` isn't returning the default (as it's documented it should...) then we can eliminate two lines of code, cutting it in half (which is a small time savings on a single lookup, but should be significant overall, given how many times we  will hit this bit of code).

More importantly, this requires only _one_ lookup against the dict, where as the original requires _two_.  And since `get` already handles all the errors for you, should eliminate all of our cases for using any try/except logic.

Also, it's more pythonic.  Because, you know, we're using built in functionality and not rolling our own. :)

Ebag333:
>I don't know why it's returning None instead of the default.

I think it's because we have modified a dict.  It's a `ModifedAttributeDict` rather than a standard dict.

Either way, it's still faster to do a quick compare than a second dict lookup.

Ebag333:
So a few things.

First, this isn't as big of a speed boost as I'd hoped.  It does seem to be faster, but running 3 recalcs when in debug and info logging, only saved me around 0.15 seconds.  Not huge.

_But_ I did find a few problems with the existing code.

```
    @property
    def modified(self):
        return self.__modified

    @modified.setter
    def modified(self, val):
        self.__modified = val
```

We didn't have a modified property, so that was just weird.  The code worked, but trying to set a breakpoint and replicate it ad hoc would fail.  Made troubleshooting...weird....

```
    def getOriginal(self, key):
        val = None

        if self.__overrides:
            val = self.__overrides.get(key, None)
        elif self.overrides:
            val = self.overrides.get(key, None)

        if isinstance(val, NoneType):
            if self.__original:
                val = self.__original.get(key, None)
            elif self.original:
                val = self.original.get(key, None)
```

This was another weird one. Set a breakpoint in here (on the original code) and `__overrides` would stacktrace when trying to access it adhoc....because it didn't exist, but `self.overrides` did.

So, seems that sometimes we pass in one or the other.  Because this is hidden behind some properties, the stack trace doesn't bubble up when we try and access one that's missing.

I'll be honest, I didn't check to see if `self.__original` had the same problem, but figured if one could do it so could the other.

Because these are several layers deep, it seems that stacktraces aren't bubbling up.  I suspect anything hidden behind a `@property` is going to have that problem. I know you don't want try\excepts everywhere @blitzmann, but I think that this is a good candidate for it

Ebag333:
>Refactor __calculateValue

I went ahead and applied the same changes inside this method.

Little surprised at the difference in time.  Averages out to .1s less time each recalc.  Okay, it's not a huge savings, but it's a good chunk.

blitzmann:
As a curiosity, why do `isinstance(thing, NoneType)` rather than `thing is None`. I have never really seen the former syntax before

> It comes from builtins.

I know it's a build-in Exception. My question was literally where and how was it being raised in pyfa code. File and line number, and a brief investigation as to why (what variables was it using, etc). Which is why I've been asking for an example. You did not produce a way to reproduce this like I've requested.  This makes me sad. See: 😢 

> We didn't have a modified property, so that was just weird. 

That's because we didn't need a modified property So, in general, `self.__modified` starts with underscores because it's an internal thing - we don't really need to access this outside of the class, hence we we never had a getter or setter for it. Things that are defined as `self.attribute` are assumed to be accessible outside the class. This is a public/private thing that is common in python applications (ever notice in the debugger things like `_Fit__myProperty` when it's defined as `self.__myProperty`? The python interpreter does this automatically for properties that start with `__` and indicates that while you can still access it outside the class, you shouldn't).

So I'm not sure why you created a property for it when it is only accessed internally?

> This was another weird one. Set a breakpoint in here (on the original code) and __overrides would stacktrace when trying to access it adhoc....because it didn't exist, but self.overrides did.

Is this the exception you were getting?

image

This seems to be a pycharm / debugger issue. Why it happens I don't know, but it doesn't happen when running normally (and only happens for me when "expanding" the variable - I don't get an error in the debugger variable window itself). While this is weird, I don't know if it's indicative of anything serious. `__overrides` does exist so I'm not sure what you mean by it doesn't exist. Screenshots would help if this isn't the issue you were describing.

> More importantly, this requires only one lookup against the dict, where as the original requires two. Now this is a small time savings on a single lookup, but should be significant overall, given how many times we will hit this bit of code. And since get already handles all the errors for you, should eliminate all of our cases for using any try/except logic.

I can get behind this and some of the other changes here, but honestly this is now touching core calculation code that is out of the scope of this PR. I must reiterate my desire to have a PR with one topic instead of tacking on additional things like this.

It can remain for now, don't worry about separating it out into a different PR, but it will take a bit to review these calculation changes and make sure everything works okay (especially since it seems we're changing default returns and what not, which can have an adverse effect on some code throughout the application). If I have any questions I'll leave a note on the line itself.

Thanks

Ebag333:
>As a curiosity, why do isinstance(thing, NoneType) rather than thing is None. I have never really seen the former syntax before

I was looking at various methods to speed things up, since this code is accessed about a bajillion times.  There was some theorizing that comparing the `Type` of an object was faster than comparing the object itself, which makes a lot of sense.

I actually tested a bit of code that looped a million times and did a `IsInstance()` vs `is None`.  The `is None` ended up being .1 seconds faster (over the million cycles), so I have a commit to submit to switch back to using that (just as soon as I reinstall Git).

In most cases I think `IsInstance` _would_ actually be faster, especially when comparing complex objects, but for this particular case, not so much.

>Which is why I've been asking for an example. You did not produce a way to reproduce this like I've requested. This makes me sad. See: 😢

I know. I've been busy, am _way_ past where I was at when I was getting the error, and recreating it would take a bunch of work.

I _really_ need to start screen shotting when these things actually happen.

>So, in general, self.__modified starts with underscores because it's an internal thing - we don't really need to access this outside of the class, hence we we never had a getter or setter for it. Things that are defined as self.attribute are assumed to be accessible outside the class. This is a public/private thing that is common in python applications (ever notice in the debugger things like _Fit__myProperty when it's defined as self.__myProperty? The python interpreter does this automatically for properties that start with __ and indicates that while you can still access it outside the class, you shouldn't).

I think you're saying the same thing as below, but I suspect you putting more on the "private" part than actually is there.

>But try to avoid the __private form. I never use it. Trust me. If you use it, you WILL regret it later.
>Explanation:
>People coming from a C++/Java background are especially prone to overusing/misusing this "feature". But __private names don't work the same way as in Java or C++. They just trigger a name mangling whose purpose is to prevent accidental namespace collisions in subclasses: MyClass.__private just becomes MyClass._MyClass__private. (Note that even this breaks down for subclasses with the same name as the superclass, e.g. subclasses in different modules.) It is possible to access __private names from outside their class, just inconvenient and fragile (it adds a dependency on the exact name of the superclass).
http://stackoverflow.com/questions/6930144/underscore-vs-double-underscore-with-variables-and-methods

You might be saying the same thing, just in a different way.

Regardless, it makes low level troubleshooting much harder, because we get the error when trying to access via pyCharm using `self.__modified`:
`{AttributeError}'ModifiedAttributeDict' object has no attribute '__modified'`
Yes, I know you can access using `self._ModifiedAttributeDict__modified` but seriously that's a knuckle breaker.

Plus, we have properties for the other objects, so keeping them all the same keeps them much simpler since they all act/behave the same.

>I can get behind this and some of the other changes here, but honestly this is now touching core calculation code that is out of the scope of this PR. I must reiterate my desire to have a PR with one topic instead of tacking on additional things like this.

I know, it grew on me. I shouldn't ever feed code after midnight.  (Or was it water? Dang, I'm screwed.)

I can break it out fairly easily if you'd like.

blitzmann:
> I really need to start screen shotting when these things actually happen.

I would appreciate it, would help a lot. As a general rule, if we're changing something (rather than adding something), we should detail why we're changing it (with the fit, the exact error / screenshot, a bit of verbosity into why it happens by digging down during debug to find where it actually happens, etc). Would be a great help in reviewing the change (and for documentation).

> Regardless, it makes low level troubleshooting much harder, because we get the error when trying to access via pyCharm using self.__modified:
{AttributeError}'ModifiedAttributeDict' object has no attribute '__modified'
Yes, I know you can access using self._ModifiedAttributeDict__modified but seriously that's a knuckle breaker.

It's common convention in Python defined in PEP8, as well as convention within pyfa. I would prefer that we still maintain it wherever it is to signify something that is internal and isn't meant to be exposed. This is a more practical convention rather than syntatical, and is something I do actively keep in mind when I'm developing - I know at a glance if a variable is private or not by it's name. If I'm working on `self.__modified`, I know that I do not have to worry about other places using it). 

Indeed, it seems pyCharm gets a bit pussy about it, but such is life. It doesn't matter unless we're actively debugging in that class, in which case it only fails when we expand the variable as shown in the screenshot (and now we both know why it happens - it's due to the private variable). I will say that this is new information for me as well - I wasn't aware pycharm broke when trying to evaluate the expression of a "private" variables. This may very well be a bug with pyCharm itself when handling private variable expansion of anything that you can drill down with

> I can break it out fairly easily if you'd like.

As stated, don't worry about it for this one unless it becomes a bit of a hassle during review this week. Just keep it in mind.

Ebag333:
@blitzmann I cleaned up the code quite a bit, which involved reverting a few things as discussed above.

I think that this is in a pretty good place now, unless you see anything glaring.

Replacing:
```
       val = self.__intermediary[key] if key in self.__intermediary else self.__preAssigns[
            key] if key in self.__preAssigns else self.getOriginal(key) if key in self.__original else default
```
with
```
        val = self.__intermediary.get(key,
                                      self.__preAssigns.get(key,
                                                            self.getOriginal(key, default)
                                                            )
                                      )
```

makes me very 😄

Issue #678, opened 2016-07-09
[Feature Request] Prices on implants and boosters.


Ligraph:
It would be amazing to be able to see prices like on modules/drones.

blitzmann:
Should be as easy as adding that column to the display. Will take a look, thanks!

blitzmann:
Took a look at this, and it's not as simple as I thought, which is probably why they were not included to begin with. I would still like to fix this eventually though.

Techy version:

The price column checks to see if the item has a valid price (isn't outdated). If it checks out, it simply displays it. If not, it stores the item's ID with the price fetching thread with a callback that updates the GUI text for the price.

Now, the actual code that gets the price is found in the price stats pane. It compiles a list of modules and send it to the price fetcher thread. It then fetches the price, and tries to find the item in the previously stored items and calls the callback.

The problem is that the implants and boosters are not compiled with the rest of the fits price, and thus the prices never get updated.

This is pretty bad design. I might be able to fiddle with it so that the wait list is what is process when the fetcher is triggered, rather than supplying a list of items to fetch. We shall see...

Ebag333:
See #1038 for a partial solution.

It doesn't add a price column.  _Yet._

blitzmann:
Fixed with #1115

PR #1038, opened 2017-03-14
Rework price fetching for items

Ebag333:
This significantly simplifies fetching prices for items.  

We add a @property named `price` on all items.  This fetches directly from the DB rather than caching it. Even a very complex fit (think full set of implants, cargo, etc) only takes around .1 seconds to full parse all the pricing.  A more standard/simple fit takes .05 seconds.  That's in full potato mode, so when compiled it should be basically nil time.

The `price` property is the full price object, so it has the timestamp, `isValid`, etc.
(This does make querying it slightly awkward, since it's `item.price.price` to get the value. Meh.)

Add new pricing full view (move existing into a minimal version).  The pricing view (both min and full) should now reflect all the tabs correctly, including amounts.

The minimal version looks essentially the same.  For the values, ship is the ship cost, and fitting is the cost of everything else (modules, drones/fighters, cargo, and implants/boosters).

For the full version, we're grouping drones and fighters together (under the "Drone" label"), and implants and boosters together (under the Character label).  I'm not terribly worried about the drone label including fighters, since half the carrier pilots I know call fighters "drones" anyway.  Character is _slightly_ more confusing, but since implants and boosters both apply to the character (rather than the ship) I think it's easy enough to figure out.

One final little thing I slipped in here, raising an error if we exception in `pyfa.py` rather than exiting, so we don't lose the stacktrace in the IDE.  `raise` will exit out so we're good either way.

.

Okay, so now for the "problem".  If we have prices cached, it's essentially the same time to load the fit as we have now.  If we _don't_ then we're blocking the process as we fetch the items, and since we're fetching one at a time it's not as fast at bulk operations (just a handful won't be a significant difference).

In testing, querying for 24 items (when none of them are cached in the DB) took just under 12 seconds.  Now that's in full potato mode so won't be what an end user will see, but still not great.

What I'd like to do is create a simple background service that updates prices in the background.  Simply put, loop through the items in the DB, and update them on a mass scale.  We can easily query for the price of dozens or hundreds of items in one shot, and it'd be trivial to loop through all the items and update them (if they're expired).

I think that this is ultimately a better solution than what we had previously, which was fairly complicated (requiring callbacks).  While ultimately this solution would be slightly slower in one particular use case than the current existing solution (since if a user loaded a fit with null/expired values very quickly, it'd take some time to update), once the application has been open for a while it would actually be quite a bit faster, because theoretically you'd never run into any items without a price.  It'd also help users like myself, where I'm using Pyfa fairly often in an offline mode.  You'd have the values for those items, where as currently you get bupkis.

There is the one drawback with this PR (once I add on the service to fetch prices), but I think that the huge amount of simplification and constantly background updating prices is worth the trade off.

### DIFF
--- a/eos/db/gamedata/queries.py
+++ b/eos/db/gamedata/queries.py
@@ -98,6 +98,11 @@ def getItem(lookfor, eager=None):
     return item
 
 
+def getAllItems():
+    items = gamedata_session.query(Item).all()
+    return items
+
+
 @cachedQuery(1, "lookfor")
 def getAlphaClone(lookfor, eager=None):
     if isinstance(lookfor, int):
--- a/eos/gamedata.py
+++ b/eos/gamedata.py
@@ -23,6 +23,7 @@ from sqlalchemy.orm import reconstructor
 
 import eos.db
 from eqBase import EqBase
+from eos.saveddata.price import Price as types_Price
 
 try:
     from collections import OrderedDict
@@ -418,6 +419,31 @@ class Item(EqBase):
                 return True
 
         return False
+
+    @property
+    def price(self):
+        try:
+            if not hasattr(self, "__price"):
+                self.__price = types_Price(self.ID)
+
+            price = eos.db.getPrice(self.ID)
+
+            if self.__price != price and price is not None:
+                self.__price = price
+
+            eos.db.add(self.__price)
+            eos.db.commit()
+
+            return self.__price
+        except Exception as e:
+            # We want to catch our failure and log it, but don't bail out for a single missing price tag.
+            pyfalog.error("Failed to get price for typeID: {0}", self.ID)
+            pyfalog.error(e)
+            return 0
+
+    @price.setter
+    def price(self, val):
+        self.__price = val
 
     def __repr__(self):
         return "Item(ID={}, name={}) at {}".format(
--- a/gui/boosterView.py
+++ b/gui/boosterView.py
@@ -43,9 +43,12 @@ class BoosterViewDrop(wx.PyDropTarget):
 
 
 class BoosterView(d.Display):
-    DEFAULT_COLS = ["State",
-                    "attr:boosterness",
-                    "Base Name"]
+    DEFAULT_COLS = [
+        "State",
+        "attr:boosterness",
+        "Base Name",
+        "Price",
+    ]
 
     def __init__(self, parent):
         d.Display.__init__(self, parent, style=wx.LC_SINGLE_SEL | wx.BORDER_NONE)
--- a/gui/builtinContextMenus/priceClear.py
+++ b/gui/builtinContextMenus/priceClear.py
@@ -3,7 +3,7 @@ import gui.mainFrame
 # noinspection PyPackageRequirements
 import wx
 import gui.globalEvents as GE
-from service.market import Market
+from service.price import Price
 from service.settings import ContextMenuSettings
 
 
@@ -22,8 +22,8 @@ class PriceClear(ContextMenu):
         return "Reset Price Cache"
 
     def activate(self, fullContext, selection, i):
-        sMkt = Market.getInstance()
-        sMkt.clearPriceCache()
+        sPrice = Price.getInstance()
+        sPrice.clearPriceCache()
         wx.PostEvent(self.mainFrame, GE.FitChanged(fitID=self.mainFrame.getActiveFit()))
 
 
--- a/gui/builtinPreferenceViews/pyfaGeneralPreferences.py
+++ b/gui/builtinPreferenceViews/pyfaGeneralPreferences.py
@@ -9,7 +9,6 @@ import gui.globalEvents as GE
 from service.settings import SettingsProvider
 from service.fit import Fit
 from service.price import Price
-from service.market import Market
 
 
 class PFGeneralPref(PreferenceView):
@@ -188,8 +187,8 @@ class PFGeneralPref(PreferenceView):
 
         fitID = self.mainFrame.getActiveFit()
 
-        sMkt = Market.getInstance()
-        sMkt.clearPriceCache()
+        sPrice = Price.getInstance()
+        sPrice.clearPriceCache()
 
         self.sFit.refreshFit(fitID)
         wx.PostEvent(self.mainFrame, GE.FitChanged(fitID=fitID))
--- a/gui/builtinPreferenceViews/pyfaStatViewPreferences.py
+++ b/gui/builtinPreferenceViews/pyfaStatViewPreferences.py
@@ -91,8 +91,6 @@ class PFStatViewPref(PreferenceView):
         rbSizerRow3 = wx.BoxSizer(wx.HORIZONTAL)
 
         self.rbPrice = wx.RadioBox(panel, -1, "Price", wx.DefaultPosition, wx.DefaultSize, ['None', 'Minimal', 'Full'], 1, wx.RA_SPECIFY_COLS)
-        # Disable minimal as we don't have a view for this yet
-        self.rbPrice.EnableItem(1, False)
         self.rbPrice.SetSelection(self.settings.get('price'))
         rbSizerRow3.Add(self.rbPrice, 1, wx.TOP | wx.RIGHT, 5)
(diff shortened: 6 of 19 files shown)


Ebag333:
Added price column to address #678 

https://puu.sh/uJ67j/4872380c5f.png

Ebag333:
>There is the one drawback with this PR (once I add on the service to fetch prices), but I think that the huge amount of simplification and constantly background updating prices is worth the trade off.

The service has been implemented and is working well.  Whee for background update threads!

So the way that this works is we can get pricing one of two different ways.

1) On demand.  If the price doesn't exist, failed, or is invalid (expired), then we're going to fetch it one at a time from EVE-Central.  For most users (if you've had Pyfa open for more than a few minutes), very few items should hit this.
2) Background thread.  This cycles through a list of items and updates them 333 at a time.  (That's an arbitrary number, eventually should expose that with an option).  The thread then sleeps for 60 seconds, then loops back around.

This means that the entire database should be updated in ~15 minutes.  This is a _very_ slow update, since it's less than one request a minute.  We could easily shorten the time and increase the number to update in a go if we want it done faster.

The Price thread isn't kicked off until the first time something calls for a price (initializes the Price class).  This is intentional, it helps spread out our thread starts a bit so they're not all starting at once, and I didn't want to potentially slow down Pyfa launching at all.  We _could_ initialize it early on if we desire, but I'd recommend against it.  The one potential downside is that if someone just launches Pyfa, then it's not going to do any background updates until the first price is queried (open fit or open compare window on item/ship).


Overall I think that this makes the whole process much simpler.  It completely divorces the `Market` service from the `Price` service, so they're no longer interdependent, and pricing is much more tightly encapsulated.

@blitzmann when you're done with the latest build, let me know what you think.  I might fiddle around with cleaning it up a bit, but basically it's ready to go.

blitzmann:
> What I'd like to do is create a simple background service that updates prices in the background. Simply put, loop through the items in the DB, and update them on a mass scale. We can easily query for the price of dozens or hundreds of items in one shot, and it'd be trivial to loop through all the items and update them (if they're expired).

Sorry bro, but I'm not on board with this. Price fetching should be a background job, yes, but it should be an on demand thing. I'm not kosher with pyfa sitting there chugging through and continuously making HTTP requests to fetch pricing data. I'm also not okay with waiting until we get the price of an item for pyfa for the fit to even load - this is bad design, even if the database will eventually catch up (and even when caught up, if the price is invalid, there is that delay to load the fit until we get the price).

> You'd have the values for those items, where as currently you get bupkis.

This shouldn't be true - if we fail a price update due to no connection, then we retry, and if that fails, we use whatever is in the cache.

Practically, from a quick test, I found these issue: 

* When trying to load a fit when a price update was presumably in progress: `sqlalchemy.exc.InvalidRequestError: This session is in 'prepared' state; no further SQL can be emitted within this transaction.`
* Loading fits, even if they have had their prices updates, seems to take longer than master and causes GUI flickering. Which is really odd... I have no idea why any of this would do that if the price has been fetched and updated unless something is blocking the thread. Can provide a gif if requested.
* When switching between fits (prices should have already been loaded) there are instances of `sqlalchemy.exc.InvalidRequestError: This Session's transaction has been rolled back due to a previous exception during flush.`
* When closing pyfa: 
```Exception in thread PriceWorker (most likely raised during interpreter shutdown):
  File "C:\Users\Ryan\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\threading.py", line 810, in __bootstrap_inner
  File "C:\Users\Ryan\Sync\Git\blitzmann\Pyfa\service\price.py", line 261, in run
  File "C:\Users\Ryan\Sync\Git\blitzmann\Pyfa\eos\gamedata.py", line 288, in requiredSkills
<type 'exceptions.TypeError'>: 'NoneType' object is not callable
```

Price fetching should definitely be an on-demand thing, and preferably with the option to in bulk as we currently do. Honestly, I think we should stick to our current system. It works, and pretty well I think, even if it's not very straightforward and a PITA to work with. There are things that can be improved upon (for example utilizing a `price` property for all Items that we can attach the price object to) but I believe that the basic infrastructure for it should remain the same.

Ebag333:
>pyfa sitting there chugging through and continuously making HTTP requests to fetch pricing data

It's not.  If we bumped up the number of items requested to 500 per run, it's going to be less than 10 HTTP requests in a 24 hour period.

> even if the database will eventually catch up

There's no real delay here, even in full potato mode.

>(and even when caught up, if the price is invalid, there is that delay to load the fit until we get the price).

Yeah. There's a delay here, but should be fairly small.

>>You'd have the values for those items, where as currently you get bupkis.
>This shouldn't be true - if we fail a price update due to no connection, then we retry, and if that fails, we use whatever is in the cache.

You misunderstand.  If I can't query, and there is no cache, then I have nothing.

>Practically, from a quick test, I found these issue:

I'll poke at these and see what I can find.

>Price fetching should definitely be an on-demand thing, and preferably with the option to in bulk as we currently do. Honestly, I think we should stick to our current system. It works, and pretty well I think, even if it's not very straightforward and a PITA to work with. There are things that can be improved upon (for example utilizing a price property for all Items that we can attach the price object to) but I believe that the basic infrastructure for it should remain the same.

`not very straightforward and a PITA to work with` is a bit of an understatement.  It's way more complex than it needs to be, IMHO, and while yes it's doing it's updates in the background that just forces multiple GUI refreshes when the prices come in, which slows things down in it's own way.  Granted that's not _as_ blocking, but wx still doesn't respond well when it's going through a refresh (which very well might be because of everything we're doing, rather than the GUI itself).

I don't think that this is a perfect solution, we're trading one set of slowness for another set, but it's vastly simpler and a lot more powerful. You have some good points with the issues, but I don't see anything that can't be addressed.

blitzmann:
Thought about this some more today, and played around with the branch a bit as well. I'm still not on board, and don't think I will be, for the reasons listed below.

I strongly feel it's simply bad design to block execution of the thread until we get the price of an item, which is about #49 on a list of top 50 most important things in pyfa. Not only that, but the other problems that this branch introduces (session issues and the like)... it's not worth it in my opinion.

> background that just forces multiple GUI refreshes when the prices come in, which slows things down in it's own way.

I mean, this is the way it should be. GUI refreshes when it's needed. It also doesn't do a full GUI refresh, it only sets that cells data. And regardless, since it's threaded, it's transparent to the user. This is like saying a progress bar is inadequate because it requires the GUI to refresh to reflect the status while the background thread works, vs doing the background thread in the foreground and having everything freeze. :/

> Yeah. There's a delay here, but should be fairly small.

For your internet connection, perhaps. For someone rural who doesn't have access to better internet (or someone traveling, for instance), this might not be the case. Additionally, there are situations that we can't control, such as routing issues, eve-central server response times, timeouts, etc. Imagine Amazon fucking up the internet again and it takes 30 seconds for the call to timeout. Imagine a company firewall which blocks the API requests, which depending on firewall config the call might not return immediately  (not that I advocate pyfa'ing during work hours :P). And even if you have decent internet, loading that first fit after a week break from pyfa, we have to go out and fetch all the items on that fit before we show it, regardless if we already have "old" data.  

There's a lot of reasons we shouldn't make the fitting window dependent on an API request that may or may not complete at some point in the future. HTTP requests should almost always be async programming, unless you're specifically waiting on that data.

Having said all that, I can definitely see the appeal to having a background thread to "pre-fetch" as many items as it can, for those folks that work in offline mode a lot, simply so that you have some sort of real data. Also, thanks for clearing it up that it's not continuously working and only happens every 24 hrs. Such a thing should have an option in the settings to start at startup (settings ftw), and utilize the current price service (fetch a list from the database, feed it through the price service, save the prices, repeat until database is stocked, much like what this PR does).

So basically, I think this PR, while aimed at simplifying how our Pricing system works, goes too far in that simplification and introduces disadvantages with no real benefit other than it's now simple and straightforward.

What I think would be a good PR:
* `price` property for items that carries around it's Price object, which would make it easier
* Sending a list of items, not item id's, to the pricing service, and then populating their `Price` object as we get data back
* Optional background task that pre-fetches pricing data
* I think I'm okay with the full / minimal pricing panel (thanks for clarifying it somewhere else)

What I don't like and will not include:
* Basic pyfa functionality / speed determined by HTTP calls

PR #1042, opened 2017-03-15
Effect Handlers Import - Less Spam

Ebag333:
Simple change to not try and reimport handlers if they have failed previously.

Mostly this just cleans up the spam, but probably does make processing _slightly_ faster.


blitzmann:
Here's a thought: every effect is a new instance of `Effect` for that item. So, for example, Web I and Web 2 each have the `remotewebifierfalloff` effect. However, they are two separate instances of what is essentially the same thing:

<eos.gamedata.Effect object at 0x0BADC590>
<eos.gamedata.Effect object at 0x0BAF5450>

This makes me wonder... why not just have a single effect object for this that multiple items can use? The effect doesn't change from item to item - it's basically a carrier of a `handler()` method, so there shouldn't any situation in which the Effect class needs to carry specific information for an item. 

I have no idea where to begin with this as it's been a while since I looked at how SQLAlchemy caches object, but it might reduce memory footprint and increase performance...

Aside form that, this bit looks good and should help. Only thing I would change is to make it a `set` and change it to a class variable. The pros to a class variable is that you avoid the `global`, and it can also be accessed by any instance of the class via `self.__class__.variable` :)

Issue #1043, opened 2017-03-16
Implant sets right click context menu no longer opening


bjjerry:
In pyfa 1.28.0 under windows 7 I am no longer able to right click in a fit's Implants tab to load a saved implants set. If memory serves, last version did not have the Variations selection in the right click context menu when selecting an implant that has been added to a fit's Implants, so maybe this bug is related to that.

blitzmann:
@Ebag333 possibly something with the new context menu settings or the variations thing we added?

Ebag333:
```
[2017-03-16 06:18:51.333000] WARNING: service.settings: Failed to get setting for 'itemRemove'. Exception: 'itemRemove'
[2017-03-16 06:18:51.335000] WARNING: service.settings: Failed to get setting for 'implantSets'. Exception: 'implantSets'
```

I've noticed this popping up.

Ebag333:
PR submitted with a fix.  Thanks @bjjerry for reporting this!

PR #1045, opened 2017-03-16
Error Dialog improvements, Pyfa.py refactoring, and misc fixes

Ebag333:
This covers a bunch of little things.

Fix all the PEP8 errors that forcefully pushed themselves in through a back door.

Fixes the issue reported in #1043.  That one was my bad, missed a setting and another was named incorrectly.

Reworks the error dialog so the layout is a bit cleaner, and we get more information.  Hopefully users will copy/paste, so this tells us more about their environment.
https://puu.sh/uLgXZ/3a3abef8d4.png

The big change in here is Pyfa.py.  Functionally it's the same, but there's been a lot of low hanging fruit in here that has irritated me for a long time.  In no particular order:
1. Move a bunch of imports to the top (basically everything that _can_ go to the top does).
2. Wrap the ones we expect might fail (wxPython and sqlAlchemy) in a try/catch so we don't fail at the top.  This lets us fail later, when we can show a pretty error dialog and log properly to the log file.
3. Rework the detection logic for wxPython. For some reason, the function we were using has some hard requirements for when you use it, instead just test for what version we have installed.  Also, adds a bit of logging to show what version is being used, useful for devs or if stack traces happen and logs dump.
4. Instead of exiting when we hit a problem (see 2 or 3 above), log it and then raise an exception.  This means that you get the nice error handling and not just a dump to `stdout()`.
5. Encapsulate (nearly) everything in a try/catch for the error handling.  Basically if something fails in `pyfa.py`, we're going to get the nice error dialog.  The one big exception is the main loop, which can't be embedded in there.
6. Adds a todo at the bottom.  We currently don't really do anything when Pyfa ends, even though we have a bunch of sub-threads.  We _should_ be cleaning up after ourselves.  This is a place holder for a future change.


blitzmann:
Please decouple the different things - the fix for that issue, pep8 fixes, and pyfa.py refactoring. Just looking at this it's hard to judge where the issue fix was, etc. 

It all looks fine on the surface, but I really want to break away from PRs that introduce things upon things upon things. These are three distinct changes, they should be in three distinct PRs. As a general rule, if the thought "Maybe I should sneak this in there while I'm in this file" pops into your head, that's probably an indication that it should go in another PR. Focus on the topics at hand :)

Additionally, that bug fix should be against master if at all possible. (I know I haven't commented back on the Gist yet regarding this - I had a comment prepared but then things happened and I forgot to post it, will comment on it sometime tonight)

Thanks!

Ebag333:
>These are three distinct changes, they should be in three distinct PRs. 

This is fair, but one point.

Anything that fails Travis should be fixed, regardless if it has anything to do with the PR or not.  _Theoretically_, we should never need to do this, but as long as commits and PRs that don't pass Travis keep getting merged in......

Ebag333:
@blitzmann please reopen as I've broken out the hotfix.

blitzmann:
> Anything that fails Travis should be fixed, regardless if it has anything to do with the PR or not. Theoretically, we should never need to do this, but as long as commits and PRs that don't pass Travis keep getting merged in......

That's fine. I still need to enable protection, which should prevent that.

PR #1050, opened 2017-03-16
Hotfix - Matplotlib import

Ebag333:
@blitzmann not sure what happened here with imports.  This is an absolute mess, from an import perspective.  There must have been a bad merge or force push at some point, because it's basically half like the branch I have where I fixed it, and half what it was originally. :(

Anyway, cleaned that up.

Apparently stacktraces can be generated deep within matplotlib, and we were only catching import errors.  Now we'll catch everything and bubble up a critical exception (to trip the FingersCrossedHandler).  Normal import errors will simply continue to be a warning and won't trip a log dump.

### DIFF
--- a/gui/graphFrame.py
+++ b/gui/graphFrame.py
@@ -29,13 +29,14 @@ import gui.mainFrame
 import gui.globalEvents as GE
 from gui.graph import Graph
 from gui.bitmapLoader import BitmapLoader
+import traceback
 
 pyfalog = Logger(__name__)
 
 try:
     import matplotlib as mpl
 
-    mpl_version = int(mpl.__version__[0])
+    mpl_version = int(mpl.__version__[0]) or -1
     if mpl_version >= 2:
         mpl.use('wxagg')
         mplImported = True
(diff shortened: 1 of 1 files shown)


blitzmann:
Will take a look when I get home, looks good though.

Were you able to reproduce? I tried the windows build before I made the release, no problems. I did not check graphs though. Did you have to do something to reproduce?

blitzmann:
Also, thanks for the super quick response time on these issues. You rock!

Ebag333:
>Were you able to reproduce? 

Nope, which is why on the issue I mentioned:

>It's not exactly a fix, but at least lets us gracefully handle the failure until we figure out the root cause.

_Something_ is causing it to stacktrace super deep within matplotlib.  Maybe it's because of the mess that the imports were in, but I don't see how that would be since it's the first import that broke.

If you look at the line it stacktraced on, it's just an import.  We'll catch that so we can gracefully fail now, and log the stacktrace to the log, but no I don't know why it's happening.

Hoping that a new build with the cleaned up imports and extra logging will shed more light on it.  (Sorry, I know what a PITA builds are.)

PR #1051, opened 2017-03-16
Fix bad merge of #943

Ebag333:
Looks like there was a bad merge or overwrite, and parts of #943 got dropped/broken.  This fixes that.

@blitmann I'm submitting this against master as it's half implemented and in a weird state.

The global ammo picker is also broken, because it's not imported, but we have it in the preference window. :(


gui/builtinContextMenus/__init__.py
```diff
@@ -1,6 +1,6 @@
 __all__ = [
```
blitzmann: If we're not doing `import *`, this file isn't needed anymore, correct?

blitzmann: Just let me know here and I can remove it, gonna merge this now

Ebag333:
We don’t technically need the all anymore. Still should have the init file itself.

Ebag333:
it doesn't hurt to have, and can be useful for tests.

PR #1057, opened 2017-03-17
Fit service logging

Ebag333:
Since we repeatedly get reports of issues from users, but can't replicate them, I went ahed and added a ton of logging to the Fit service.  Not all of the log messages capture as much data as I'd like (I got lazy about halfway through), but it'll shed a _lot_ of light when users are having stacktraces or hangups.


blitzmann:
thanks, should help figure out the steps taken by the user. :)

Issue #1055, opened 2017-03-17
Crash on deleting fit


Ebag333:
Steps to reproduce:

1. Create new fit.
2. Rename.
3. Change character.
4. Open new tab
5. Copy first fit.
6. Change character.
7. Delete first fit

I could _only_ get this to work when running in non-debug mode.  Grrrr.

I did catch this error in the logs:
`[2017-03-17 17:58:37.261000] WARNING: gui.builtinViews.fittingView: Caught dead object`

I'm suspecting we might be trying to recalc the fit we just deleted.  Also, we have absolutely no logging on deleting objects, which is just garbage.

A similar bug (same basic steps):
>response : desired fit is deleted from the Fitting Browser pane alas the character selected on the fit I'm viewing changes to the same as in the first (leftmost) of the open fits on screen

Thanks to doc for reporting this on Pyfa Slack!

Ebag333:
First method _should_ fall back to the first page, then recalc.

Another way to crash it.

1. Launch pyfa
2. Clone existing fit.
3. Delete existing fit.

Ebag333:
From `fittingView.py`:

```
    def Destroy(self):
        self.parent.Unbind(EVT_NOTEBOOK_PAGE_CHANGED, handler=self.pageChanged)
        self.mainFrame.Unbind(GE.FIT_CHANGED, handler=self.fitChanged)
        self.mainFrame.Unbind(gui.shipBrowser.EVT_FIT_RENAMED, handler=self.fitRenamed)
        self.mainFrame.Unbind(gui.shipBrowser.EVT_FIT_REMOVED, handler=self.fitRemoved)
        self.mainFrame.Unbind(gui.marketBrowser.ITEM_SELECTED, handler=self.appendItem)

        d.Display.Destroy(self)
```

The culprit is `d.Display.Destroy(self)`.  It's literally destroying itself, which means that everything else we attach to self (like `self.getActiveFit()` and `self.mainFrame`) all no longer work.  This is an old bit of code (2010), so most likely we weren't attaching all this stuff to `self` back then.

Removing this line of code doesn't _seem_ to break anything, but @blitzmann can you run it through it's paces and see if you see any issues?

I'm going to submit a PR shortly with this change.  Since this is a fairly easily crash to run into, I'll push it against `master`.

blitzmann:
[...]

PR #1058, opened 2017-03-17
Hotfix - Fixes crashes when pyfa destroys self

Ebag333:
Just a quick one liner to address the crash issues in #1055


Ebag333:
Huh. Not sure what happened here....

PR #1059, opened 2017-03-17
Hotfix - Fixes crashes when pyfa destroys self

Ebag333:
Just a quick one liner to address the crash issues in #1055


blitzmann:
Per https://github.com/pyfa-org/Pyfa/issues/1055#issuecomment-287590814 this causes the page to never actually delete when you close a tab. This in't the proper solution to the fit deletion crash.

Ebag333:
Maybe give me at least a few minutes to respond before closing a PR? :)

PR #1062, opened 2017-03-22
Real Eos Tests!!!!!!!!!!!!!!!!!!

Ebag333:
Thanks to some (okay, a lot of) help from @blitzmann, I got unit tests working for Eos!

These are setup a _little_ differently than the rest of the code, so bear with me.

In order to not have to do a ton of steps repeatedly (or cram all of our unit tests into a single test, _which is bad_), I broke up the various build functions into fixtures.  These are quite slick, but have one catch, we can only return one thing from them.

So, doing what I do, I stuffed them into a named dict. :D

This lets us do things like have a helper for all the Saveddata:

```
@pytest.fixture
def Saveddata():
    from eos.saveddata.ship import Ship
    from eos.saveddata.fit import Fit
    from eos.saveddata.character import Character
    from eos.saveddata.module import Module, State

    helper = {
        'Ship'     : Ship,
        'Fit'      : Fit,
        'Character': Character,
        'Module'   : Module,
        'State'    : State,
    }
    return helper
```    

To access it, you simply do (from your `test_` module):

```
from helpers import DBInMemory as DB
from helpers import Saveddata

def test_something(DB, Saveddata):
    ship = Saveddata['Ship'](item)
    fit = Saveddata['Fit'](ship, "My Rifter Fit")
```

This lets us do a few things.

1. Have a single easily callable source for things, without having to remember/import the whole thing.
2. Create the DB connection simply by importing then passing it as a parameter.
3. We can have fixtures inside fixtures (for example, one fixture uses `DB, Gamedata, Saveddata` to create `Rifter`, which our test function uses via `test_rifter_cpu_output(DB, Saveddata, Rifter)`.

The whole idea was to be able to write tests quickly and painlessly, without having to remember which `Fit` you are referring to (we have two in Eos!), and we can easily expand a single location rather than rewriting import lines repeatedly.

In the example @blitzmann gave me for building a Rifter, this took a 40 line test down to 22.  And in many ways, that's not a particularly good test (even though I'm including it here). It's a great _example_, but in my experience if you have tests that read as:
```
<do stuff>
assert thing == something
<do more stuff>
assert thing == something
<do even more stuff>
assert thing == something
```

That tends to be a hard test to read/troubleshoot, and is a huge PITA when it breaks and you have to figure out why.

A better test design:
```
<do stuff>

assert thing1 != thing2
assert redfish != bluefish
assert eggs == green
assert ham == green
```

Something to be aware of, if you want your unit test to access the database, you _must_ call the database build as the first parameter.  Only the first test that runs will actually build it, the rest will used cached versions, but each and every one must have it in order to guarantee no issues.


Ebag333:
Well, the test is passing on Windows but failing on Travis.  Not sure why yet.

Ebag333:
We have test coverage!

https://puu.sh/uVxtR/44f4e91696.png

Ebag333:
Replaced with #1065

Issue #686, opened 2016-07-21
Shield recharge rate is actually omni tank


rptb1:
The first column in the "Shield Recharge rates" has a figure labelled on the icon tooltip as "Passive shield recharge" but is in fact the omni tank figure.  This can be verified by installing or removing a module that changes resistances: it changes the figure, but does not alter the passive regen.  So this figure (while very useful) is not labelled correctly.

To reproduce, import this fit, and toggle the EM Ward Amplifier.

[Gila, Gila]

Imperial Navy Drone Damage Amplifier
Caldari Navy Shield Power Relay
Drone Damage Amplifier II

Large Shield Extender II
Large Shield Extender II
Large Shield Extender II
Large Shield Extender II
Pithum B-Type Thermal Dissipation Amplifier
Pithum B-Type EM Ward Amplifier

Rapid Light Missile Launcher II
Rapid Light Missile Launcher II
Rapid Light Missile Launcher II
Rapid Light Missile Launcher II
[Empty High slot]

Medium Core Defense Field Purger II
Medium Core Defense Field Purger II
Medium Core Defense Field Purger II

Vespa I x6
Vespa II x4

Scourge Fury Light Missile x1936
Caldari Navy Inferno Light Missile x2016
Nanite Repair Paste x20

OISumeko:
You are looking at the EHP values for the regen, that is to say that the number shown indicates the amount of damage you can tank against the currently selected damage profile passively. If you click the 'EHP' button in the top right the values will show raw hp regen.

![](http://i.imgur.com/q1GygT3.png)

Now if you add/remove your EM ward you will see the raw value doesn't change. 

![](http://i.imgur.com/Ys5BOH8.png) ![](http://i.imgur.com/y2YZa5X.png)

rptb1:
Thank you for clarifying that.

Could I suggest that the label you get when you mouse over the column changes when the EHP button is up or down? Or the labels HP/s change to EHP/s? I'm not sure how to make this clearer, but it puzzled several of us!

blitzmann:
I can look into it, no promises though ;)

blitzmann:
Found this while going through old issues, didn't take long at all to implement. 

https://i.gyazo.com/b2bda98ff63a5f6a7419169f916aa626.mp4

Hope that's what you wanted. :)

PR #1071, opened 2017-03-26
Switch units in recharge pane based on effective HP toggle

blitzmann:
See #686

https://gyazo.com/b2bda98ff63a5f6a7419169f916aa626


Ebag333:
Not to put a wrinkle in this, but this is completely rewritten in Gnosis. ;)

Issue #961, opened 2017-01-29
Database schema improvments


blitzmann:
Starting a list to determine what we can do about database schema (specifically user database) to improve performance / relational integrity.

* Add `sqlite_autoincrement=True` to tables which have a primary key (maybe not all tables). This will prevent SQLite from reusing previous IDs, which has been a thorn in our side with database corruption for years. See http://docs.sqlalchemy.org/en/latest/dialects/sqlite.html#sqlite-auto-incrementing-behavior.
* Add CreateDate and ModifyDate columns to most tables, and have a SQLAlchemy hook that updates these on every record INSERT / UPDATE. This will help trouble shoot when things might happen. Need to look into how we could do this and what kind of size increase this instills on the database.
* Discuss column names. I've always hated how everything is ID instead of CharacterID or FitID. I also am partial to microsofts DB naming conventions (TitleCase). This would be a much bigger change if we were top change the column names, as we would have to ensure that everything that uses them uses the properly. As a stopgap, we could have aliases for them that are used in the code, but that could become messy as well. 

None of this is set in stone or planned, just throwing it out there for tracking.

Ebag333:
>Discuss column names. I've always hated how everything is ID instead of CharacterID or FitID. 

It depends on which use case you're talking about.

Take `Boosters`.  You have:
`ID, itemID, fitID, active`

This is a good table design.  You _don't_ want:
`boosterID, itemID, fitID, active`

because other tables will use `boosterID` to reference this table.  So by this convention, `ID` is always self referential (and _always_ increments, never null, etc) while `boosterID` is referencing a foreign table, and shouldn't increment, might or might not be null, etc.

It becomes redundant when you're referencing it if you do it the second way because you will end up with something akin to:
`WHERE foreigntable.boosterID = booster.boosterID`

If you do it the way `boosters` is currently laid out then you have:
`WHERE foreigntable.boosterID = booster.ID`

Which is much more readable IMHO.  It's also a more common convention that I see very often.


Now, speaking of ID's you have tables like `characterSkills` which drives me nuts.  There's no unique ID in this table, which is problematic on a number of levels.  So this table should be changed from:
`characterID, itemID, _Skill__Level`
to:
`ID, characterID, itemID, skillLevel`

Where `ID` auto increments.  Yes our code doesn't care about the self referential ID on characterSkills, but it's a standard practice and we can do a number of things with it (such as checking for duplicates more easily).  When querying counts, currently you have to do this:
[...]

PR #1073, opened 2017-03-27
Create and Modify fields for most database tables

blitzmann:
As discussed in #961. Preliminary testing shows this works well for existing and new databases. It's completely transparent to the user, and not something that is exposed at the moment (but we could later on if deemed necessary).

The primary motivation for this addition was to help with troubleshooting. If a user sends us a database file, having the times in which modules are added / modified, it's one more data point available to us to help piece together what happened. It also allows for some nifty things in the future (recently modified fits, for instance).

Three things to note:
- DATETIME is apparently a data type in SQLite, even though I continuously read that it isn't. This is the way SQLAlchemy creates it with it's own `DateType` class, but creating a table or column with that as part of the schema works fine. ¯\_(ツ)_/¯
- `fits` still contains a timestap of when the fit was created. Eventually I want to change over to the `created` field (or even `modified`) and abolish the `timestamp` field, but that's for another rainy day.
- Again on `fits`, I originally wanted the modified field to reflect all changes associated with that fit (modules added / removed, drones added / removed, etc) so that it represents the fit being modified on the whole, instead of simply modifying the name, character, etc. But I didn't see an easy way off hand, so this can be added later if deemed necessary.

Additionally, this does not include support for `sqlite_autoincrement` as discussed in the issue. I will submit another PR for that once it's been developed and tested.

### DIFF
--- a/eos/db/migrations/upgrade22.py
+++ b/eos/db/migrations/upgrade22.py
@@ -1,0 +1,46 @@
+"""
+Migration 22
+
+- Adds the created and modified fields to most tables
+"""
+import sqlalchemy
+
+
+def upgrade(saveddata_engine):
+
+    # 1 = created only
+    # 2 = created and modified
+    tables = {
+        "boosters": 2,
+        "cargo": 2,
+        "characters": 2,
+        "crest": 1,
+        "damagePatterns": 2,
+        "drones": 2,
+        "fighters": 2,
+        "fits": 2,
+        "projectedFits": 2,
+        "commandFits": 2,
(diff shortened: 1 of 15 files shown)


blitzmann:
This should be completed - I've tested on/off for the past week without issue. 

@Ebag333, since this is touching user databases, I'd like for you to test it out with the variety of databases you have. I expect that it would work fine, but please let me know. :)

blitzmann:
This is working fine as far as I'm aware, merging in.

PR #1074, opened 2017-03-27
Re-introduce Command Fit State

blitzmann:
Simple QoL improvement / a fix for a regression dating back to November.

Bit of background: I disabled the state column for the Command / Fleet view when we made the change over to the new command burst mechanics. The reason behind this is because the State column, which controls which image is show to represent state of item, only received `stuff`. Normally we check what `stuff` is, and figure out what to do from there (eg: if `stuff` is a module, check the modules state and display an image that works).

The problem originally with command fits is that it shares the same `stuff` type as projected fits. So there wasn't a way to differentiate a projected fit from a command fit in the State class.

But holy shit, this was super easy after taking a look at it again tonight. We can't differentiate on `stuff`, but what we do have is access to the calling view. So to re-enable the State column for command view, we check to see if we're calling from command or projected, then get it's information, then display based on active flag.

### DIFF
--- a/gui/builtinViewColumns/state.py
+++ b/gui/builtinViewColumns/state.py
@@ -68,11 +68,17 @@ class State(ViewColumn):
                                                                 "gui")
         elif isinstance(stuff, Fit):
             fitID = self.mainFrame.getActiveFit()
-            projectionInfo = stuff.getProjectionInfo(fitID)
 
-            if projectionInfo is None:
+            # can't use isinstance here due to being prevented from importing CommandView.
+            #  So we do the next best thing and compare Name of class.
+            if type(self.fittingView).__name__ == "CommandView":
+                info = stuff.getCommandInfo(fitID)
+            else:
+                info = stuff.getProjectionInfo(fitID)
+
+            if info is None:
                 return -1
-            if projectionInfo.active:
+            if info.active:
                 return generic_active
             return generic_inactive
         elif isinstance(stuff, Implant) and stuff.character:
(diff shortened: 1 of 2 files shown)


gui/builtinViewColumns/state.py
```diff
@@ -68,11 +68,17 @@ def getImageId(self, stuff):
                                                                 "gui")
         elif isinstance(stuff, Fit):
             fitID = self.mainFrame.getActiveFit()
-            projectionInfo = stuff.getProjectionInfo(fitID)
 
-            if projectionInfo is None:
+            # can't use isinstance here due to being prevented from importing CommandView.
+            #  So we do the next best thing and compare Name of class.
+            if type(self.fittingView).__name__ == "CommandView":
```
Ebag333: Could use:
`self.fittingView.__class__.__name__`

Not sure it's any faster though.

gui/builtinViewColumns/state.py
```diff
@@ -68,11 +68,17 @@ def getImageId(self, stuff):
...
+                info = stuff.getCommandInfo(fitID)
+            else:
+                info = stuff.getProjectionInfo(fitID)
+
+            if info is None:
                 return -1
-            if projectionInfo.active:
+            if info.active:
```
Ebag333: `info` is used _all over the place_ (281 occurrences).  I liked `projectionInfo` better. :)

gui/builtinViewColumns/state.py
blitzmann: True, I'll change is as it's not a function call. Thanks!

gui/builtinViewColumns/state.py
blitzmann: Well it's no longer projection info. It can be command info now, so it needs to be more generic. It's a variable scoped to this function, so there shouldn't be any issue in determining the meaning of it.

PR #1079, opened 2017-03-28
Locale Tests

Ebag333:
See #1008 for history.


Ebag333:
>`[Pyfa] New branch "locale_tests_3" was pushed by Ebag333`

Well that's weird.

I guess now that I have access, pyCharm wants to move the branch to Pyfa-org when creating the PR.

Ebag333:
>`================= 12 passed, 3 pytest-warnings in 7.95 seconds =================`

Well ain't that pretty?

@blitzmann I think this is ready to merge.

blitzmann:
Sorry, thought I had taken care of this one. Looks fine. I like the done tag, thank you for that. :)

BTW, I read an interesting article about squashing pull requests when merging feature branches (github has an option to do that), and I want to make that a thing. Not sure if you've ever looked into it, but it basically it takes all the "fluff" commits within the branch (so, commits to fix tox, merging dev to update branch, etc) and the "content" commits and squashes them into a single commit that is basically the end result of the feature being added. Great example would be the other Tests PR that you have active where you were making commits to figure out what Travis was doing. I think it would keep our history much cleaner.

I've been doing it for my pull requests to get rid of the tox commits and dev merges and it works out well. Would also help in case something needs to be reverted - just revert that one commit instead of trying to revert multiple commits that revolve around that feature. It's not an exact representation of what happened in a branch, but we don't necessarily need that for the pyfa-org repo. If we want to keep the branch around on our local repos for historical purposes, that could be a thing.

The only thing that I'm not sure of is who gets credit when squashing the PR and committing... would it be the person that does the action? I want to make sure users that do PRs get the credit, so this may all be a moot point. I'm going to try to test this stuff on my branches to see. 

lol, sorry for the ramble :D

Ebag333:
I'm not opposed to it, but you do lose commit history, so some might not like that. 

You can squash when you rebase too, I was tempted to do that with this one because of the iterations, but rebasing public branches is dangerous. 

And time you want me to squash commits, I can recreate the branch very quickly and fairly painlessly, it'll be a replay of the commits so hashes will differ but meh, who cares if the original branch dies. So if you really are going to flip a table over it, let me know and I will clean it up for you.

blitzmann:
> I'm not opposed to it, but you do lose commit history, so some might not like that.

I'm aware, which is the other side of the coin of "should we squash". But that's why if we do start using squashing and want granular commit history we would keep the branches on our personal repos. And then obviously the branch should end once squashed and merged, and not continued to be developed on.

I just think squashing is cleaner for the history - we can have pyfa-org include the path from A to B, without having to include all the truckstops, carwashs, and waypoints in between (fuck that was super corny, but whatevs).

We'll see what the tests show of how credit is shown, then we can talk further strategy on how we may want to do it and manage it. :D

Ebag333:
Toss a test commit one one of these PRs then  squash  and merge. :)

blitzmann:
> Toss a test commit one one of these PRs then squash and merge. :)

Tested on my repo, seems whoever pushes the button gets credit, which makes sense but is kinda shitty. 

But since you have write access, you should be able to squash and merge yourself so that you get credit. 

image
 
We can do this on a case by case basis for branches that have a lot of fluff and continuous additions to them. If I need a squash I'll have you do it so that you get the creds. 

For third-party contributors, we'll just merge their pull requests in and not squash - they are few and far between anyway.

Fire away :|)

Ebag333:
>You're not authorized to merge this pull request

Also, is that what the force push against dev was about? I'd rather lose credit than have another force push.

Look at the history in Slack, Dev looks pretty messed up.

blitzmann:
>Also, is that what the force push against dev was about? I'd rather lose credit than have another force push.

I force pushed my dev repo, not pyfa-org. pyfa-org/dev looks fine to me.

> You're not authorized to merge this pull request

I'll look into the permissions... thought that was squared away

Ebag333:
>I force pushed dev to my repo after testing, not pyfa-org. pyfa-org/dev looks fine to me.

Yup.  It's the way it's formatted on Slack, makes it look like it was to Pyfa-org when it wasn't.  😡

PR #1080, opened 2017-03-29
System wide installation

StinGer-ShoGuN:
I've been working those last weeks on providing Gentoo ebuilds for Pyfa.
Pyfa would be installed in the `site-packages` directory of the local Python installation... which raises some issues...

Some files, such as `eve.db` or `iconsIDs.yaml` do not belong there and should be moved to more appropriate locations (e.g. `/usr/share/pyfa`).
In addition, default HTML export path would not be writeable for a user so HTML exports would fail.

This pull request addresses those two issues and would also probably help making other Linux distribution packages.
It uses a long time available feature of Pyfa, the `configforced.py` file which would specify another directory to look for `eve.db`.
And it changes the default HTML export path to `~/.pyfa/pyfaFits.html`.

The possibly most bothering change is the default export path: I'm open to proposals ! 😄 

*For now, I'm focusing on providing Gentoo ebuilds for every Pyfa release. This is still work in progress. I may try to build `.deb` and `.rpm` (and other...) later.*

### DIFF
--- a/config.py
+++ b/config.py
@@ -26,6 +26,7 @@ expansionVersion = "1.0"
 evemonMinVersion = "4081"
 
 pyfaPath = None
+dataPath = None
 savePath = None
 saveDB = None
 gameDB = None
@@ -58,6 +59,7 @@ def getDefaultSave():
 def defPaths(customSavePath):
     global debug
     global pyfaPath
+    global dataPath
     global savePath
     global saveDB
     global gameDB
@@ -92,6 +94,14 @@ def defPaths(customSavePath):
 
     # The database where we store all the fits etc
     saveDB = os.path.join(savePath, "saveddata.db")
+
+    # pyfa shared data directory (used for Linux system-wide installation).
+    dataPath = getattr(configforced, "dataPath", dataPath)
+    if dataPath is None:
+        if "site-packages" in pyfaPath and sys.platform.startswith("linux"):
+            dataPath = os.path.join("/usr", "share", "pyfa")
+        else:
+            dataPath = pyfaPath
 
     # The database where the static EVE data from the datadump is kept.
     # This is not the standard sqlite datadump but a modified version created by eos
--- a/config.py
+++ b/config.py
@@ -96,7 +96,9 @@ def defPaths(customSavePath):
     # The database where the static EVE data from the datadump is kept.
     # This is not the standard sqlite datadump but a modified version created by eos
     # maintenance script
-    gameDB = os.path.join(pyfaPath, "eve.db")
+    gameDB = getattr(configforced, "gameDB", gameDB)
+    if not gameDB:
+        gameDB = os.path.join(pyfaPath, "eve.db")
 
     # DON'T MODIFY ANYTHING BELOW
     import eos.config
--- a/config.py
+++ b/config.py
@@ -108,7 +108,7 @@ def defPaths(customSavePath):
     # maintenance script
     gameDB = getattr(configforced, "gameDB", gameDB)
     if not gameDB:
-        gameDB = os.path.join(pyfaPath, "eve.db")
+        gameDB = os.path.join(dataPath, "eve.db")
 
     # DON'T MODIFY ANYTHING BELOW
     import eos.config
--- a/service/settings.py
+++ b/service/settings.py
@@ -264,7 +264,7 @@ class HTMLExportSettings(object):
     def __init__(self):
         serviceHTMLExportDefaultSettings = {
             "enabled": False,
-            "path"   : config.pyfaPath + os.sep + 'pyfaFits.html',
+            "path"   : os.path.join(config.savePath, 'pyfaFits.html'),
             "minimal": False
         }
         self.serviceHTMLExportSettings = SettingsProvider.getInstance().getSettings(
--- a/gui/bitmapLoader.py
+++ b/gui/bitmapLoader.py
@@ -93,7 +93,7 @@ class BitmapLoader(object):
             except KeyError:
                 print("Missing icon file from zip: {0}".format(path))
         else:
-            path = os.path.join(config.pyfaPath, 'imgs' + os.sep + location + os.sep + filename)
+            path = os.path.join(config.dataPath, 'imgs' + os.sep + location + os.sep + filename)
 
             if os.path.exists(path):
                 return wx.Image(path)


Ebag333:
>Pyfa would be installed in the site-packages directory of the local Python installation... which raises some issues...

I'm not sure Pyfa belongs there to be honest. Eos, sure, but it doesn't seem like the right place for an interactive GUI app to me. 

>Some files, such as eve.db or iconsIDs.yaml do not belong there and should be moved to more appropriate locations (e.g. /usr/share/pyfa).

Both those are essentially read only. I don't have an issue with adding eve.db to config.forced, but I still generally think that it belongs there. 

>In addition, default HTML export path would not be writeable for a user so HTML exports would fail.

We were just talking about this feature in another PR. It actually just needs to die entirely, it causes problems and adds a ton of complexity for almost no gain. 

>For now, I'm focusing on providing Gentoo ebuilds for every Pyfa release. This is still work in progress. I may try to build .deb and .rpm (and other...) later.

Have you looked at the recent build stuff added? See #978 and #968

I think that this is a better solution that what you seem to be attempting to do.

StinGer-ShoGuN:
I'm not replying in order. Just because. 👅 
>Have you looked at the recent build stuff added? See #978 and #968
>
>I think that this is a better solution that what you seem to be attempting to do.

Well, this is not "Gentooic". Plus I don't like it myself neither because it looks like a Windows thing to me. With such method, you are storing duplicates and duplicates and duplicates of your libs, and you load the same lib over and over again. I know resources are not an issue nowadays, but I don't like this way of programming and I prefer sharing/optimising when it's possible.
In addition, the idea of Gentoo is to build/run from source and fine tune your install. So freezing everything to run a binary blob is not really the spirit here.

>I'm not sure Pyfa belongs there to be honest. Eos, sure, but it doesn't seem like the right place for an interactive GUI app to me.

Haaa... That's an ever lasting debate... You know [Spyder](https://github.com/spyder-ide/spyder) ? Well, it's installed in `site-packages` and ran from there...

Don't get me wrong, I'm not saying here that Pyfa **must** be installed in `site-packages`. I quite agree with you and find it strange to have apps installed in `site-packages`. But given how Python works with `import`, if you don't want to clobber out your code with `sys.path.append`, I don't know any other way of getting it to work.
On the other hand, C/C++ apps have their shared object files installed in `/usr/lib64` for instance. If you think about it, it's kind of similar.

>Both those are essentially read only. I don't have an issue with adding eve.db to config.forced, but I still generally think that it belongs there.

Let me clarify myself: those files do not belong there *when* installing Pyfa in `site-packages`. When running from one folder, their location is fine. 😄

>We were just talking about this feature in another PR. It actually just needs to die entirely, it causes problems and adds a ton of complexity for almost no gain.

Can I kill it myself ? 👿

blitzmann:
On the whole this doesn't look problematic. I can see the reasoning behind setting the html export file to the user's directory by default, less so using a config force for the database, but it couldn't hurt. However...

I wholly agree with @Ebag333 here in that there seems like a better solution here to what you're attempting. I'm no Linux guru, so I'm speaking as someone looking in, but I'm not exactly sure why you're trying to install pyfa into `site-packages`. `site-packages` is meant for... well, packages. pyfa as a whole isn't a package - it's an application that uses packages. You don't `import pyfa`, althought I guess you could finagle it to work. It should be installed into `usr/bin/pyfa` if you want to use it system-wide, set the execution bit on `pyfa.py` (or create some other execution file that bootstraps and launches pyfa), and then install the pyfa dependencies into the `site-packages` directory. That way the application stays together and you don't have. Again, this comes form someone who's user Arch, Ubuntu, Mint... never used Gentoo, so it may be a completely different culture.

Additionally, you mentioned `iconsIDs.yaml`, so I assume this is getting it from git. If you're developing a `stable` package for your distro, you may want to look into using the github api to download the linux package, which removes the directories that the end-user doesn't need (such as `scripts/`).

> We were just talking about this feature in another PR. It actually just needs to die entirely, it causes problems and adds a ton of complexity for almost no gain.

It's not going to die entirely, only the auto-generation of the script. There's no problems with manually setting it off. :)

So yeah, the actual content of the PR is fine from my point of view. Just curious as to why you're doing what you're doing ;)

config.py
```diff
@@ -96,7 +96,9 @@ def defPaths(customSavePath):
     # The database where the static EVE data from the datadump is kept.
     # This is not the standard sqlite datadump but a modified version created by eos
     # maintenance script
-    gameDB = os.path.join(pyfaPath, "eve.db")
+    gameDB = getattr(configforced, "gameDB", gameDB)
+    if not gameDB:
+        gameDB = os.path.join(pyfaPath, "eve.db")
```
Ebag333: This is fine. I doubt it'll be used much, but no harm here.

service/settings.py
```diff
@@ -264,7 +264,7 @@ def getInstance(cls):
     def __init__(self):
         serviceHTMLExportDefaultSettings = {
             "enabled": False,
-            "path"   : config.pyfaPath + os.sep + 'pyfaFits.html',
+            "path"   : os.path.join(config.savePath, 'pyfaFits.html'),
```
Ebag333: This isn't fine.

Unfortunately it's a problem for paths with unicode in them.  (Any `os.path` stuff is.)  Hence the hacky way of dealing with that here.

Umm, plus as mentioned, this isn't really the right spot for these to go.

Ebag333:
>It's not going to die entirely, only the auto-generation of the script. There's no problems with manually setting it off. :)

Not to derail this, but let me ask this question, what's the benefit?

There's no more in game browser, so having it run a webservice to show fits is basically useless.  Nothing else takes HTML based fits natively (that I know of), it's mostly EFT based.

Keeping functionality around that's fairly complex when it's not really used doesn't make much sense.  I know you hate killing things, but the fact is that we've got two active devs to maintain a pretty large project, and we're spending cycles on something that's basically a dead function (even if it's just to tell people to turn it off).

If you can name an actual _in use_ use case for it, I'd love to hear it.

service/settings.py
StinGer-ShoGuN: Could you elaborate on this please. It seems to be a Windows related issue and I can't reproduce.

StinGer-ShoGuN:
> Additionally, you mentioned iconsIDs.yaml, so I assume this is getting it from git. If you're developing a stable package for your distro, you may want to look into using the github api to download the linux package, which removes the directories that the end-user doesn't need (such as scripts/).

Actually `iconsIDs.yaml` is not needed for running Pyfa. I kept it erroneously. My bad. (I work on the git repo but use the zip release to test install/execution.)

To clarify things a bit, `pyfa.py` will not reside in `site-packages`. It will actually be located in `/usr/lib/python-exec/python2.X/pyfa` (where `X` is the minor version of Python 2 installed on the system). And there will be a symlink `/usr/bin/pyfa` pointing to `/usr/lib/python-exec/python-exec2`.
I don't know if you are familiar with `python-exec`: it's a wrapper around Python scripts to run it with a given version of Python 2 or 3. So that you can have several versions of Python installed on the system at the same time and set particular apps to run on given versions, without having to launch your app with e.g. `python2.7 app.py` or whatever.

> On the whole this doesn't look problematic. I can see the reasoning behind setting the html export file to the user's directory by default, less so using a config force for the database, but it couldn't hurt. However...

Any suggestion on where to put the eve.db file is welcome ! And the way to point to it ! 😄 
I put it in `/usr/share/pyfa` because this is the usual location apps store their data.

StinGer-ShoGuN:
Off topic but
>There's no more in game browser, so having it run a webservice to show fits is basically useless. Nothing else takes HTML based fits natively (that I know of), it's mostly EFT based.

True. Hence the end of HTML exports. They can't be used any more anyway...

StinGer-ShoGuN:
Don't consider those last 3 commits, they are dev purposes only (should have created another branch...).

Ebag333:
>Any suggestion on where to put the eve.db file is welcome ! And the way to point to it ! 😄
I put it in /usr/share/pyfa because this is the usual location apps store their data.

>The imgs directory should be moved there also to be honest...

I don't have issues with you using config.forced to push those wherever you want.  If you want them in the user directory, no concerns here (as long as we're setting them either through config.forced or a run time flag).

Since it's static data, I'm still of the opinion that it's less terrible to put in site-packages than  other bits, but again, no concerns with moving it.

service/settings.py
Ebag333: It's actually not a Windows issue but a `os.path` issue.

If you pass a string (type of `str`) into `os.path`, and the string has unicode chars in it, it doesn't know how to handle it.

_Theoretically_, and this has only been proven in some isolated tests, all we have to do is:
`os.path.join(unicode(config.savePath), 'pyfaFits.html'),`

blitzmann:
Still very weird to me to put pyfa in site-packages, but who am I to say how you install and run things lol

I agree with ebag though about the path thing and wrapping `unicode()` around it to prevent issues. 

> Don't consider those last 3 commits, they are dev purposes only (should have created another branch...).

Please revert these commits then. They are too environment-specific for inclusion within pyfa. Once that's done, I don't really see a problem with merging this in if it helps you in whatever it is your trying to accomplish, as weird as it may be :P

And about the HTML export, I dislike removing functionality if it's not getting in the way - I never know who is using it and what the reach is. Removing the ability to automatically generate it will suffice. I don't particularly see much use in it, but I'm not the other 22K people that have downloaded pyfa. However, if something happen in which the export needs to be significantly refactored, we'll drop it completely at that time.

StinGer-ShoGuN:
I'll do it and rewrite branch history. 👿

Ebag333:
Once branch is published, rewriting branch history is generally a bad idea. We've learned that the hard way.

StinGer-ShoGuN:
Will create a new PR.

blitzmann:
Just FYI

> I'll do it and rewrite branch history. 

You can still revert. This is different than nuking the commits and force pushing. Will check the new PR, thanks :)

PR #1083, opened 2017-03-30
Add fit caching

Ebag333:
Adds fit caching to the Fit service.  When we recalc a fit, we stash it in the fit cache.  When we switch fits or get a fit, if it exists in the cache we pull it from there, and skip the recalc.

Doesn't seem to make a massive difference in behavior, but should be a good first step toward better cache handling.

### DIFF
--- a/service/fit.py
+++ b/service/fit.py
@@ -200,6 +200,10 @@ class Fit(object):
         if fit is None:
             fit = eos.db.getFit(fitID)
 
+        projected_fits = len(fit.commandFits) + len(fit.projectedFits)
+
+        # If the fit  has fits projected on it, always recalc because another fit affecting this one could have changed.
+        if 0 < projected_fits:
             if self.serviceFittingOptions["useGlobalCharacter"]:
                 if fit.character != self.character:
                     fit.character = self.character
@@ -227,23 +231,26 @@ class Fit(object):
         if fit is None:
             fit = eos.db.getFit(fitID)
 
-            if basic:
-                return fit
+        if basic:
+            return fit
 
-            inited = getattr(fit, "inited", None)
+        inited = getattr(fit, "inited", None)
 
-            if inited is None or inited is False:
-                if not projected:
-                    for fitP in fit.projectedFits:
-                        self.getFit(fitP.ID, projected=True)
-                    self.recalc(fit, withBoosters=True)
-                    fit.fill()
+        projected_fits = len(fit.commandFits) + len(fit.projectedFits)
 
-                # Check that the states of all modules are valid
-                self.checkStates(fit, None)
+        # If the fit is not initialized, or is projected, or has fits projected on it, always recalc because another fit affecting this one could have changed.
+        if inited is None or inited is False or 0 < projected_fits or projected is True:
+            if not projected:
+                for fitP in fit.projectedFits:
+                    self.getFit(fitP.ID, projected=True)
+                self.recalc(fit, withBoosters=True)
+                fit.fill()
 
-                eos.db.commit()
-                fit.inited = True
+            # Check that the states of all modules are valid
+            self.checkStates(fit, None)
+
+            eos.db.commit()
+            fit.inited = True
 
         return fit
 
@@ -1096,7 +1103,7 @@ class Fit(object):
 
     def recalc(self, fit, withBoosters=True):
         start_time = time()
-        pyfalog.info("=" * 10 + "recalc" + "=" * 10)
+        pyfalog.info("=" * 10 + "recalc: {0}" + "=" * 10, fit.name)
         if fit.factorReload is not self.serviceFittingOptions["useGlobalForceReload"]:
             fit.factorReload = self.serviceFittingOptions["useGlobalForceReload"]
         fit.clear()
--- a/service/fit.py
+++ b/service/fit.py
@@ -52,6 +52,7 @@ class Fit(object):
 
     def __init__(self):
         pyfalog.debug("Initialize Fit class")
+        self.cached_fits = []
         self.pattern = DamagePattern.getInstance().getDamagePattern("Uniform")
         self.targetResists = None
         self.character = saveddata_Character.getAll5()
@@ -194,18 +195,21 @@ class Fit(object):
         if fitID is None:
             return None
 
-        fit = eos.db.getFit(fitID)
+        fit = next((x for x in self.cached_fits if x.ID == fitID), None)
 
-        if self.serviceFittingOptions["useGlobalCharacter"]:
-            if fit.character != self.character:
-                fit.character = self.character
+        if fit is None:
+            fit = eos.db.getFit(fitID)
 
-        if self.serviceFittingOptions["useGlobalDamagePattern"]:
-            if fit.damagePattern != self.pattern:
-                fit.damagePattern = self.pattern
+            if self.serviceFittingOptions["useGlobalCharacter"]:
+                if fit.character != self.character:
+                    fit.character = self.character
 
-        eos.db.commit()
-        self.recalc(fit, withBoosters=True)
+            if self.serviceFittingOptions["useGlobalDamagePattern"]:
+                if fit.damagePattern != self.pattern:
+                    fit.damagePattern = self.pattern
+
+            eos.db.commit()
+            self.recalc(fit, withBoosters=True)
 
     def getFit(self, fitID, projected=False, basic=False):
         """
@@ -217,25 +221,30 @@ class Fit(object):
         pyfalog.debug("Getting fit for fit ID: {0}", fitID)
         if fitID is None:
             return None
-        fit = eos.db.getFit(fitID)
 
-        if basic:
-            return fit
+        fit = next((x for x in self.cached_fits if x.ID == fitID), None)
 
-        inited = getattr(fit, "inited", None)
+        if fit is None:
+            fit = eos.db.getFit(fitID)
 
-        if inited is None or inited is False:
-            if not projected:
-                for fitP in fit.projectedFits:
-                    self.getFit(fitP.ID, projected=True)
-                self.recalc(fit, withBoosters=True)
-                fit.fill()
+            if basic:
+                return fit
 
-            # Check that the states of all modules are valid
-            self.checkStates(fit, None)
+            inited = getattr(fit, "inited", None)
 
-            eos.db.commit()
-            fit.inited = True
+            if inited is None or inited is False:
+                if not projected:
+                    for fitP in fit.projectedFits:
+                        self.getFit(fitP.ID, projected=True)
+                    self.recalc(fit, withBoosters=True)
+                    fit.fill()
+
+                # Check that the states of all modules are valid
+                self.checkStates(fit, None)
+
+                eos.db.commit()
+                fit.inited = True
+
         return fit
 
     @staticmethod
@@ -1093,5 +1102,8 @@ class Fit(object):
         fit.clear()
 
         fit.calculateModifiedAttributes(withBoosters=False)
+
+        if fit not in self.cached_fits:
+            self.cached_fits.append(fit)
 
         pyfalog.info("=" * 10 + "recalc time: " + str(time() - start_time) + "=" * 10)


blitzmann:
Nice try, but it's not going to be this simple. ;)

The reason fits aren't cached like this at the moment (trust me, I tried something very similar a couple years back) is because any change to any fit needs to be accounted for in other open fits. This PR simply loads a fit into a ~~dictionary~~ list (dict would be faster, btw) and if it's there, uses it when switching tabs.

The problem stems from this: I have two fits open: A and B. B projects onto A. I change something in B, switch over to A's tab, and that change is not reflected until A is refreshed.

This requires a bit more nuance unfortunately - I have a couple ideas on how to incorporate it, and already have a local branch with some changes to help with that. If we think this is priority, I can definitely start looking at it again.

Ebag333:
>The problem stems from this: I have two fits open: A and B. B projects onto A. I change something in B, switch over to A's tab, and that change is not reflected until A is refreshed.

We can handle this one of two ways.

1) Flag any fits projected onto that they need to be refreshed.
2) If a fit has ships projecting onto it, always refresh.

Either would take care of this scenario and would be easy to add in.

Ebag333:
Okay, added an exception that if the fit is a projected fit, or has anything projected onto it, we always rerun the recalcs.

Here's how this looks when run currently:
```
Original:

[2017-03-30 18:09:15.647000] INFO: service.fit: ==========recalc: Heavy Tackle==========
[2017-03-30 18:09:23.729000] INFO: service.fit: ==========recalc time: 8.08200001717==========
[2017-03-30 18:09:23.743000] INFO: service.fit: ==========recalc: Heavy Tackle==========
[2017-03-30 18:09:24.129000] INFO: service.fit: ==========recalc time: 0.385999917984==========
[2017-03-30 18:09:28.254000] INFO: service.fit: ==========recalc: Gaston's Weird Burner==========
[2017-03-30 18:09:28.911000] INFO: service.fit: ==========recalc time: 0.65700006485==========
[2017-03-30 18:09:28.921000] INFO: service.fit: ==========recalc: Gaston's Weird Burner==========
[2017-03-30 18:09:29.216000] INFO: service.fit: ==========recalc time: 0.295000076294==========
[2017-03-30 18:09:29.583000] INFO: service.fit: ==========recalc: Gaston's Weird Burner==========
[2017-03-30 18:09:29.879000] INFO: service.fit: ==========recalc time: 0.296000003815==========
[2017-03-30 18:09:32.696000] INFO: service.fit: ==========recalc: Heavy Tackle==========
[2017-03-30 18:09:33.072000] INFO: service.fit: ==========recalc time: 0.375999927521==========
[2017-03-30 18:09:33.221000] INFO: service.fit: ==========recalc: Heavy Tackle==========
[2017-03-30 18:09:33.596000] INFO: service.fit: ==========recalc time: 0.375==========
[2017-03-30 18:09:35.853000] INFO: service.fit: ==========recalc: Gaston's Weird Burner==========
[2017-03-30 18:09:36.153000] INFO: service.fit: ==========recalc time: 0.301000118256==========
[2017-03-30 18:09:36.291000] INFO: service.fit: ==========recalc: Gaston's Weird Burner==========
[2017-03-30 18:09:36.596000] INFO: service.fit: ==========recalc time: 0.305999994278==========
```

To explain what's going on here:
1. Launch Pyfa
2. Open `Heavy Tackle`
3. Create new tab
4. In new tab, open `Gaston's Weird Burner`
5. Switch to the `Heavy Tackle` tab
6. Switch to the `Gaston's Weird Burner` tab.

Neither of these fits have projections on them (regular or command).  The `Heavy Tackle` fit is one I regularly use to stress test as it is a fairly complex fit (RAH, etc).

All in all, this took 11.07400012 seconds to run.

Now let's do the same steps with the new code using caching:
```
[2017-03-30 18:05:37.186000] INFO: service.fit: ==========recalc: Heavy Tackle==========
[2017-03-30 18:05:45.208000] INFO: service.fit: ==========recalc time: 8.02199983597==========
[2017-03-30 18:05:49.511000] INFO: service.fit: ==========recalc: Gaston's Weird Burner==========
[2017-03-30 18:05:50.158000] INFO: service.fit: ==========recalc time: 0.647000074387==========
[2017-03-30 18:05:50.527000] INFO: service.fit: ==========recalc: Gaston's Weird Burner==========
[2017-03-30 18:05:50.851000] INFO: service.fit: ==========recalc time: 0.325000047684==========
[2017-03-30 18:05:55.997000] INFO: service.fit: ==========recalc: Heavy Tackle==========
[2017-03-30 18:05:56.391000] INFO: service.fit: ==========recalc time: 0.394000053406==========
[2017-03-30 18:06:00.374000] INFO: service.fit: ==========recalc: Gaston's Weird Burner==========
[2017-03-30 18:06:00.700000] INFO: service.fit: ==========recalc time: 0.327000141144==========
```

Total time? 9.715000153 seconds.

That's over a 12% time savings! 😀 

Okay, so right now what we're doing is _marginally_ hacky by just always running recalcs for fits that are projected or have projections.  Still means that the majority of fits will be significantly faster, since most fits won't have projections on them.

What we _can_ do is when we change a fit that projects onto something, we can loop through the cached fits and flag them that they need to rerun a recalc.  Then we'd only run the recalc when needed.


What's also interesting is that even though `switch` and `get` are now blocked from running a recalc, there's still one being triggered somewhere anyway (see how the two fits are recalced when switching tabs), so I suspect that the projected fits issue wouldn't _actually_ be an issue.  But better to be conservative on this.

Ebag333:
>What's also interesting is that even though switch and get are now blocked from running a recalc, there's still one being triggered somewhere anyway (see how the two fits are recalced when switching tabs), so I suspect that the projected fits issue wouldn't actually be an issue. But better to be conservative on this.

Ahah.  It's because the two fits have different characters on them.

Hmm.  We don't handle that very well. 🤔

Ebag333:
Here's a .gif showing the difference.

Current dev:
https://puu.sh/v3J1A/4800621d8e.gif

This PR:
https://puu.sh/v3ITH/12e12ef22e.gif

Both fits were opened, then closed, that way we avoided the lag from loading the effects.

It's not a huge difference between the two, but it is noticeable.

Ebag333:
```
        if inited is None or inited is False:
            if not projected:
                for fitP in fit.projectedFits:
                    self.getFit(fitP.ID, projected=True)
                self.recalc(fit, withBoosters=True)
                fit.fill()
```

Well this is interesting.  Basically, we'll never ever call `getFit` if the fit is `inited`.  So if we open a fit, then project on it, it won't ever hit this bit of code, making it a bit questionable why we do it in the first place.

Though, to be fair, it will load all the projected fits into the cache, so would make that faster....

blitzmann:
Sorry bud, this is a pretty hacky way of doing it and I'm not on board with it. I don't want to keep a list of "cached" fits around for now unless it's absolutely necessary, and it really shouldn't be (btw, fits are already cached at the DB level, so there's the small overhead of maintaining two caches of fits)

Not only that, but this is happening when opening up a retribution fit with three modules and a projected ship with your latest branch:

```
[2017-03-31 01:31:26.793000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:26.897000] INFO: service.fit: ==========recalc time: 0.104000091553==========
[2017-03-31 01:31:26.910000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:27.015000] INFO: service.fit: ==========recalc time: 0.105000019073==========
[2017-03-31 01:31:27.016000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:27.120000] INFO: service.fit: ==========recalc time: 0.103999853134==========
[2017-03-31 01:31:27.135000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:27.250000] INFO: service.fit: ==========recalc time: 0.115000009537==========
[2017-03-31 01:31:27.266000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:27.374000] INFO: service.fit: ==========recalc time: 0.108000040054==========
[2017-03-31 01:31:27.443000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:27.552000] INFO: service.fit: ==========recalc time: 0.108999967575==========
[2017-03-31 01:31:27.554000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:27.659000] INFO: service.fit: ==========recalc time: 0.105000019073==========
[2017-03-31 01:31:27.668000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:27.777000] INFO: service.fit: ==========recalc time: 0.108999967575==========
[2017-03-31 01:31:27.785000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:27.901000] INFO: service.fit: ==========recalc time: 0.115999937057==========
[2017-03-31 01:31:27.910000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:28.016000] INFO: service.fit: ==========recalc time: 0.107000112534==========
[2017-03-31 01:31:28.024000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:28.128000] INFO: service.fit: ==========recalc time: 0.104000091553==========
[2017-03-31 01:31:28.129000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:28.237000] INFO: service.fit: ==========recalc time: 0.108000040054==========
[2017-03-31 01:31:28.247000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:28.353000] INFO: service.fit: ==========recalc time: 0.105999946594==========
[2017-03-31 01:31:28.363000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:28.498000] INFO: service.fit: ==========recalc time: 0.134999990463==========
[2017-03-31 01:31:28.500000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:28.620000] INFO: service.fit: ==========recalc time: 0.119999885559==========
[2017-03-31 01:31:28.628000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:28.739000] INFO: service.fit: ==========recalc time: 0.111000061035==========
[2017-03-31 01:31:28.801000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:28.910000] INFO: service.fit: ==========recalc time: 0.108999967575==========
[2017-03-31 01:31:28.914000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:29.022000] INFO: service.fit: ==========recalc time: 0.108000040054==========
[2017-03-31 01:31:29.033000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:29.140000] INFO: service.fit: ==========recalc time: 0.107000112534==========
[2017-03-31 01:31:29.156000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:29.267000] INFO: service.fit: ==========recalc time: 0.111000061035==========
[2017-03-31 01:31:29.303000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:29.415000] INFO: service.fit: ==========recalc time: 0.111999988556==========
[2017-03-31 01:31:29.418000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:29.527000] INFO: service.fit: ==========recalc time: 0.108999967575==========
[2017-03-31 01:31:29.535000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:29.642000] INFO: service.fit: ==========recalc time: 0.106999874115==========
[2017-03-31 01:31:29.651000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:29.760000] INFO: service.fit: ==========recalc time: 0.108999967575==========
[2017-03-31 01:31:29.768000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:29.875000] INFO: service.fit: ==========recalc time: 0.107000112534==========
[2017-03-31 01:31:29.883000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:29.991000] INFO: service.fit: ==========recalc time: 0.108000040054==========
[2017-03-31 01:31:29.992000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:30.102000] INFO: service.fit: ==========recalc time: 0.109999895096==========
[2017-03-31 01:31:30.110000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:30.221000] INFO: service.fit: ==========recalc time: 0.111000061035==========
[2017-03-31 01:31:30.232000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:30.341000] INFO: service.fit: ==========recalc time: 0.108999967575==========
[2017-03-31 01:31:30.343000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:30.457000] INFO: service.fit: ==========recalc time: 0.114000082016==========
[2017-03-31 01:31:30.465000] INFO: service.fit: ==========recalc: Retribution fitq==========
[2017-03-31 01:31:30.572000] INFO: service.fit: ==========recalc time: 0.107000112534==========
```

This causes very noticeable performance problems, more so if there is a command fit attached as well. I don't know where it stems from since I haven't looked too far at the code, but I assume it's that additional for loops you added in the latest commit. v0v.

I have a branch out there with work started on this that I think is more simple and elegant. It isn't completed, and needs to be thoroughly tested (for example, I never tested command boosts), but it utilizes the `__calculated` field of the fit that we already kinda do (but don't use to my knowledge) to determine if it needs to recalc on switch or not. 

I cleaned up the crap from that branch and pushed here: https://github.com/blitzmann/Pyfa/tree/qol/fit-calculated-flag

This is a valiant effort, and it definitely goes in a direction, but I don't think it's the right direction as to how we handle this. :)

Ebag333:
>Sorry bud, this is a pretty hacky way of doing it and I'm not on board with it. I don't want to keep a list of "cached" fits around for now unless it's absolutely necessary, and it really shouldn't be (btw, fits are already cached at the DB level, so there's the small overhead of maintaining two caches of fits)

It's not hacky at all.  The objects are floating in memory anyway, so this just lets us reuse them as needed.

And yes, the fits are cached at the DB level, but I clearly just showed how it's a _massive_ performance increase by using the fit in memory vs the DB cache.

We're already maintaining two caches of fits, each and every fit floating loose, and the DB.

>Not only that, but this is happening when opening up a retribution fit with three modules and a projected ship with your latest branch:

Yup, but that has to do with a bad `if` statement and not the principle behind it.

A `recalc fit` flag is a good idea (in fact one mentioned earlier....) and we already have one we can use in `inited`.  Nothing's really using it, it's only used if it's the first time a fit has been loaded, but very little extra is done when a fit hasn't been inited.  Again, another feature that seems half implemented.

I believe I've also found other ways to significantly improve speed savings (namely all the other functions going back to the DB) as well as found a function call that seems to be formatted incorrectly.

Once I'm done with this, should be a huge difference in speed.

Ebag333:
>Once I'm done with this, should be a huge difference in speed.

Okay, maybe not a _huge_ difference in speed, but shaves another tenth of a second off for the same test as outlined above (open two fits, flip between them).

I also tested projected fits.  Works beautifully.  No more recalc spam due to a bad `if` statement. :)

@blitzmann seems I've plugged the big hole you brought up.  Care to try and find another? :)

Ebag333:
@blitzmann in commit 08be50caf2122ae8ff27b85677b7944e8d05af12 why did you disable calculating attributes with boosters?

Line in question is:
`fit.calculateModifiedAttributes(withBoosters=False)`

Used to be:
`fit.calculateModifiedAttributes(withBoosters=withBoosters)`

Ebag333:
@blitzmann I rewrote `calculateModifiedAttributes`. 😁 

1. _MUCH_ simpler logic flow.
2. Breaks out the flow of projected/command/self.  This gives us more control over when we run these, and lets us always run self last. Since it's recursive, it will walk up the chain.  I think that this will help us quite a bit with some of our issues with timing command bursts and projections.
3. Removes some nasty shadowing with `calculateModifiedAttributes`.  Since we have a ton of other stuff that has a _different_ method with the same name, and we call that method from inside this one, it's extremely confusing.  This should remove a lot of confusion and make things easier to troubleshoot.
4. Gets rid of a case when we were deleting `self`. We still have to reassign which is gross, but it's better than what we had before.
4. Fixes a bunch of little one line bugs that I stumbled across.

All the tests pass, and I tested a bunch of different scenarios.

Ebag333:
Implemented @blitzmann branch, and built on it.

```
[2017-04-01 22:58:14.380000] WARNING: __main__: For (Lowsec bait) returning chain have been calculated
[2017-04-01 22:58:14.381000] WARNING: __main__: For (1mn ab dual rep scram) returning chain have been calculated
[2017-04-01 22:58:14.381000] WARNING: __main__: For (Gaston's Weird Burner) returning chain have been calculated
[2017-04-01 22:58:14.382000] WARNING: __main__: All fits in chain have already been calculated.  Skipping recalcs.
```

Ignore the `WARNING`, for some reason that module isn't outputting `pyfalog` stuff for some reason.  So this is just `print()` output, which still works...

In this case we've got a chain of 4 fits.

Projection is:
Heavy Tackle <- Gaston's Weird Burner <- 1mn ab dual rep scram <- Lowsec bait

So you can see it's rolling up the chain to make sure we have everything calculated.

Ebag333:
```
[2017-04-01 23:40:53.108000] INFO: service.fit: ==========recalc: Heavy Tackle==========
[2017-04-01 23:40:53.287000] WARNING: __main__: Validating: Heavy Tackle Status: False
[2017-04-01 23:40:53.287000] WARNING: __main__: For (Heavy Tackle) returning chain have been calculated
[2017-04-01 23:40:53.287000] WARNING: __main__: Calculating fit attributes for: Heavy Tackle
[2017-04-01 23:40:53.288000] WARNING: __main__: Runtime: early
[2017-04-01 23:40:57.417000] WARNING: __main__: Runtime: normal
[2017-04-01 23:41:00.959000] WARNING: __main__: Runtime: late
[2017-04-01 23:41:01.043000] INFO: service.fit: ==========recalc time: 7.93499994278==========
[2017-04-01 23:41:05.464000] INFO: service.fit: ==========recalc: Gaston's Weird Burner==========
[2017-04-01 23:41:05.487000] WARNING: __main__: Validating: Gaston's Weird Burner Status: False
[2017-04-01 23:41:05.488000] WARNING: __main__: For (Gaston's Weird Burner) returning chain have been calculated
[2017-04-01 23:41:05.488000] WARNING: __main__: Calculating fit attributes for: Gaston's Weird Burner
[2017-04-01 23:41:05.488000] WARNING: __main__: Runtime: early
[2017-04-01 23:41:05.516000] WARNING: __main__: Runtime: normal
[2017-04-01 23:41:05.931000] WARNING: __main__: Runtime: late
[2017-04-01 23:41:05.950000] INFO: service.fit: ==========recalc time: 0.486000061035==========
```

Same example as above (same two fits, open each, then toggle between them.

Original time was: 11.07400012 seconds

This time is: 8.421000003815 seconds

That is a _24%_ increase in speed.  The reason why the logs only show two fits? We don't log anything when just swapping tabs, since no recalcs are needed can toggle between the tabs a million times and it'll still stay at 8.421 seconds for recalc times. :)

I also removed the `inited` check as it's redundant.  We now always run a check on the modules states every time the fit is opened (to make sure it's valid), if the fit hasn't been calced or the module state is invalid it'll trigger a module recalc.  This lets us leverage the recalcs done for other fits.

blitzmann:
> but I clearly just showed how it's a massive performance increase by using the fit in memory vs the DB cache.

I don't know what this means. The DB cache is in memory - it's a python list. Look at the `@cachedQuery` stuff. Granted, this isn't something that should be accessed by the application directly, but the fact remains that we are effectively holding two containers of roughly the same data. 

> Care to try and find another? :)

Yes, the general way this is working. This feels tacked on and hacky, IMO, and there are better ways to try to do this. Does this work? Sure, it gets the job done, but there is just no need for the additional overhead which complicates matters. All so that switching fits is a few hundred ms faster. 

The thought process behind my branch is this: if it's already calculated (which is a flag that is already set for a fit), skip the calc on tab switch. However, if we're working on a fit that affects others some how, reset that calculated flag so that next time it's active, it gets calculated. This is simple, easy, intuitive, not doesn't include a bunch of logic to check to see if we've already cached, if it's projected, and if so what to do, etc. It's just cleaner.

Now, that all assumes it will work the way I expect it to. Again, my branch is a quick proof of concept that I probably need to expand on, it needs to be tested completely and thoroughly. But I believe it will work. If it doesn't, we can discuss other options, including this PR. But for now, please put a plug on this branch. I don't want you to potentially waste time :)

> There's a bunch of booster stuff in there (which is a terrible name, who picked that name?!?) that I feel is backwards (and wouldn't work if we didn't hard set withBoosters to always be False).

A lot of that is legacy stuff that I have been meaning to clean up. The change to command boosts obviously required a bit of reworking of the meaning of the variables, and there was a deadline to have that stuff implemented, so I didn't have time to go through and really scrub it clean. 

`withBoosters` is always initialized to False because, frankly , we don't need it at the onset anymore. We used to need it, but without looking at the old fleet code I can't recall exactly how it functioned (I think we loaded the fleet data up in the `getFit` and had to determine if the fit has boosters assigned to it, in which case it would do some additional stuff... it was very wonky). But right now it's used as a flag to tell us that we are stepping into a fit that is supposed to do command effects on a target fit. But yes, the variables could be a bit more intuitive.

> @blitzmann I rewrote calculateModifiedAttributes

Please don't take offense, but this isn't something I want touched without a damn good reason, and touching it has to be incremental. I'm very firm on this. Every single line in that function is there for a reason, and unless there's a bug to prove something it off in that function (for example, the recent one with command bursts), I will decline changes, especially if they aren't explicitly and thoroughly explained. This is the monolithic calc function that makes pyfa what it is, so I'm always very cautious about mucking with stuff in there.

I will make it a point in the next couple days to do two things:

* Continue testing the calculated "caching" thing and developing out that stuff on my branch so that fit switching is as efficient as possible. If you have ideas or test cases to test for, please let me know :D
* Go through and re-document the calc function, clean up the variables, etc. Hopefully that could give a bit more insight into why things are done the way they are. 

I'll submit PRs for both that we can discuss. :)

Ebag333:
>I don't know what this means.

It means that using the existing `Fit` is 10% faster that fetching it out of the DB cache. That's not a few milliseconds, it's significantly faster. 

The fit objects exist. We discard them into the ether, that makes no sense. This is much more Pythonic by reusing existing objects instead of constantly discarding and recreating them. 

It's not hacky at all, and you even have suggested a similar approach in the past. If nothing else we would want to register the object for garbage collection and cleanup, right now that's only done when Pyfa shuts down. 

>The thought process behind my branch is this: if it's already calculated (which is a flag that is already set for a fit), skip the calc on tab switch. However, if we're working on a fit that affects others some how, reset that calculated flag so that next time it's active, it gets calculated. This is simple, easy, intuitive, not doesn't include a bunch of logic to check to see if we've already cached, if it's projected, and if so what to do, etc. It's just cleaner.

Already done. One is not exclusive of the other. 

Also, it's not "a bunch of logic" to reuse the fit object. It's seriously like 4 lines. 

>But for now, please put a plug on this branch. I don't want you to potentially waste time

Your about a day late. :) 

>A lot of that is legacy stuff that I have been meaning to clean up. 

That method is pretty awful. 

>Please don't take offense, but this isn't something I want touched without a damn good reason, and touching it has to be incremental. I'm very firm on this. Every single line in that function is there for a reason, and unless there's a bug to prove something it off in that function (for example, the recent one with command bursts), I will decline changes, especially if they aren't explicitly and thoroughly explained. This is the monolithic calc function that makes pyfa what it is, so I'm always very cautious about mucking with stuff in there.

Again, too late. :) 

How about this for a reason, super complicated, slow, confusingly written, and some stuff is just wrong, but in this case two wrongs made a right. 

Also, bugs with recalc ordering that could not be easily handled with the current method. 

I would say that a 15% increase in recalc times is a pretty damn good reason, all the others aside. 

Oh, and being able to easily write meaningful tests against it, because the logic has been greatly simplified. The old/current one? Good luck hitting every edge scenario, not that half of it could be reached anyway. 

>This is the monolithic calc function that makes pyfa what it is, so I'm always very cautious about mucking with stuff in there.

Which is what tests are for, and why I didn't touch it until we had tests in place. All the tests pass, btw, and I extensively smoke tested it as well. I know that I didn't hit EVERY scenario, but I hit a lot. 

>Continue testing the calculated "caching" thing and developing out that stuff on my branch so that fit switching is as efficient as possible.

Caching and smart recalcs are two different things, albeit joined at the hip. 

Caching should be done at the Fit service level, thats where the fit objects live, and we always pass those in via self. The fit service handles fetching from the DB and processing actions against the fit, so it HAS to be loaded there anyway. If we try caching lower inside Eos, then we're throwing away what the Fit service passed in, and will see zero benefit from caching, since the Fit service will independently load it regardless. 

Smart recalcs (not rerunning recalcs when not needed) is something I mentioned early on as a good next step. I took what you had and ran with it. It was, btw, exactly what I was thinking. 

The two things are not exclusive of each other. Caching makes the most sense at the fit service level. Smart recalcs make the most sense inside Eos.

>Go through and re-document the calc function, clean up the variables, etc. Hopefully that could give a bit more insight into why things are done the way they are.

Documentation is a good idea. I added some, will go back and add more.

Ebag333:
Okay, boiled this down to the bare components and ran some tests.

Literally. I used our new test environment to _only_ load the bare components we needed.

Here are the times reported by the tests (this includes some overhead of creating/destructing objects)
https://puu.sh/v82Zb/f2c3bc815c.png

And here are the times for _just_ getting the fits (either from the local cache, or the DB cache).
```
1000000 of the Rifter fit (new): 9.21399998665
1000000 of the Rifter fit (old): 10.3989999294
1000000 of the four fits (new): 7.20000004768
1000000 of the four fits (old): 10.3380000591
```

In each case, we fetch one million fits.  The first two simply loop a million times and grab the same fit repeatedly, you'll see that the times are pretty close, though grabbing it out of the local cache is still 11% faster.  Significant increase in speed, but not game changing.

The second two only loop 250k times, but each loop grabs 4 fits.  There's much less overhead as we have a quarter of the time spent processing our `for` and `xrange` (250k loops vs 1M loops), so you want to be a little careful directly comparing the first two tests to the second two, but they are fairly comparable.

What's super interesting is that there is a _30% increase in speed_ for using the local cache over the DB cache.

This actually matches my tests that I ran in situ for the `getFit` function.  If I forced it to load a single fit repeatedly from the DB cache, the times were fairly close to on par with loading from a local cache (similar to test above).  If I forced it to load different fits in a row, the time _nearly doubled_ for loading with the DB cache.

The DB cache can retrieve the same fit fairly quickly (almost, though not quite, as fast as a local cache).  It is not very good at switching what it hands back.

It makes sense that I'd see 12-16% improvements "in the wild", because we mostly retrieve the same fit repeatedly (we run `getFit` a good 20-30 times), but the test I ran did include switching between fits a few times.

Regardless, even if we shove the "switching fits" scenario aside, I would still say that a 11% increase in speed is worth it.  Not only that, but the ability to be able to garbage collect and cleanup down the road, because we have fits that have been opened registered, rather than discarding them and recreating repeatedly.  I've noticed (as have others) that Pyfa becomes sluggish over time (if left running and used over several days), this would be a step toward being able to manage that.

Issue #1081, opened 2017-03-29
Projecting Linked Guidance Disruptors yields wrong result


lunedis:
# Submit a bug report bug report or feature request
### Expected behavior:
Missile range should show 94km * (1-0.381)^2 = 36km

### Actual behavior:
Missile range is 94km * (1-0.309)^2 = 45km

30.9% is the GD strengh without link

### Detailed steps to reproduce:

Apply Proteus Links to Crucifier.

Project Crucifier onto Caracal

### Fits involved in EFT format (Edit > To Clipboard > EFT):

```
[Crucifier, Crucifier fit]

[Empty Low slot]
[Empty Low slot]
[Empty Low slot]

Guidance Disruptor II, Missile Range Disruption Script
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]

[Empty High slot]
[Empty High slot]

[Empty Rig slot]
[Empty Rig slot]
[Empty Rig slot]
```
```
[Caracal, Caracal fit]

[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]

[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
Heavy Missile Launcher II, Caldari Navy Scourge Heavy Missile
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]

[Empty Rig slot]
[Empty Rig slot]
[Empty Rig slot]
```
```
[Proteus, Boost]

[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]

[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]

Information Command Burst II, Electronic Hardening Charge
Information Command Burst II, Sensor Optimization Charge
Information Command Burst II, Electronic Superiority Charge
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]

Medium Command Processor I
Medium Command Processor I
[Empty Rig slot]

Proteus Engineering - Augmented Capacitor Reservoir
Proteus Defensive - Warfare Processor
Proteus Offensive - Dissonic Encoding Platform
Proteus Propulsion - Gravitational Capacitor
Proteus Electronics - Dissolution Sequencer


Information Command Mindlink
```

### Release or development git branch? Please note the release version or commit hash:
1.28.1 Stable

[...]

PR #1085, opened 2017-03-30
Fix for #1081

blitzmann:
Break project application code out of main item iteration, and apply it after command boosts have been applied to the fit.

I do want to take another peak at it in a few days to see if there's any fat we can trim with that code now being in it's own block (for example review the items that are chained to see if there's anything unnecessary that's currently being run for projections, such as character, that I didn't bother to put brain cycles towards tonight). But for now this seems to fix the issue and works well.


blitzmann:
Couldn't do tox test as I'm on my new laptop which is having an issue with the interpreter. Will fix in branch asap. :D

PR #1086, opened 2017-03-30
Tests

Ebag333:
Fixes an existing test.  Adds a new one.

Here's a curiousity for you @blitzmann.

From `test_fit.py`:
```
def test_getAllFits(DB, RifterFit, KeepstarFit):
    assert len(Fit.getAllFits()) == 0

    DB['db'].save(RifterFit)
    DB['db'].save(KeepstarFit)

    # For some reason in Travis this adds the first fit twice.  WHY?!?
    assert len(Fit.getAllFits()) != 0

    # Cleanup after ourselves
    DB['db'].remove(RifterFit)
    DB['db'].remove(KeepstarFit)
```

So here's the curious part.

When run on Windows, it counts as you'd expect.  After we save the Rifter fit, we have one.  After the Keepstar fit, we have 2 fits.

In Travis, after saving the Rifter fit, we have _two_ fits.  It's the Rifter fit twice.  After saving the Keepstar fit, we have 3.

Odd, no?

.

Oh, and did I mention we can write tests against wx now? :D

Obviously we can't do any tests that require actual GUI interaction, but we can run tests against anything that imports wx.  So this opens up the `\GUI\` folder to have tests added to it.


blitzmann:
> `Tweak travis`

Christ

image

>In Travis, after saving the Rifter fit, we have two fits. It's the Rifter fit twice. After saving the Keepstar fit, we have 3.

It is odd... When you say save, you mean saved to the user database? As in, there's two separate Rifter fits in the sqlite database? 

Perhaps there's a test that is done earlier in Travis that is saving the Rifter, and then it gets to this test and saves it again. This is coming from someone who doesn't even have tests working on his local machine, so take that suggestion with a grain of salt. Dunno how the tests work, but maybe `select * from fits` after each test to see which tests are adding it... I assume Travis can handle print statements that we can then see?

Apart from that, maybe test in Linux and see if it has the same behavior. ¯\_(ツ)_/¯

> Oh, and did I mention we can write tests against wx now? `D we can't do any tests that require actual GUI interaction, but we can run tests against anything that imports wx. So this opens up the \GUI\ folder to have tests added to it.

I figured that's what you were aiming for when doing that `FakeModule` stuff in that one test I commented out. Should be interesting to see how we can utilize it.

Ebag333:
>>Tweak travis
>Christ

Let's see you do anything better starting from scratch. 😄 

>It is odd... When you say save, you mean saved to the user database? As in, there's two separate Rifter fits in the sqlite database?

Yup, I checked that.  Saved to the virtual DB.

>Perhaps there's a test that is done earlier in Travis that is saving the Rifter, and then it gets to this test and saves it again. 

NOPE! :)

Again, checked for this.  When the test runs, getAllFits returns 0.

If we don't clean up after the test, this could be true.  But in this case, it's not.

```
    def test_getAllFits(DB, RifterFit, KeepstarFit):
    assert len(Fit.getAllFits()) == 0 # <- true

    DB['db'].save(RifterFit) <- ends up creating two Rifters
    DB['db'].save(KeepstarFit) <- Creates 1 Keepstar

    # For some reason in Travis this adds the first fit twice.  WHY?!?
    assert len(Fit.getAllFits()) != 0
```

>I assume Travis can handle print statements that we can then see?

It can, which is the source for a lot of the commits you're bitching about. 😁

blitzmann:
Dunno then, seems very odd. I would say test in Linux to see if it shows same functionality as int Travis,. If that doesn't get you anywhere, then eh. Do you want to investigate more, or do you think this is good to be included as is?

Also, if it's a PR for tests, brief description on what the tests do instead of "this haz tests". In this situation it's easy enough to see, but for future reference. :)

Ebag333:
>Dunno then, seems very odd. I would say test in Linux to see if it shows same functionality as int Travis,. If that doesn't get you anywhere, then eh. 

Agreed, but would be nice if you could test as well, being on Mint and all. 

>Do you want to investigate more, or do you think this is good to be included as is?

Merge it. I did that terrible work around to get it to pass Travis. We can always iterate on it.

blitzmann:
> Agreed, but would be nice if you could test as well, being on Mint and all.

I haven't been able to get tests to work on my windows machine even. Can you verify which packages I need, and how I would start the tests?

Ebag333:
All the test requirements should be under requirements_test.  Not much that's special.

In pyCharm, should just be right click and run. Right now you have to run the tests individually, hopefully soon will have a fix for that (assuming that PR that you closed gets opened and merged :) ). 

Let me know on Slack and if I have my laptop handy I can help you get it running on Windows, Mint should be the same if you have pyCharm installed.

Ebag333:
Also, requirements_test has a bunch of Travis stuff (code coverage and stuff). That isn't strictly necessary for local tests.

blitzmann:
@Ebag333 once Travis approves the update from dev, please squash and merge. :)

blitzmann:
yay it worked

PR #1087, opened 2017-03-31
calculateModifiedAttributes cleanup and general improvements

Ebag333:
@blitzmann this is something I could use your help with.

I'm trying to re-enable calculating projected fits (it was hard disabled by you), and generally clean it up and make the logic more sane.  I've done a first pass at this, but I can't get the test I wrote to return a value other than the base value.  Want to take a look at it and give me a hand here, maybe learn a bit about writing tests while you're at it? :)


Ebag333:
Gonna fold this back into the original PR, they're too intertwined...

Issue #1096, opened 2017-04-02
Capital modules can be fit on subcapital ships


BartMassey:
## Bug Report

Pyfa allows fitting a capital module to a subcapital ship. In-game (at least on Singularity) this is not allowed.

### Expected behavior:

Pyfa refuses to allow fitting a Capital F-S9 Regolith Compact Shield Extender to a Dominix Navy Issue.

### Actual behavior:

Pyfa allows fitting a Capital F-S9 Regolith Compact Shield Extender to a Dominix Navy Issue.

### Detailed steps to reproduce:

1. Get a Dominix Navy Issue hull.

2. Fit a Capital F-S9 Regolith Compact Shield Extender.

### Release or development git branch? Please note the release version or commit hash:

1.28.1 Stable - YC119.3 1.0

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):

Debian Testing (Squeeze)

blitzmann:
These items don't have any attributes to restrict their use to certain ships. I think I remember this coming up at some point, and it was thought that this is actually something CCP hardcodes into the game, and isn't available via item attributes. Basically if it's a capital-sized module (defined as having a volume >4k), then they throw an error in the client stating that you can't do that.

I think we left it alone at that point as it's not something that's very detrimental to pyfa functioning correctly. However, I'll double check the issues to see if I can find any mention of this. Could you provide a screenshot of the error the client gives?

@Ebag333 could you maybe chat with Larrikin and see if you can get a firm answer on how this is handled? As stated, it's been discussed before that this is based on the volume of the module. Would be nice to know even if we choose not to implement it. :)

BartMassey:
![capital-shield-on-subcap](https://cloud.githubusercontent.com/assets/117151/24588520/c50c8eda-177e-11e7-925a-50d3911136f6.png)

blitzmann:
hahaha :)

Well, based on current assumptions (capital module is defined as having a volume >= 4000), we can easily restrict capital modules from being fitted onto subcap hull:

https://github.com/blitzmann/Pyfa/commit/3274130c78676a3bf1b533ebde8a95072b89a131

This prevents pyfa from even adding the module to the fit. To be able to add it, but cause some sort of warning, would be a lot more work without any real benefit methinks.

I want to dig a bit further to make sure this is indeed how things are defined before implementing it though.

blitzmann:
[...]

PR #1097, opened 2017-04-02
Restrict cap modules to only cap ships

blitzmann:
This is a quick fix for #1096 - adds a check in `Module.fits()` to make sure we aren't trying to put a cap module on a sub cap ship.

### DIFF
--- a/eos/saveddata/module.py
+++ b/eos/saveddata/module.py
@@ -200,6 +200,10 @@ class Module(HandledItem, HandledCharge, ItemAttrShortcut, ChargeAttrShortcut):
             return self.owner.modules.index(self)
 
     @property
+    def isCapitalSize(self):
+        return self.getModifiedItemAttr("volume", 0) >= 4000
+
+    @property
     def hpBeforeReload(self):
         """
         If item is some kind of repairer with charges, calculate
@@ -420,7 +424,7 @@ class Module(HandledItem, HandledCharge, ItemAttrShortcut, ChargeAttrShortcut):
 
         # EVE doesn't let capital modules be fit onto subcapital hulls. Confirmed by CCP Larrikin that this is dictated
         # by the modules volume. See GH issue #1096
-        if (fit.ship.getModifiedItemAttr("isCapitalSize", 0) != 1 and self.getModifiedItemAttr("volume", 0) >= 4000):
+        if (fit.ship.getModifiedItemAttr("isCapitalSize", 0) != 1 and self.isCapitalSize):
             return False
 
         # If the mod is a subsystem, don't let two subs in the same slot fit
--- a/eos/saveddata/module.py
+++ b/eos/saveddata/module.py
@@ -418,6 +418,11 @@ class Module(HandledItem, HandledCharge, ItemAttrShortcut, ChargeAttrShortcut):
         if isinstance(fit.ship, Citadel) and len(fitsOnGroup) == 0 and len(fitsOnType) == 0:
             return False
 
+        # EVE doesn't let capital modules be fit onto subcapital hulls. This seems to be dictated by the modules volume
+        # See GH issue #1096
+        if (fit.ship.getModifiedItemAttr("isCapitalSize", 0) != 1 and self.getModifiedItemAttr("volume", 0) >= 4000):
+            return False
+
         # If the mod is a subsystem, don't let two subs in the same slot fit
         if self.slot == Slot.SUBSYSTEM:
             subSlot = self.getModifiedItemAttr("subSystemSlot")
--- a/eos/saveddata/module.py
+++ b/eos/saveddata/module.py
@@ -418,8 +418,8 @@ class Module(HandledItem, HandledCharge, ItemAttrShortcut, ChargeAttrShortcut):
         if isinstance(fit.ship, Citadel) and len(fitsOnGroup) == 0 and len(fitsOnType) == 0:
             return False
 
-        # EVE doesn't let capital modules be fit onto subcapital hulls. This seems to be dictated by the modules volume
-        # See GH issue #1096
+        # EVE doesn't let capital modules be fit onto subcapital hulls. Confirmed by CCP Larrikin that this is dictated
+        # by the modules volume. See GH issue #1096
         if (fit.ship.getModifiedItemAttr("isCapitalSize", 0) != 1 and self.getModifiedItemAttr("volume", 0) >= 4000):
             return False
 


blitzmann:
> Add a preference for this. We spoke about this ages ago, and both agreed that this shouldn't be hard and fast. Our new fitting engine and preferences pane would be a good spot for this.

Did we? I remember discussing it at some point, but I think it revolved around the new eos, not specifically pyfa. Do you remember why we decided to allow it in the current version of pyfa?

I dunno about adding a pref... should we then have a pref to allow triage module to be fit to a Avatar? Or to bypass `maxGroupFitted`? I feel if EVE completely disallows it, it's a moot point to allow it in pyfa. Now, if someone really wants to add a cap module to a rifter, they can override the `volume` property... I don't see a usecase for this, other than to get a lol-tanked Rifter

I'm not gonna be able to write a test for it because I don't have tests working yet. :((((((

Ebag333:
>Did we? I remember discussing it at some point, but I think it revolved around the new eos, not specifically pyfa. Do you remember why we decided to allow it in the current version of pyfa?

>I dunno about adding a pref... should we then have a pref to allow triage module to be fit to n Avatar? Or to bypass maxGroupFitted? I feel if EVE completely disallows it, it's a moot point to allow it in pyfa. Now, if someone really wants to add a cap module to a rifter, they can override the volume property... I don't see a usecase for this, other than to get a lol-tanked Rifter

We did, because you didn't think that this was a thing. So I sent you a screenshot of a rookie ship fitting with all cap mods. :)

I say a preference because:
1) We allow it currently
2) Most of our fitting restrictions are soft, not hard.  You can do lots of things EVE won't allow you to do.
3) Yes, lolfits are hilarious.  But there's also use for it in edge cases, like testing battleship rebalancing, industrial command ships, etc.
4) We have people who...uh...use pyfa in ways that are non-traditional.  They aren't building fits that players will ever fly. :)

>I'm not gonna be able to write a test for it because I don't have tests working yet. :((((((

It shouldn't be difficult to get tests working, and this is a good one because it's a _very_ simple test.  If you break it out into a `@property` it becomes even more simple.

Like I said before, get with me on Slack and I'll help you get tests working.

blitzmann:
> We did, because you didn't think that this was a thing. So I sent you a screenshot of a rookie ship fitting with all cap mods. :)

Ah yeah, I think I remember. And I think I said we couldn't do it because EVE doesn't give us any way of restricting that viam dogma attributes. I must have overlooked the volume thing or not had known about it until after. Maybe. Was a while ago lol

> Most of our fitting restrictions are soft, not hard. 

I would say it's the other way around. Most are hard and we have a few exceptions (like enabling more drones than the user can use). And I don't really see the balancing use-case - why on earth would you put a cap mod on a BS to test out balancing for that BS? But this can be extended to anything - what if we wanted to test to see what other ships do with a triage module, or with multiple Industrial Cores. Again, developing out multiple preferences for these restrictions seems like wasted time. These are hard caps and should remain so - if the user wants to test these lolfits out, well, that's why I added the attribute overrides a while back: to test out balancing and enable oddities. I would say 95% of the user base doesn't care, and would prefer to have pyfa work the way EVE does.

@BartMassey thoughts on how we should handle this? I want pyfa to remain versatile, but also don't want to provide a preference for every thing that we do. I think the attribute overrides sufficiently takes care of it, but want a third opinion.

Ebag333:
>Again, developing out multiple preferences for these restrictions seems like wasted time. These are hard caps and should remain so - if the user wants to test these lolfits out, well, that's why I added the attribute overrides a while back: to test out balancing and enable oddities. I would say 95% of the user base doesn't care, and would prefer to have pyfa work the way EVE does.

I think you missed where I ninja edited and added that we don't need multiple preferences, just a single one. ;)

BartMassey:
I think that even a single preference switch that overrides all the hard fitting constraints would be a lot of work to implement. But I'd be fine with it if @Ebag333 wanted to implement it :-) and if the constraints (including this one) were hard by default as they are currently. Again, I hate to, uh, see random people less skilled than myself make unflyable lolfits by accident.

In general, it would be nice if Pyfa popped up some kind of explanation when a user tried to violate a hard fitting constraint. Just silently refusing can be a bit frustrating sometimes. It's not obvious why you can't add an extra Ancillary Armor Repairer, for example, even though you can add as many Ancillary Shield Boosters as you like. As a beginning fitter, I thought it was a Pyfa bug and had to spend some time researching it. (Why the CCP description of the AAR doesn't mention this restriction is a mystery to me. I should probably file a bug with CCP.) But adding this notification subsystem would be a lot of work too, and I'm not willing to do it :-(. Besides, I'm wildly off-topic now. Apologies.

I think @Ebag333's plan of adding a synthetic property is a good one. This at least makes it possible to figure out why Pyfa won't let you do what you're trying to do.

I can't think of a great compromise offhand. Maybe just hide the capital modules when fitting a subcap? This would make it pretty hard to screw up, but you could still import a fit from the clipboard or something if you wanted to do it anyway. But then you couldn't use Pyfa to check other folks' fits for legality, as I have occasionally done. So I don't know.

I don't know.

Ebag333:
>I think that even a single preference switch that overrides all the hard fitting constraints would be a lot of work to implement. 

```
@property
isCapitalSize(self):
    if setting.strictFitting() is True:
        return False

    if self.size > 4000:
        return True
    else:
        return False
```

Doesn't have to be complicated. :) 

An explanation of fitting restrictions would be a nice addition to it, and could be iterated on.

BartMassey:
@Ebag333 Having the preference just override this one hard constraint seems more confusing than helpful to me, but it's likely no one but the three of us will ever care about it anyway. My personal opinion is that we shouldn't make the preference visible until we have all the checks we can find in place, but I'm fine if all y'all wanna do it some other way. I do feel strongly that the preference should default to hard: wanting Pyfa to not enforce fit validity is by far the special case IMHO.

Ebag333:
>Having the preference just override this one hard constraint seems more confusing than helpful to me, but it's likely no one but the three of us will ever care about it anyway. 

Definitely not the case. I know somebody who uses pyfa to build non-player ships, and then there's a whole bunch of people I regularly have discussions with on how to rebalance things, on the Tweet Fleet Slack.

As far as adding one at a time, there's no reason why we can't go back in at a whole bunch if we can find them. If we implement this house is we are changing the behavior, my concern is that we be able to have a way for users to go back to the old Behavior. I don't mind defaulting to the new behavior and strict restrictions, but they should be able to back that out if they so desire.

blitzmann:
This was touched upon in a discussion in Tweetfleet Slack. It would be nice to have it be a toggle on the fit (instead of hidden in global prefs where users may forget about it).

I'll put an issue out for it, but for now I'm going to merge this one in.

PR #1102, opened 2017-04-04
Cleanup conversions.

Ebag333:
Conversions prevents us from doing builds, because it uses a very hacky method to extract dicts out of various .py files.

This consolidates it down, gets rid of the hacky method, and overall simplifies things.


blitzmann:
God damn this is ugly and unmaintainable being in a single large file.

> Conversions prevents us from doing builds

cx_freeze works fine for me. What builds are you talking about? pyinstaller? I assume it's due to getting a list of files? Do migrations work with builds? If so, do what we do with migrations so that these can remain split out into their own files.

As a reminder and fyi, I haven't yet had time to mess with pyinstaller, so I dunno if that is even a thing yet.

Ebag333:
>Do migrations work with builds? If so, do what we do with migrations so that these can remain split out into their own files.

We could, but then it still wouldn't work. :)

Migrations don't work, and that code isn't very handsome either.  You'll need to figure out a way to handle that if you want the automatic builds to work.

Ebag333:
RIP

PR #1103, opened 2017-04-05
Fit Calculation Code Cleanup

blitzmann:
This PR does a few things:

* Removes `dirtyStorage` since it's literally never used (was legacy fleet boost parameter)
* Developed a simpler way of determining which type of calc we're currently iterating through. Instead of using a combination of `targetFit` (used with both projected and command fits), and `withBoosters` (name coming from legacy fleet system, but retro-fitted with command bursts to signal command calculations), we now use a simple ENUM which tells us what kind of calculations we're in.
* Updates / adds various comments about calculations.

Actual functionality shouldn't change with this PR (and I have tested what I could think of - I'm seeing identical behavior with `master`). There was one bug that I noticed that I will make an issue for - currently in master, but I believe it's low-priority and somewhat of an edge case.

In addition to this, I created a basic doc as I was working through the logic jotting down my notes. Definitely need to expand on this though: https://github.com/pyfa-org/Pyfa/wiki/Fit-Calculation-Notes

Gonna be testing this out a few more days with general usage to make sure it works as intended before merging in.


Ebag333:
Few thoughts.

One, recursions should be broken out of the `calculateModifiedAttributes` method.  One, recursion isn't actually calculating modified attributes, it's doing something different.  Two, by handling recursion inside this method it makes it _vastly_ more complicated.

See:
```
    def calculateFitAttributes(self, targetFit=None, withBoosters=False):
        """
        This method handles recursion through the chain of fit, projected fits, and command fits.  We start from our current fit (self), then recurse up through
        the chain of projected fits.  Because this is self recursive, each level will then recurse further up the chain.  We then do the same for command fits
        (bursts or formerly gang links).

        In practice if we have Fit B projected onto Fit A, and Fit C projected onto Fit B, it will end up looking like:
        - Find Fit A, process projected fits
        - Find Fit B, process projected fits
        - Find Fit C, no projected fits.
        - Calculate Fit C
        - Project Fit C onto Fit B
        - Calculate Fit B
        - Project Fit B onto Fit A
        - Calculate Fit A

        :param targetFit:

        :param withBoosters:
        :return:
        """
        pyfalog.debug("Starting fit calculation.")

        # Follow the chain, if we find any fits not calculated, recalc them all.
        if not self.validateFitChainCalculated():
            self.clearFitChainCalculated()

        if withBoosters:
            # Recalc ships projecting onto this fit
            for projected_fit in self.projectedFits:
                if projected_fit.getProjectionInfo(self.ID).active:
                    if projected_fit is self:
                        # If fit is self, don't recurse
                        self.calculateModifiedFitAttributes(targetFit=self)
                    else:
                        projected_fit.calculateFitAttributes(withBoosters=withBoosters, targetFit=self)

            for command_fit in self.commandFits:
                if command_fit.getCommandInfo(self.ID).active:
                    if command_fit is self:
                        # If fit is self, don't recurse
                        self.calculateModifiedFitAttributes(targetFit=self)
                    else:
                        command_fit.calculateFitAttributes(withBoosters=withBoosters, targetFit=self)

        self.calculateModifiedFitAttributes()

        if targetFit:
            self.calculateModifiedFitAttributes(targetFit=targetFit)
```

This is a half a dozen lines of code, and makes it crystal clear where recursion is happening.

`withBoosters` is better used as an identifier if we want to recurse through the fits or not.  We never really needed it to know anything about targetFit, projected, or command fit.

There's lots of times when we _don't_ need to recurse, can just recalc the one local fit which is a lot faster.

More thoughts later....

Ebag333:
I'm not sure why you did two PRs that both change the same method.  That makes it a lot harder to review. 🤕 

So one issue I have is that this is basically the same performance as Master/Dev. While there's a bit of speedup when switching fits, just opening fits in general is no faster.

From a quick smoketest (opening 22 different fits and adding up the recalc times), the branch I worked on is 14% faster than this one.  So it's not near as performant.

I'm not sure why you went and redid the work I had already done.  I understand you didn't like 4 lines of code, but.....

blitzmann:
> I'm not sure why you did two PRs that both change the same method.

> So one issue I have is that this is basically the same performance as Master/Dev. 

methinks you missed this: "Actual functionality shouldn't change with this PR"

This is clean/clarification of the function, my aim wasn't to change anything functional with it.

One thing that we may want to look into, as you suggested, is breaking out the function into two. I've looked into doing this before, but didn't see a clean way of doing it, and regardless we would still need the same logic to perform calculations (unless we can clean up some loops to be more performant)

Additionally, they are two topics. Ones clean up without functional changes, the other has functional changes. Don't look at both as a whole.

> There's lots of times when we don't need to recurse, can just recalc the one local fit which is a lot faster

I'm aware. During my review of the code I found that already handle that (maybe), but haven't confirmed how well it works. See https://github.com/pyfa-org/Pyfa/blob/216dac068d9b68d2446dac61c27c8f41812f76b0/eos/saveddata/fit.py#L730

How it's supposed to work: if the fit is calculated, then skip _local_ calculations. At this time we still need the other stuff so that we project onto the target fit.

The aim for this PR was to clarify the logic of that function, so that we could better discuss further _incremental_ improvements, but those discussions don't belong in this PR.

Ebag333:
>methinks you missed this: "Actual functionality shouldn't change with this PR"

methinks you missed that actual functionality doesn't change in mine either. 😉 
(Or it's not supposed to at least.)

>How it's supposed to work: if the fit is calculated, then skip local calculations. At this time we still need the other stuff so that we project onto the target fit.

That's just the thing.  The calling function (way back when recalcs is called) already says whether it wants with or without boosters.  By throwing that away, you'll just be doing full recalcs more often.

blitzmann:
> The calling function (way back when recalcs is called) already says whether it wants with or without boosters. 

You mean the `withBoosters` stuff? That is not actually doing anything anymore, hence why it's removed in this PR.

Or are you suggesting we add flags to let the `calculateModifiedAttributes` function know whether or not to also run projections and boosters? I don't think that would work, as whenever we recalc we want to make sure we we also _apply_ the projections and boosters, not just calc them. But I haven't put any brain cycles towards it. You have a shit load of open PRs now, so I'm sure you have _something_ to say about it in one of those. I'll reserve this PR until those are evaluated and accepted / denied.

PR #1105, opened 2017-04-05
Do not recalc fits on tab switch

### LINKED ISSUE #987: Do not recalculate fit when simply switching tabs
Right now, if you have two fits open, switching between the two recalculates them every time. This was initially done intentionally to allow fits that were affected by other fits to update correctly (fit A is projecting onto fit B, fit A has a change, when we switch over to Fit B we recalc to ensure we always get that change)

However, most of the times fits don't have anything to do with one another. We should be smarter about this, as this can cause pyfa to be perceived as slow.

Plan:
1) When a fit is calculated, it sets a flag `__calculated = True`., 
2) When the fit is calculating, it goes through all fits that it affects (is the command booster or projected fit), and set those fits as `__calculated == False`. That way, when we switch back to the other fit, we can pick up that it's no longer calculated.

Seems easy, but the whole fit loading / switching / and calculation workflow is kind of a mess. It will require more work and research into what data is loaded at what time, what events are being used, etc, and extensive testing to make sure that calculations are being updated correctly and are sound.

blitzmann:
Fixes #987

This is a simple change that will make user interaction with pyfa a bit less of a headache, and continues with our search for performance and efficiency gains. 

Once a fit is calculated, a flag is set to signal at such (`self.__calculated`). This has always been a thing, but we only used it in very limited situations. Now we check to see if a fit is already calculated before we `recalc()` when switching fits.

We did not do this previously since fit B, which projects onto fit A, may have been modified, and switching to fit A would require us to recalc to be sure we have the latest data. This is solved by iterating through a fits projectedOnto and boostedOnto fits, and setting their calculated status to `False`. Now when fit B changes, fit A is no longer marked as calculated, and thus is recalced when switched to. If nothing in fit B changes, nothing happens. 

This is a good first step into "smart recalcs" as @Ebag333 coined, only recalculating when needed. I don't think we can every get to 100% smart recalcs without a rewrite of the engine which is pointless at this stage, but there are sure to be places where we can implement similar behavior.


Ebag333:
This won't work.  There's no recursion, so fits changed several levels deep won't get caught.

This is what I came up with.  It could be simplified a little bit by caching a record of the fits you've cleared so you don't clear the same one twice if it's in the chain multiple times.

```
    def validateFitChainCalculated(self):
        """
        Walks up the chain for the fit and anything projected or command fits haven't been calculated

        :return:
        True if all fits are calculated, False if one (or more) is not
        """
        # print("Validating: " + str(self.name) + " Status: " + str(self.calculated))

        for projected_fit in self.projectedFits:
            if projected_fit.getProjectionInfo(self.ID).active:
                if projected_fit is not self:
                    if projected_fit.calculated is False:
                        return False

                    projected_calculated = projected_fit.validateFitChainCalculated()
                    if projected_calculated is False:
                        return False

        for command_fit in self.commandFits:
            if command_fit.getCommandInfo(self.ID).active:
                if command_fit is not self:
                    if command_fit.calculated is False:
                        return False

                    command_calculated = command_fit.validateFitChainCalculated()
                    if command_calculated is False:
                        return False

        # print("For (" + str(self.name) + ") returning chain have been calculated")
        return True

    def clearFitChainCalculated(self):
        """
        Walks up the chain for the fit and clear the calculated flag on any projected or command fits

        :return:
        True if all fits are calculated, False if one (or more) is not
        """
        # print("Clearing calculated flag on: " + str(self.name))

        for projected_fit in self.projectedFits:
            if projected_fit.getProjectionInfo(self.ID).active:
                if projected_fit is not self:
                    projected_fit.clearFitChainCalculated()

        for command_fit in self.commandFits:
            if command_fit.getCommandInfo(self.ID).active:
                if command_fit is not self:
                    command_fit.clearFitChainCalculated()

        self.calculated = False
```


You also don't remove the `self.calculated` flag on `clear()`, so this isn't going to work correctly.

blitzmann:
> This won't work. There's no recursion, so fits changed several levels deep won't get caught.

According to my knowledge of how calcs work when dealing with projection and command fits, this will work. We don't need to recurse (... word?) down further since we only ever look at max two levels of fit (`local < projected < command` being the largest chain that all affect one another). I wrote up a quick wiki note about how these thinks interact - here I exploit the fact that projected effects don't affect others, and command effects don't affect others. So we're able to do this a bit more efficiently than recursing down the projected and command trees

Granted, it's not 100% smart - as I mentioned, it's probably never going to be. Take the situation:

` local A < projected fit B < command fit C < command fit D`

If you change `D`, we don't have to change `A` or `B` since there aren't any affecting effects that travel down that far. `D` will only ever affect `C`. So you can change `D` and then switch to `A` without a problem.

However, if you change `D`, `C` gets flagged for recalc on switch. Part of `C`'s recalc includes resetting `B` and `A` (since we don't know _why_ `C` recalc'd, we just know it needs to, and thus need to be sae and reset it's dependents). So if you click down the line from `C` to `B` / `A`, you'll noticed those fits recalc even though they aren't technically affected by `D`'s change (or a simple switch recalc for `C`). In this case, it's extra calcs, but at this time without knowing why something is recalcing, this is the best solution.

It's hard for me to really explain this. Try ti out, if you notice any actual calculation bugs as a result of fits not recalcing after command / projection changes, please let me know with fits and config. It may totally be the case that this is incorrect, but all tests and combinations that I've tried show idential stats to `master`. :D

Ebag333:
I get your explanation.  I'm pretty sure we've run into scenarios before where deeper recursion was happening that that, however.  I'll give you that I don't remember them off the top of my head. ;)

That just goes back to my other point, on the other PR, however.  The branch I've been working on is 14% faster....and it's doing full recursion (which is probably unnecessary beyond a certain depth).

blitzmann:
>  The branch I've been working on is 14% faster.

There were a lot of problems with that branch that I didn't like, and I for one did not see 14% performance improvements as stated and shown in Slack (and I tested on the three different computers eventually, even my shitty laptop). So I dunno how you are getting that number. There's definitely potential for improvement, no doubt, but that isn't the topic of this PR, so lets both try to stay on topic for once. ;)

> I'm pretty sure we've run into scenarios before where deeper recursion was happening than that, however. I'll give you that I don't remember them off the top of my head.

Not to my knowledge. AFAIK, there's a cap at how far effects go. Which makes sense as a game design thing - I'm sure CCP doesn't want to calculate a chain of 40 projected fits all the time when calculating just one fit. This is an assumption that pyfa has had for a while, and we haven't had any issues with users reporting that their 4' fit isn't applying projections. 

If you run into a situation in which the calc is incorrect, by all means. I want pyfa to be correct :D

PR #1107, opened 2017-04-05
Clean up fit service

Ebag333:
Performs a bunch of cleanup.

1. Redirects everything to use `getFit` rather than rolling their own DB fetch.
2. Moves redundant code into a single location (mostly `db.commit()` into the recalc).
3. Removes `refreshFit` because very little used it, and what did could use `recalc` just fine.
4. Makes what's requesting a full recalc (fit + boosters + command bursts) vs what's requesting just a local recalc a bit more sane.
5. `recalc` defaults to a full recalc now.


Ebag333:
RIP

PR #1108, opened 2017-04-05
Fit calculated flag

Ebag333:
Since this was already done.....

Handles a fit.calculated flag so we don't recalc fits if not needed.


Ebag333:
Bah. Wrong repo.

PR #1110, opened 2017-04-05
Refactor caclculateModifiedAttributes.

Ebag333:
Refactors caclculateModifiedAttributes.

Few things this does.
1. Moves a bunch of stuff that was done in separate steps together (greatly simplifies the logic).
2. Breaks out the recurse into it's own function.  We can handle it at a more granular level.
3. Don't delete `self`. We've learned that _this is a bad thing to do_.  When we are done with our shadow copy, we put things back like we found them because that's what you do.
4. Greatly simplifies the idea of `targetFit`.  Because we're handling recursion at a higher level, and because we handle our shadow copy early on, this makes it very easy to follow the logic.

Earlier tests have shown that this is a fairly significant performance increase, around the 10-15% mark.

### DIFF
--- a/eos/saveddata/fit.py
+++ b/eos/saveddata/fit.py
@@ -23,6 +23,7 @@ from itertools import chain
 from math import sqrt, log, asinh
 
 from sqlalchemy.orm import validates, reconstructor
+from sqlalchemy.exc import InvalidRequestError
 
 import eos.db
 from eos import capSim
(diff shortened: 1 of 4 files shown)


blitzmann:
This is a valiant effort, but it doesn't work quite the way you think. It may very well be possible to split out the calc function, but the way that you have it set up in this is incorrect according to my tests. You can't just calc and apply the remote fits in this order. As far as I'm aware, the calc function is currently the way it is specifically because it needs to be - I've tried multiple time to split it out with bad results. 

Closing this as branch currently breaks calcs. If it's due to a missing conditional like the other PR, feel free to let me know and I can re-evaluate. Whether you want to try to fix it is up to you - I would recommend leaving the calc function the way it is. It's not a pretty function, but sometimes to get it to work is not so pretty. :/

PR #1112, opened 2017-04-05
Fit calculated flag

Ebag333:
Handles a fit.calculated flag so we don't recalc fits if not needed.

### DIFF
--- a/eos/saveddata/fit.py
+++ b/eos/saveddata/fit.py
@@ -183,6 +183,15 @@ class Fit(object):
         self.__character = char
 
     @property
+    def calculated(self):
+        return self.__calculated
+
+    @calculated.setter
+    def calculated(self, value):
+        # todo: brief explaination hwo this works
+        self.__calculated = value
+
+    @property
     def ship(self):
         return self.__ship
 
@@ -392,7 +401,7 @@ class Fit(object):
         self.__droneVolley = None
         self.__droneYield = None
         self.__ehp = None
-        self.__calculated = False
+        self.calculated = False
         self.__capStable = None
         self.__capState = None
         self.__capUsed = None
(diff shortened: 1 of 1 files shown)


blitzmann:
Unfortunately, this is fundamentally flawed here: https://github.com/pyfa-org/Pyfa/pull/1112/files#diff-12319c39afbe21ce25350ff3158c7d24R736

These leads to incorrect calculations for my fits which have command boosts (didn't test projected). We cannot stop calculation here, otherwise the fits that apply projected / command effects on the local fit will not be run. This is why we have this line which was removed:

https://github.com/pyfa-org/Pyfa/pull/1112/files#diff-12319c39afbe21ce25350ff3158c7d24L727

Furthermore, I'm not sure if there is a reason to recursively loop through the things that project / command onto the local fit... I get what you're trying to do, but this is probably better handled from a top-down approach (#1105) instead of looking up the chain to determine if we need to recalc, which makes sense because if we are recalcing our local fit, there's nothing our local fit does that affects remote effects on it, so I don't see why we should reset their calc flags... (maybe I'm missing something?)  

Regardless, if a fit is already calc'd it should be be handled by the line which was removed, and allows them to still apply their effects. (again, this isn't a claim that I thoroughly tested with my recent frolics through this function)

Ebag333:
>These leads to incorrect calculations for my fits which have command boosts (didn't test projected). We cannot stop calculation here, otherwise the fits that apply projected / command effects on the local fit will not be run. This is why we have this line which was removed:

And this is why when you take 4 or 5 things that all touch the same method, and all interlock, and all work closely together, and break them apart, it isn't easy to do.  Nor is putting them back together (which we've seen with several merge problems now).

It's a simple thing that was missed. ` and not targetFit`.

It was working in the original branch. It's working in my branch (since I fixed it).  Since you're determined to not use this, I'm not going to bother fixing it here.

blitzmann:
Hey now, you're the one who was interested in refactoring a complex function (which isn't really about complexity as it is more about keeping in mind various conditionals). Sorry that you have issues with juggling your branches and what goes in/out of them. Also doesn't address my second point of the recursive looping. If you know of a condition in which that is required, please let me know. :)

PR #1114, opened 2017-04-05
Refactor Pricing

### LINKED ISSUE #678: [Feature Request] Prices on implants and boosters.

Ebag333:
The main highlights:
- Removes the market price cache, and instead we append our price object to all items.  Now the item itself is the cache.
- Move the price thread to the price service. Makes more sense.
- Clean up a bunch of the logic
- Remove unused functions, and merge functions where possible.
- Make a number of things more object oriented (no taking a list of objects to get a list of IDs just to convert back to objects).
- Adds a new full pricing pane that breaks down value more discretely.
- Move the old pricing pane to a minimal version, so folks can still use that if they so desire.
- Adds pricing column to implants/boosters, because why not? Fixes #678

Unlike the prior PR (#1038), this leaves the thread handling intact.  We rework it a little to better use the caching of pricing on items, but we're still using it to let our panel know that it needs an update.

### DIFF
--- a/eos/gamedata.py
+++ b/eos/gamedata.py
@@ -23,6 +23,7 @@ from sqlalchemy.orm import reconstructor
 
 import eos.db
 from eqBase import EqBase
+from eos.saveddata.price import Price as types_Price
 
 try:
     from collections import OrderedDict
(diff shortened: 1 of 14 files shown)


Ebag333:
Hmm. Weird, it's pulling from yours but I pushed it to mine.....

PR #1115, opened 2017-04-05
Pricing Refactor

### LINKED ISSUE #678: [Feature Request] Prices on implants and boosters.

Ebag333:
The main highlights:
- Removes the market price cache, and instead we append our price object to all items.  Now the item itself is the cache.
- Move the price thread to the price service. Makes more sense.
- Clean up a bunch of the logic
- Remove unused functions, and merge functions where possible.
- Make a number of things more object oriented (no taking a list of objects to get a list of IDs just to convert back to objects).
- Adds a new full pricing pane that breaks down value more discretely.
- Move the old pricing pane to a minimal version, so folks can still use that if they so desire.
- Adds pricing column to implants/boosters, because why not? Fixes #678

Unlike the prior PR (#1038), this leaves the thread handling intact.  We rework it a little to better use the caching of pricing on items, but we're still using it to let our panel know that it needs an update.

@blitzmann I think I might have solved a _very_ annoying issue that's been plaguing us a long time.  It's in this PR because I would occasionally roll into DB rollbacks.
from `eos.db.init`
`sd_lock = threading.Lock()`

I believe that this should be
`sd_lock = threading.RLock()`

>The main difference is that a Lock can only be acquired once. It cannot be acquired again, until it is released. (After it's been released, it can be re-acaquired by any thread).
>An RLock on the other hand, can be acquired multiple times, by the same thread. It needs to be released the same number of times in order to be "unlocked".
>Another difference is that an acquired Lock can be released by any thread, while an acquired RLock can only be released by the thread which acquired it.

This would explain the rollbacks we get from time to time.  It would also solve the issues with rollsbacks you saw when testing price fetching as a background service.

### DIFF
--- a/eos/db/__init__.py
+++ b/eos/db/__init__.py
@@ -68,7 +68,7 @@ else:
     saveddata_meta = None
 
 # Lock controlling any changes introduced to session
-sd_lock = threading.Lock()
+sd_lock = threading.RLock()
 
 # Import all the definitions for all our database stuff
 # noinspection PyPep8
--- a/eos/gamedata.py
+++ b/eos/gamedata.py
@@ -23,6 +23,7 @@ from sqlalchemy.orm import reconstructor
 
 import eos.db
 from eqBase import EqBase
+from eos.saveddata.price import Price as types_Price
 
 try:
     from collections import OrderedDict
@@ -437,6 +438,39 @@ class Item(EqBase):
                 return True
 
         return False
+
+    @property
+    def price(self):
+        try:
+            if not hasattr(self, "__price"):
+                self.__price = types_Price(self.ID)
+
+            # Get the price from the DB
+            price = eos.db.getPrice(self.ID)
+
+            if price:
+                if self.__price.time > price.time:
+                    # The object is newer than the DB, update the DB.
+                    eos.db.add(self.__price)
+                    eos.db.commit()
+
+                if self.__price.time < price.time:
+                    # DB object is newer than local object, update the local object
+                    self.__price = price
+            else:
+                pyfalog.debug("Unable to fetch item price from database.")
+
+            return self.__price
+
+        except Exception as e:
+            # We want to catch our failure and log it, but don't bail out for a single missing price tag.
+            pyfalog.error("Failed to get price for typeID: {0}", self.ID)
+            pyfalog.error(e)
+            if not self.__price.price:
+                self.__price.price = 0
+            self.__price.failed = True
+
+        return self.__price
 
     def __repr__(self):
         return "Item(ID={}, name={}) at {}".format(
--- a/eos/saveddata/price.py
+++ b/eos/saveddata/price.py
@@ -21,6 +21,9 @@
 import time
 
 from sqlalchemy.orm import reconstructor
+from logbook import Logger
+
+pyfalog = Logger(__name__)
 
 
 class Price(object):
@@ -29,7 +32,6 @@ class Price(object):
         self.time = 0
         self.price = 0
         self.failed = None
-        self.__item = None
 
     @reconstructor
     def init(self):
--- a/gui/boosterView.py
+++ b/gui/boosterView.py
@@ -43,9 +43,12 @@ class BoosterViewDrop(wx.PyDropTarget):
 
 
 class BoosterView(d.Display):
-    DEFAULT_COLS = ["State",
-                    "attr:boosterness",
-                    "Base Name"]
+    DEFAULT_COLS = [
+        "State",
+        "attr:boosterness",
+        "Base Name",
+        "Price",
+    ]
 
     def __init__(self, parent):
         d.Display.__init__(self, parent, style=wx.LC_SINGLE_SEL | wx.BORDER_NONE)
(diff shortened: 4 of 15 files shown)


blitzmann:
Just ran it for a quick test, works well and doesn't block execution like the other one, so that makes me happy. :) I'll dig into the code as I have time, but on first glance this looks good. 

Two things:

image

image


This looks funky when you resize the window (bottom one happened when I maximized the window when it was like the first one). I'm not sure if it's because of a particular thing you're doing in that pane or how it's laid out, or if it's always been a thing but not is manifesting since we have graphics stacked down there.

Not too worries about it

2) What happened to my little reload symbol in the price column when fetching the price ;_;

> This would explain the rollbacks we get from time to time.

I don't get rollbacks "from time to time". 0.o

Ebag333:
>I don't get rollbacks "from time to time". 0.o

Your pyfa also runs on unicorn rainbow farts. :P

blitzmann:
I mean, honestly, the only times I've seen weird SQLAlcehmy exceptions recently is when I'm running your branches during evaluations. Not hating, just stating that it's been a looong while since I've seen it on master or dev, and it's usually when I'm working on a feature that does low-level DB shit. I also saw it sometimes with the html auto export crap. I feel if this was a common occurrence for the users we would know about it, since it tends to prevent pyfa from actually running IIRC

One may argue that if you're getting rollbacks due to locks, find where it's locking and fix the underlying code. But if there's no other issues with this then I don't really care lol. We'll have to keep it in the back of our minds once implemented that if users start to experience weird stuff to revert the change and see if that works.

Ebag333:
>I feel if this was a common occurrence for the users we would know about it, since it tends to prevent pyfa from actually running IIRC

We don't have a ton of stuff in threads that commit straight to the DB, they usually pass it back (which is largely how we avoid it).

The explanation makes sense though, if a thread (or callback) and the main thread try and hit the DB at the same time, one of those is going to rollback.

A bunch of the PRs you're complaining about were trying to do things in a thread with the database.  There's no reason we _shouldn't_ be able to do so, but it clearly wasn't locking properly otherwise the thread wouldn't get a rollback (or the main thread getting one if the thread was using it).

Ebag333:
You might want to roll this back.  You'll likely see periodic session issues.

The problem, at its root, is that SQLAlchemy doesn't really handle threaded applications by default.  There are features to handle them built in, but we don't use any of those.

And then we pretend threads don't exist at all by reusing the same session everywhere (something you're not really supposed to do).

And then we force set it to pretend that threads don't exist using undocumented switches.

I'm working on revamping the thread handling, it's not a difficult task, just a little tedious making sure it's all talking and working properly.

Ebag333:
Alternatively, could just wait until I'm done then pull in the fixes. Probably easier than rolling it back.

blitzmann:
I've already fixed it. You beat me to the comments to tell you. 

I'm not sure what you're talking about with threads - the root of the problem, at least for me, lies in the fact that this PR didn't properly couple the Price object that was created with the item with the database record. So there were times when we would try to insert a price for type id 1234, and it already existed (I was able to do this by invalidating the time so that we would try to refresh the price).

Ebag333:
Your fix addresses it, but it's still a threading issue. ;)

It's actually similar to the way I was thinking about handling it (thought in the callback functions rather than in gamedata):
`if self.__price is not None and getattr(self.__price, '_sa_instance_state', None):`

What's happening is that the price is getting updated in a different thread than our main.  So we're getting collisions between the two.  It already exists because  the two threads are running simultaneously.

blitzmann:
> What's happening is that the price is getting updated in a different thread than our main. So we're getting collisions between the two. It already exists because the two threads are running simultaneously.

I haven't been able to reproduce any threading issues, do you have a method to test this with? The price has always been updated in another thread, but it used to be committed in the main thread. If there truly is an issue with committing to the database on another thread, then a simple fix should be moving the `db.commit()` back to the main thread one price updates have finished, yeah?

Ebag333:
I don't really.

It probably just takes a merge to deal with it.  I ran into that because I could get a stacktrace occasionally when  deleting fits (love the new error window, have found at least a half dozen on going issues because of it).

Basically changing:
```
def remove(stuff):
    removeCachedEntry(type(stuff), stuff.ID)
    with sd_lock:
        # Merge the object to make sure we don't already have it in session
        saveddata_session.delete(stuff)
    commit()
```

to 
```
def remove(stuff):
    removeCachedEntry(type(stuff), stuff.ID)
    with sd_lock:
        # Merge the object to make sure we don't already have it in session
        saveddata_session.merge(stuff)
        saveddata_session.delete(stuff)
    commit()
```

I suspect that something similar with pricing would fix it, but did not mess around with it too much.  Your method works (mostly).  I killed the state checking, because it would end up being true when it really shouldn't have been and deleted prices constantly (turn on debug mode and you'll see what I mean).

I still need to tinker with the price column as occasionally it doesn't refresh properly.  Went in and fixed the item compare price column too.

You can cherry pick that PR if you so desire.....

PR #1118, opened 2017-04-06
Migration refactor

Ebag333:
Refactors migration to let it work properly with builds.

```
[2017-04-06 18:30:30.384000] DEBUG: service.prefetch: Run database migration.
[2017-04-06 18:30:30.411000] INFO: eos.db.migration: Database upgrade in progress.
[2017-04-06 18:30:30.413000] INFO: eos.db.migration: Upgrading database from version 21 to upgrade22
[2017-04-06 18:30:30.426000] INFO: eos.db.migration: Running upgrade.
[2017-04-06 18:30:31.418000] INFO: eos.db.migration: Database upgrade complete.
```

Adds logging.

Removes about 2 or 3 different hacks to replace with a single hack.  So now 66% less hacky.

Maintains same structure as before.  

Conversions couldn't be done exactly this way, because the goal of the refactors was to maintain the same structure.  Plus, this requires a hack, conversion refactor has none.


blitzmann:
Couldn't you just maintain a list of the the upgrades? And then use the import method we use for effects to iterate through that list and import, and viola, bob's your uncle? Also hacky and not as automatic as currently, but it'll work while keeping the structure.

Also, you keep saying builds, but never confirmed: is this for pyinstaller?

Ebag333:
>Couldn't you just maintain a list of the the upgrades? And then use the import method we use for effects to iterate through that list and import, and viola, bob's your uncle? Also hacky and not as automatic as currently, but it'll work while keeping the structure.

This keeps the current structure and design.  So yes, we could, but then we have to use the hacky method used for imports, this keeps the hacks to a minimum.

>Also, you keep saying builds, but never confirmed: is this for pyinstaller?

Most build programs will have the same problem with it.  It won't know about these imports, and won't include them.

cxFreeze works differently than most of the rest because it's literally just extracting and running the whole repo.

But yes, I'm using pyinstaller to test with.

Ebag333:
RIP

Issue #1119, opened 2017-04-10
Delete All for Damage Pattern and Target Resists


Hiwashi:
There's an Import/Export for Damage Patterns and Target Resists but no way to delete all of them at once, making it very time consuming to remove everything if you have several patterns.

Ebag333:
@blitzmann can pull this PR: https://github.com/Pyfa-fit/Pyfa/pull/22

blitzmann:
Thanks @Ebag333 

Fixed it up, tested, included. @Hiwashi, should be available in the 1.29 release. :)

blitzmann:
@Hiwashi out now with v1.29 :)

PR #1120, opened 2017-04-11
"Delete all" buttons for Prices, Target Resists, Damage Profiles

blitzmann:
For #1119, includes work by @Ebag333

### DIFF
--- a/eos/db/saveddata/queries.py
+++ b/eos/db/saveddata/queries.py
@@ -332,11 +332,25 @@ def getDamagePatternList(eager=None):
     return patterns
 
 
+def clearDamagePatterns():
+    with sd_lock:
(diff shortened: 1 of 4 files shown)


Ebag333:
`DefaultDatabaseValues.importRequiredDefaults()`

This was in there for a reason.  You'll cause stacktraces without it, because you're literally nuking every profile. ;)

blitzmann:
No I'm not

https://github.com/pyfa-org/Pyfa/pull/1120/files#diff-8106b35ba0b00a4e0d2caa22bdaf2685R337

Ebag333:
Fair point.

I still think it's a good idea to run the defaults, but yeah, it won't stacktrace.

PR #1124, opened 2017-04-12
Allow system wide installation

StinGer-ShoGuN:
Rewrite/continuation of #1080.

I made the requested changes. And there are more to come.

For instance, on Linux systems, applications user data should be stored in `$XDG_DATA_HOME` which should default to ` ~/.local/share/[app-name]` if the variable is not set. See [here for details](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

In addition, I would like to be able to move the `imgs` directory to `/usr/share` as well, `$XDG_DATA_DIRS` eventually. When I'm talking about moving, I mean being able to run Pyfa as it is meant to be now and add the possibility to run it with the `imgs` directory in this new path.

#### Unicode issue
I've been trying to reproduce the issue with `os.path.join` and unicode strings. And the only error I have is with this:
```python
os.path.join(unicode(os.path.expanduser("~"), sys.getfilesystemencoding()), "çÁè")
```

But, for instance:
```python
os.path.join(os.path.expanduser("~"), "Développement")
```
does work like a charm... In fact, I've got errors when I am encoding one or more of the strings given to `os.path.join` with `unicode` (or prefixing it with `u""`) and I also give a string with unicode chars in it but without "unicoding" it.

At least on Linux... Not tested yet on Windows.

### DIFF
--- a/config.py
+++ b/config.py
@@ -96,7 +96,9 @@ def defPaths(customSavePath):
     # The database where the static EVE data from the datadump is kept.
     # This is not the standard sqlite datadump but a modified version created by eos
     # maintenance script
-    gameDB = os.path.join(pyfaPath, "eve.db")
+    gameDB = getattr(configforced, "gameDB", gameDB)
+    if not gameDB:
+        gameDB = os.path.join(pyfaPath, "eve.db")
 
     # DON'T MODIFY ANYTHING BELOW
     import eos.config
--- a/service/settings.py
+++ b/service/settings.py
@@ -263,7 +263,7 @@ class HTMLExportSettings(object):
 
     def __init__(self):
         serviceHTMLExportDefaultSettings = {
-            "path"   : config.pyfaPath + os.sep + 'pyfaFits.html',
+            "path"   : config.savePath + os.sep + 'pyfaFits.html',
             "minimal": False
         }
         self.serviceHTMLExportSettings = SettingsProvider.getInstance().getSettings(


Ebag333:
This won't work for unicode. Reach out to me, I have a fix I've been testing.

StinGer-ShoGuN:
Hmmm, I'm not sure to understand what you're talking about... ❓

Ebag333:
What you're doing won't work for unicode paths.

First off, what's in place today won't work, it's wrong.  It especially is broken in environments with multiple locales (say a English Windows environment with a Russian folder path).

Secondly, certain unicode chars are handled specially.  Mostly looking from a Windows perspective, characters like `é` can be converted by the built in OS (even if you're in a vanilla English setup), while Russian/Chinese/etc are not.

Finally, there are issues with `expanduser` and Windows.  There are `wontfix` issues with handling unicode, long story short is that it simply won't ever work in Windows for unicode paths.


I've actually already fixed this over at my fork (Pyfa.fit).  I dunno if @blitzmann will back port it or not.

blitzmann:
Thanks for the changes. As stated in the previous PR, I'm totally cool with the having a configforced value to change DB path, as well as saving the html file to the save path by default (which probably should have happened a long time ago)

As for the unicode stuff, I'm not too concerned about it tbh - any locale issues would be, I feel, pretty edge case. And we're not doing anything funky here with paths other than simply using one that the user provides in config forced.

@Ebag333, are you talking about this bit? http://stackoverflow.com/questions/23888120/an-alternative-to-os-path-expanduser I ran into this post a while back, but it included imports that I wasn't sure would work with the builds... but since this is windows only, should be easy enough to update... Will have to take another look

The `$XDG_DATA_HOME` thing - we know this is the preferred place to store application data. It's pretty low priority for pyfa to do this - we would have to migrate data files, or come up with a system that supports existing users as well as uses the new method in new installations. It's something that I want to do (yay standards); someone expressed interest in doing this at one time, dunno how far they got. I may look into it again soon.

All that said, @StinGer-ShoGuN, this is ready to merge, correct?

StinGer-ShoGuN:
> All that said, @StinGer-ShoGuN, this is ready to merge, correct?

Yes.

#### XDG standards
I know there's a python module somewhere to handle local storage, user storage etc... handling Mac, Linux and Windows. I found it once but I can't find it any more... 😒
My idea would be to use this new module and make something similar on Windows as well, storing data in `%USERPROFILE%/AppData/Local/` (or `Roaming` actually, who knows...).

The entire thing without breaking existing installs, obviously.

But, one thing after another... 😃 

#### Unicode
Yeah well, it's broken in Python 2, I think there is no need to develop workarounds. The best move would be to move to Python 3. 👿

I've had another good reason also to migrate Pyfa to Python3, but I can't remember which one... 👼

Ebag333:
> are you talking about this bit?

That's part of it.  Look at the PR I submitted on my branch. :)

>Yeah well, it's broken in Python 2, I think there is no need to develop workarounds. The best move would be to move to Python 3. 👿

Unfortunately that won't happen any time soon.  Would have to move to wxPython Phoenix, and there's still not feature parity.

We'd be better off just using a new GUI entirely.....

StinGer-ShoGuN:
> We'd be better off just using a new GUI entirely.....

HAHA HAHA worse... Who said Qt ?!?!! 👀

Ebag333:
>HAHA HAHA worse... Who said Qt ?!?!! 👀

I dunno about that. EVE uses QT. 

But I was more thinking of something like Flask and React. That's all a pipedream though anyway.

blitzmann:
EVE uses QT? You're shitting me...

@StinGer-ShoGuN If you want to tackle XDG, go for it, under the restriction that current users aren't affected - moving and manipulating user data usually scares the shit out of me since any number of things can go wrong. But I'm all for it if we're just going to do it for new installs (for now).

Get with me on Slack if you can so we can discuss a general process on how to move forward with it. It really shouldn't be a difficult thing to do but I want to make sure all bases are covered. :)

Anyway, thanks for the PR :)

Issue #1148, opened 2017-05-10
Error Window when trying to open a fitting


AlhiraK:
```
OS version: 	Windows-8-6.2.9200
Python: 	2.7.10
wxPython: 	3.0.2.0
SQLAlchemy: 	1.0.5
Logbook: 	1.0.0
pyfa version: 1.29.0 Stable - YC119.5 1.0
pyfa root: C:\Program Files (x86)\pyfa
save path: C:\Users\Tom\.pyfa
fs encoding: mbcs

EXCEPTION: This session is in 'prepared' state; no further SQL can be emitted within this transaction.

  File "C:\Program Files (x86)\pyfa\library.zip\gui\builtinViews\fittingView.py", line 78, in fitSelected
    view.fitSelected(event)
  File "C:\Program Files (x86)\pyfa\library.zip\gui\builtinViews\fittingView.py", line 300, in fitSelected
    self.slotsChanged()
  File "C:\Program Files (x86)\pyfa\library.zip\gui\builtinViews\fittingView.py", line 473, in slotsChanged
    self.generateMods()
  File "C:\Program Files (x86)\pyfa\library.zip\gui\builtinViews\fittingView.py", line 431, in generateMods
    fit = sFit.getFit(self.activeFitID)
  File "C:\Program Files (x86)\pyfa\library.zip\service\fit.py", line 253, in getFit
    for fitP in fit.projectedFits:
  File "C:\Program Files (x86)\pyfa\library.zip\eos\saveddata\fit.py", line 261, in projectedFits
    return [fit for fit in self.__projectedFits.values() if not fit.isInvalid]
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\ext\associationproxy.py", line 257, in __get__
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\ext\associationproxy.py", line 302, in _new
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\ext\associationproxy.py", line 460, in __call__
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\attributes.py", line 237, in __get__
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\attributes.py", line 578, in get
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\strategies.py", line 528, in _load_for_state
  File "<string>", line 1, in <lambda>
[...]

Issue #1146, opened 2017-05-10
Error when adding warp stab


treky928:
# Submit a bug report bug report or feature request

Here you can inform pyfa developers of potential bugs or suggest features / improvements to the project. Please check
to make sure that the bug hasn't been reported or feature requested before submitting. If you have general questions
about the project and want to reach out to the developers personally, please check out out our [Slack]
(https://pyfainvite.azurewebsites.net/).

---

## Bug Report

```
OS version: 	Windows-8-6.2.9200
Python: 	2.7.10
wxPython: 	3.0.2.0
SQLAlchemy: 	1.0.5
Logbook: 	1.0.0
pyfa version: 1.29.0 Stable - YC119.5 1.0
pyfa root: C:\Program Files (x86)\pyfa
save path: C:\Users\Finch\.pyfa
fs encoding: mbcs

EXCEPTION: This Session's transaction has been rolled back by a nested rollback() call.  To begin a new transaction, issue Session.rollback() first.

  File "C:\Program Files (x86)\pyfa\library.zip\gui\builtinViews\fittingView.py", line 333, in appendItem
    populate = sFit.appendModule(fitID, itemID)
  File "C:\Program Files (x86)\pyfa\library.zip\service\fit.py", line 515, in appendModule
    self.recalc(fit)
  File "C:\Program Files (x86)\pyfa\library.zip\service\fit.py", line 1153, in recalc
    fit.calculateModifiedAttributes()
  File "C:\Program Files (x86)\pyfa\library.zip\eos\saveddata\fit.py", line 779, in calculateModifiedAttributes
    item.calculateModifiedAttributes(self, runTime, False)
  File "C:\Program Files (x86)\pyfa\library.zip\eos\saveddata\character.py", line 258, in calculateModifiedAttributes
    skill.calculateModifiedAttributes(fit, runTime)
  File "C:\Program Files (x86)\pyfa\library.zip\eos\saveddata\character.py", line 404, in calculateModifiedAttributes
    effect.handler(fit, self, ("skill",))
  File "C:\Program Files (x86)\pyfa\library.zip\eos\effects\accerationcontrolskillabmwdspeedboost.py", line 10, in handler
    level = container.level if "skill" in context else 1
  File "C:\Program Files (x86)\pyfa\library.zip\eos\saveddata\character.py", line 337, in level
    if self.character.name == "All 5":
[...]

Issue #1145, opened 2017-05-10
Error message on startup


peldar:
### Expected behavior:
No error messages

### Actual behavior:
Installed latest release, selected a fit, got this "pyfa error" popup window with this text:

```
OS version: 	Windows-8-6.2.9200
Python: 	2.7.10
wxPython: 	3.0.2.0
SQLAlchemy: 	1.0.5
Logbook: 	1.0.0
pyfa version: 1.29.0 Stable - YC119.5 1.0
pyfa root: C:\Program Files (x86)\pyfa
save path: C:\Users\olaf_\.pyfa
fs encoding: mbcs

EXCEPTION: (sqlite3.InterfaceError) Cursor needed to be reset because of commit/rollback and can no longer be fetched from.

  File "C:\Program Files (x86)\pyfa\library.zip\gui\implantView.py", line 142, in fitChanged
    self.update(stuff)
  File "C:\Program Files (x86)\pyfa\library.zip\gui\display.py", line 293, in update
    self.refresh(stuff)
  File "C:\Program Files (x86)\pyfa\library.zip\gui\display.py", line 252, in refresh
    newImageId = col.getImageId(st)
  File "C:\Program Files (x86)\pyfa\library.zip\gui\builtinViewColumns\state.py", line 84, in getImageId
    elif isinstance(stuff, Implant) and stuff.character:
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\attributes.py", line 237, in __get__
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\attributes.py", line 578, in get
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\strategies.py", line 528, in _load_for_state
  File "<string>", line 1, in <lambda>
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\strategies.py", line 598, in _emit_lazyload
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\query.py", line 2398, in all
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\loading.py", line 84, in instances
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\util\compat.py", line 199, in raise_from_cause
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\loading.py", line 65, in instances
[...]

PR #1154, opened 2017-05-11
Do not commit price objects to database from within thread

### LINKED ISSUE #1151: SQLAlchemy Session Error
Was adding ammunition to cargo of a Jackdaw fit via right click > add 1000; following error occurred on adding Mjolnir Rage Rockets. After closing error pane, was impossible to select any other item in the market section. Close Pyfa and reopen to fix.

Haven't been able to reproduce.
```
OS version: 	Windows-7-6.1.7601-SP1
Python: 	2.7.10
wxPython: 	3.0.2.0
SQLAlchemy: 	1.0.5
Logbook: 	1.0.0
pyfa version: 1.29.0 Stable - YC119.5 1.0
pyfa root: D:\CCP\Pyfa
save path: C:\Users\Alisdair\.pyfa
fs encoding: mbcs

EXCEPTION: This session is in 'prepared' state; no further SQL can be emitted within this transaction.

  File "D:\CCP\Pyfa\library.zip\gui\contextMenu.py", line 138, in handler
    menuHandler.activate(context, selection, i)
  File "D:\CCP\Pyfa\library.zip\gui\builtinContextMenus\cargoAmmo.py", line 35, in activate
    sFit.addCargo(fitID, typeID, 1000)
  File "D:\CCP\Pyfa\library.zip\service\fit.py", line 688, in addCargo
    cargo = es_Cargo(item)
  File "<string>", line 4, in __init__
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\state.py", line 306, in _initialize_instance
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\util\langhelpers.py", line 60, in __exit__
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\state.py", line 303, in _initialize_instance
  File "D:\CCP\Pyfa\library.zip\eos\saveddata\cargo.py", line 40, in __init__
    self.__itemModifiedAttributes.overrides = item.overrides
  File "D:\CCP\Pyfa\library.zip\eos\gamedata.py", line 272, in overrides
    overrides = eos.db.getOverrides(self.ID)
  File "D:\CCP\Pyfa\library.zip\eos\db\saveddata\queries.py", line 496, in getOverrides
    return saveddata_session.query(Override).filter(Override.itemID == itemID).all()
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\query.py", line 2398, in all
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\query.py", line 2515, in __iter__
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\query.py", line 2528, in _execute_and_instances
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\query.py", line 2519, in _connection_from_session
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\session.py", line 882, in connection
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\session.py", line 887, in _connection_for_bind
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\session.py", line 313, in _connection_for_bind
  File "C:\python-2.7.10\lib\site-packages\sqlalchemy\orm\session.py", line 202, in _assert_active
```
---

blitzmann:
Trying to manipulate and save objects from one thread in another will cause sqlalchemy session exceptions. This brings it back to the original method of modifying the price object in another thread, but committing it in the main thread, avoiding the issues.

Will have to look for a better way later (not really supposed to share session objects at all with other threads), but this should work in the interim. 

Should resolve #1151, #1148, #1146, #1145

### DIFF
--- a/service/price.py
+++ b/service/price.py
@@ -151,7 +151,6 @@ class Price(object):
             priceobj.time = time.time() + REREQUEST
             priceobj.failed = True
 
-
     @classmethod
     def fitItemsList(cls, fit):
         # Compose a list of all the data we need & request it
--- a/service/price.py
+++ b/service/price.py
@@ -127,9 +127,6 @@ class Price(object):
                 priceobj.time = time.time() + VALIDITY
                 priceobj.failed = None
 
-                # Update the DB.
-                db.commit()
-
                 # delete price from working dict
                 del priceMap[typeID]
 
@@ -141,9 +138,6 @@ class Price(object):
                 priceobj = priceMap[typeID]
                 priceobj.time = time.time() + TIMEOUT
                 priceobj.failed = True
-
-                # Update the DB.
-                db.commit()
 
                 del priceMap[typeID]
         except:
@@ -157,8 +151,6 @@ class Price(object):
             priceobj.time = time.time() + REREQUEST
             priceobj.failed = True
 
-            # Update the DB.
-            db.commit()
 
     @classmethod
     def fitItemsList(cls, fit):
@@ -207,6 +199,7 @@ class Price(object):
             except Exception as e:
                 pyfalog.critical("Callback failed.")
                 pyfalog.critical(e)
+
             db.commit()
 
         if waitforthread:


Ebag333:
This isn't an issue in my branch. Think I already pointed out the difference. 

Also, a lot of things can trigger that error. This isn't the only thing that will cause it.

blitzmann:
>This isn't an issue in my branch. Think I already pointed out the difference.

Which differences? And I wouldn't be so sure. These errors are pretty elusive and only happen after the object changes build up, took me a while to figure out a way to reproduce (clear cache, open lots of fits). Taking a quick peak at your latest release, this most certain does happen on yours. ¯\_(ツ)_/¯

> Also, a lot of things can trigger that error. This isn't the only thing that will cause it.

Not sure what you're getting at here :/ Yeah, a lot of things can trigger SQLAlchemy errors. However, in these particular cases, since the session was being modified in a separate thread, the session got jacked beyond repair. It doesn't break immediately, which is why I never caught it before. This is what is causing them as best as I can tell.

PR #1194, opened 2017-05-29
Pricing pane 1158

petosorus:
I have gone with the context menu solution to set options.
I have a list of labels in the ContextMenu definition, but I'm not sure this is the right way/right place. I put it there to always have the list in the same order.

I haven't used the bitmask thing yet but I could, as you wish.

The full pane looks self-explaining to me, but the minimal could benefit from a breakdown in a tooltip.
I also grouped price options as they were grouped in the pane (implants are grouped with boosters).

### DIFF
--- a/gui/builtinContextMenus/priceOptions.py
+++ b/gui/builtinContextMenus/priceOptions.py
@@ -11,6 +11,7 @@ class PriceOptions(ContextMenu):
     def __init__(self):
         self.mainFrame = gui.mainFrame.MainFrame.getInstance()
         self.settings = PriceMenuSettings.getInstance()
+        self.optionList = ["Ship", "Modules", "Drones", "Cargo", "Character"]
 
     def display(self, srcContext, selection):
         return srcContext in ("priceViewFull", "priceViewMinimal")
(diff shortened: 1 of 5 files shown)


blitzmann:
This looks great, and I think it will help alleviate the issues with the current pricing pane. The tooltip is something that can be worked on later as a separate item (I may tackle this soon). I've tested with the ful pane, not yet with the minimal pane, so once I do I can get it merged in. Great work! 

> I have a list of labels in the ContextMenu definition, but I'm not sure this is the right way/right place. I put it there to always have the list in the same order.

I was thinking the same. Feels odd to put it there, but it works and there isn't really any other place to put it.

> I haven't used the bitmask thing yet but I could, as you wish.

Meh. It's a nice to have, but no need at the moment, and it would do effectively the same thing as this. I'm cool with the way it's done.

Only one thing - it fails Travis with the following formatting issues:

```
gui/builtinStatsViews/priceViewFull.py:163:5: E303 too many blank lines (3)
gui/builtinContextMenus/priceOptions.py:6:1: F401 'service.price.Price' imported but unused
gui/builtinContextMenus/priceOptions.py:7:1: F401 'service.settings.ContextMenuSettings' imported but unused
gui/builtinContextMenus/priceOptions.py:50:1: E305 expected 2 blank lines after class or function definition, found 1
```

I can fix these if needed, but if you have some time go for it. :)

petosorus:
Yeah, I forgot to check the format things on commit in pycharm >.>

blitzmann:
Excellent work @petosorus! Thanks for the PR :)

Issue #1149, opened 2017-05-10
'decimal' codec can't encode character u'\uff00' 


jasconius:
## Bug Report

App crashes when dragging certain items into fit

### Expected behavior:

Item goes into fit

### Actual behavior:

Crash with error report

### Detailed steps to reproduce:

Was using a Tormenter. Tried to add ancillary armor rep to empty low slot, crash

### Release or development git branch? Please note the release version or commit hash:

Release (latest as of May 10)

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):

Mac OS 10.12.4

### Other relevant information:

EXCEPTION: 'decimal' codec can't encode character u'\uff00' in position 4: invalid decimal Unicode string

  File "/Applications/pyfa.app/Contents/Resources/lib/python2.7/site-packages.zip/gui/builtinViews/fittingView.py", line 114, in OnData
  File "/Applications/pyfa.app/Contents/Resources/lib/python2.7/site-packages.zip/gui/builtinViews/fittingView.py", line 204, in handleListDrag

blitzmann:
@jasconius does this happen with other modules? Did you drag it from the market list, or from something like cargo? Unfortunately I can't seem to reproduce :(

Can you send me your `/.pyfa` directory zipped to sable.blitzmann@gmail.com? It contains logs, database, and user settings that might prove useful in determining what's happened.

jasconius:
Yeah it happened with some pulse lasers too. It's only partially reproducible.

The modules were dragged from the market tab. Into both empty slots and slots with stuff already in it (to replace)

Only way to recover was to reboot, and after a reboot, sometimes you could repeat the exact same action and it would not error, but sometimes it would. Some sort of race condition...?

I'll see if I can find the directory

blitzmann:
It's just a very weird error. It doesn't make any sense to me why it would happen sometimes and not other. is it always the same error?

Are you using the `mac-deprecated` or `mac` build? I tried both, couldn't reproduce, but will try some more since you state it's not easy to reproduce. It could also possibly be related to other issues that may have been fixed, though I am skeptical of that since it's not a database issue. Very strange. I'll await your files - if you have trouble finding them, let me know. :)

ghesla:
I received a very similar error this morning when dragging a script into the cargo hold of a fit. 
Mac OS 10.12.5

> OS version: 	Darwin-16.6.0-x86_64-i386-64bit
[...]

PR #1196, opened 2017-05-31
updated drag and drop behavior to not trigger a suspected wx bug on Mac

jasconius:
This patch addresses #1149 

I believe that wx 3.x for newer Mac operating systems has a memory buffer error allowing garbage data to be passed around through wx gui event handlers, resulting in a critical error that makes the Mac client unusable without a restart

I shortcut this problem by avoiding the wx-provided event handler data variable by using a simple python static variable, and apply the fix where I could find similarly implemented DragAndDrop events

I will note that this implementation is both quick and dirty, as my goal here was to influence the code base as little as possible. So I even leave in a conditional check `if self.GetData():` which is a call to a inherited wx function, because I did not want to make any assumptions about what this code was actually doing.

Therefore, you will notice where I assign this static variable, I am still allowing the data to also be assigned to the wx object, as to avoid the chances of causing an unintentional regression in behavior.

I leave the task of studying the functionality of `GetData()` and determining if it, and the related code, should be replaced to the maintainers of Pyfa.

This patch has received limited testing on the latest public version of Mac OS as of the date of this PR, only to confirm that the issue in question no longer occurs. No other testing has been conducted.

### DIFF
--- a/gui/boosterView.py
+++ b/gui/boosterView.py
@@ -37,7 +37,8 @@ class BoosterViewDrop(wx.PyDropTarget):
 
     def OnData(self, x, y, t):
         if self.GetData():
-            data = self.dropData.GetText().split(':')
+            dragged_data = DragDropHelper.data
+            data = dragged_data.split(':')
             self.dropFn(x, y, data)
         return t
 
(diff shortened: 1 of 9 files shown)


blitzmann:
@jasconius I've tested on Windows and this seems to work without a hitch. Excellent work on diagnosing the issue and coming up with a fix! 

Could you update the branch to fix the travis errors? After that, I can merge in :)

jasconius:
should be all set now

PR #1243, opened 2017-06-30
Fix loki effects

wendazhou:
Fix effects for loki subsystems:

- defensive subsystem "Adaptive Defense Node" should apply overheat bonus to both shield and armor repair amount bonus
- offensive subsystem "Launcher Efficiency Configuration" should apply missile velocity and missile explosion velocity bonuses to charge instead of launcher.


blitzmann:
Hey @wendazhou , just saw this pull request. Thanks for the fixes!

As an aside, I did a pass today and fixed a few effects as well, but just so happens these ones slipped through. Gonna have to do another final pass right before release. :)

Issue #1261, opened 2017-07-28
Allow use of Variations menu on modules in cargo hold


pcd1193182:
When you right-click on a module in the actual fit, you can use the Variations menu to switch between meta versions of the same module. It would be nice if you could do that in the cargo hold as well; this frequently comes up when importing a fit from my fleet's doctrine where there are refits in the cargo hold. Often these refits are designed assuming all excellent skills or no excellent skills, when in practice most people have a mix; the ability to meta up and down the modules that I can't use with just a right click would be very helpful.

It might also be nice to be able to see which of those modules are and aren't usable without having to drag them into the fit, but I'm not sure what a good UI for that would necessarily look like.

burnsypet:
Seems like a really useful feature and have made a pull request.

pcd1193182:
Closed, fixed!

PR #1262, opened 2017-07-30
Issue #1261 show variations menu for cargo items

burnsypet:
https://github.com/pyfa-org/Pyfa/issues/1261

Added logic to condition statement in metaSwap.py to accommodate cargoItem context. Will replace existing cargo item with selected item and retain the original stack amount.

Seems like a fairly useful feature and I didn't realize how much I wanted it until I read the ticket. Tested on Windows 7, unable to test on any other platforms.

Screenshot:

cargo_metaswap


blitzmann:
Excellent, thank you! And sorry about the late reply, haven't found much time to keep up with things in July/Aug, but working through them now :D

This will be included in into tonight's release :)

Issue #1241, opened 2017-06-30
select effective dps issue


oity:
## Bug Report

In Firepower panel
After clicking on 'toggle to mining yeld' icon and click again to went back to 'Firepower view', right click to select 'target resist' stop working, you have to close and reopen pyfa to get it back working properly.
Selecting other fits doesen't fix the issue 

pyfa 1.29.4 Stable - YC119.5 1.0
on Ubuntu 17.04

EVE Data: 	1142796
Python: 		2.7.13
wxPython: 	3.0.2.0
SQLAlchemy: 	1.0.14
matplotlib: 	 2.0.0
---

blitzmann:
This is a known problem, and has been for a while. I think it's due to a bad binding or something. It does work, but you have to be very precise in where you spawn the context menu, and it's very very finicky. Will take a look more closely as I have time :)

blitzmann:
This was fixed a little while ago, forgot to mark it as such :)

PR #1268, opened 2017-08-16
Issue #1241 Bind contextHandler to children of contentPanel

burnsypet:
Changes to statsPane.py:
Added new method applyBinding which loops through all the children
of the contentPanel passed to it and binds contextHandler to each of them

Changes to miningyieldViewFull.py:
Call to new method applyBinding

---

#1241  

It looks like the child elements were never having any binding applied to them so I have created a new method to do this. 

There still appears to be an issue with the miningyieldViewFull panel displaying the context menu for firepowerViewFull but I believe that this is because binding occurs on the panel at load time when statsPane.py runs for all of the views. We would probably need something to remove the binding from this panel when it is toggled onto.

I'm not really an expert though, what do you think @blitzmann ?

### DIFF
--- a/gui/builtinStatsViews/miningyieldViewFull.py
+++ b/gui/builtinStatsViews/miningyieldViewFull.py
@@ -122,6 +122,8 @@ class MiningYieldViewFull(StatsView):
         self.parent.views.append(view)
         # Get the TogglePanel
         tp = self.panel.GetParent()
+        # Bind the new panel's children to allow context menu access
+        self.parent.applyBinding(self.parent, tp.GetContentPane())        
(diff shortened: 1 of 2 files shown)


blitzmann:
Excellent, and thank you for tackling this! 

>I'm not really an expert though,

I don't think anyone except Robin Dunn is an expert in wxPython, to be frank lol. While functionality is still a bit off on the mining panel (still have to click between the icons), this absolutely seems to fix the issue of it not spawning for the firepower panel, which is the pain point. I'm not too worried about the question on if the context menu should or shouldn't spawn in the mining panel - that's a discussion we can have at a later point. :)

Issue #1264, opened 2017-08-03
Crash


Pizzakira:
I was modifying implants as some fits were open with projections to some spaceship in between.

OS version: 	Windows-8-6.2.9200
Python: 	2.7.10
wxPython: 	3.0.2.0
SQLAlchemy: 	1.0.5
Logbook: 	1.0.0
pyfa version: 1.30.0 Stable - YC119.7 1.0
pyfa root: C:\Program Files (x86)\pyfa
save path: C:\Users\Vinx\.pyfa
fs encoding: mbcs

EXCEPTION: C++ assertion "menu" failed at ..\..\src\common\wincmn.cpp(2955) in wxWindowBase::PopupMenu(): can't popup NULL menu

  File "C:\python-2.7.10\lib\site-packages\wx-3.0-msw\wx\_core.py", line 16766, in <lambda>
  File "C:\Program Files (x86)\pyfa\library.zip\gui\characterEditor.py", line 596, in spawnMenu
    self.PopupMenu(menu)
  File "C:\python-2.7.10\lib\site-packages\wx-3.0-msw\wx\_core.py", line 11144, in PopupMenu

Pizzakira:
I could reproduce the issue. That happen when right clicking on implants in the character editor, in implants tab, right side list.

burnsypet:
Was able to reproduce this on version 1.31.0, debug log shows:

[2017-08-17 08:19:27.033000] CRITICAL: __main__: Exception in main thread: C++ assertion "menu" failed at ..\..\src\common\wincmn.cpp(2955) in wxWindowBase::PopupMenu(): can't popup NULL menu

Will do some digging when I get home.

PR #1272, opened 2017-08-19
Issue #1264 Don't attempt PopupMenu if getMenu returned False

burnsypet:
Change to characterEditor.py:
Added logic to prevent a crash when attempting to execute
self.PopupMenu(menu) when menu was False. This would occur in the
situation where there were no saved implant sets.

---

#1264 

Fixed a scenario where spawnMenu in characterEditor.py was attempting to create a list of implant sets even when the list of implant sets contained no items which resulted in a crash. 

The new behavior if there are no implant sets is to do nothing. I've put a trace in there but that might be unnecessary and could be removed.


blitzmann:
Thanks! I noticed some other issues with the implant context menu in the character editor, went ahead and applied those fixes as well #1284

Issue #1273, opened 2017-08-19
Granting Missing skills doesn't grant prereqs.


pcd1193182:
## Bug Report
When using the "Grant Missing Skills" option, only toplevel skills are granted, not prerequisites. This results in some confusion when attempting to make skill plans and fits for various skill levels.

### Expected behavior:
Granting Missing Skills should grant all skills necessary to get the character in question to be able to fly the ship; the granted state should always be a valid, achievable state for a character.

### Actual behavior:
Only toplevel skills are granted.

### Detailed steps to reproduce:
1.  Create a new character with all 0s
2.  Open a Sabre fit
3.  Right click on the skill icon, Grant Missing Skills
4.  Note that the only granted skills are interdictors 1 and Minmatar Destroyer 5, which require between them spaceship command IV. graviton physics I, science V, power grid management V, propulsion jamming V, cpu management III, and Navigation II.

### Fits involved in EFT format (Edit > To Clipboard > EFT):
[Sabre, temp]

[Empty Low slot]
[Empty Low slot]

[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]

[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]

[Empty Rig slot]
[Empty Rig slot]

### Release or development git branch? Please note the release version or commit hash:
Release 1.31.0 stable

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
OS X 10.12.3

burnsypet:
Looks like grantMissingSkills in characterSelection.py is not going through the list of requirements recursively, it's only adding the highest level entries in self.reqs.iteritems() to the skillsMap.

In theory it is giving you all the skills you need to operate the ship but I think this is probably an error and we should be giving all the levels for the pre-requisite skills. I'll look into fixing this.

burnsypet:
Have created a pull request to fix this #1274 however when attempting to apply a large set of missing skills the application can lock up for quite a while so this may not be accepted.

blitzmann:
@pcd1193182 The functionality is already an option, see Preferences > Fitting Engine > Enforce Strict Skills

![image](https://user-images.githubusercontent.com/3904767/30569524-60955f44-9ca9-11e7-8330-77428a1e9d04.png)

[...]

PR #1274, opened 2017-08-22
Issue #1273 Add prereq skills when granting missing skills

burnsypet:
Changes to characterSelection.py:
Remove logic for deciding what skills to add from grantMissingSkills
into a new method generateMissingSkillsMap which functions similarly
to _buildSkillsTooltipCondensed and constructs a complete list of
skills including prereqs.
grantMissingSkills then iterates through this list and applies them
to the character.

---

#1273 

Moved the logic around in characterSelection.py and generating the missing skills map is now done for a new method. The list of missing skills will now include all the prerequisite skills if there are any. The only thing I am concerned about is that if a large list of skills are generated it can take quite a while to apply them to a new character. Applying 15 skills to a new blank all 0 character locked the application up for just about 4 seconds, can understand if you don't want to accept this fix if locking for this long is not desirable behavior.

### DIFF
--- a/gui/characterSelection.py
+++ b/gui/characterSelection.py
@@ -108,18 +108,42 @@ class CharacterSelection(wx.Panel):
         charID = self.getActiveCharacter()
         sChar = Character.getInstance()
 
-        skillsMap = {}
-        for item, stuff in self.reqs.iteritems():
-            for things in stuff.values():
-                if things[1] not in skillsMap:
-                    skillsMap[things[1]] = things[0]
-                elif things[0] > skillsMap[things[1]]:
-                    skillsMap[things[1]] = things[0]
+        skillsMap = self.generateMissingSkillsMap(self.reqs, skillsMap={})
 
         for skillID, level in skillsMap.iteritems():
             sChar.changeLevel(charID, skillID, level, ifHigher=True)
 
         self.refreshCharacterList()
+
+    def generateMissingSkillsMap(self, reqs, currItem="", tabulationLevel=0, skillsMap=None):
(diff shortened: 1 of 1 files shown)


pcd1193182:
This may be a silly question, but why aren't we using the map that's already been built for the tooltip? Or at least reusing the code to generate it? Your new function appears to be almost line-for-line identical to _buildSkillsTooltipCondensed.

burnsypet:
@pcd1193182 _buildSkillsTooptipCondensed builds a skillsMap which contains the skill name and level, but sChar.changeLevel() which is used to change the characters level appears to require a skill id. 

A possible solution to avoid duplicating this code would be to use the existing skillsMap and interate through it generating a new array of levels and ids as part of grantMissingSkills(), however I don't really know enough about how Pyfa works to say if there is a common function for getting the skill id based on it's name.

blitzmann:
@burnsypet thanks for the contribution, however as stated in the issue, this should already be a thing if the user has the option set. It's if very inefficient, as you've noticed; I would like to clean it up at some point and devise a way to make it quicker, but for now it's what we have (see https://github.com/pyfa-org/Pyfa/blob/73dfa25ef063807a5df22dc0c06ccf88b691d7cd/eos/saveddata/character.py#L361 for the code on this). I would be open to hearing thoughts on how it can be improved upon, but for now it is what it is. I'll be closing this issue since functionality already existed - can reconsider as new data / conversations merits. :)

burnsypet:
That's no problem I wasn't aware this was a feature already.

Issue #1289, opened 2017-09-20
Grant Missing Skills (when strict skill levels is on) misses a few


blitzmann:
While looking into #1273, I noticed that, while not exactly the same issue, there is some sort of issue revolving around the grant missing skills option.

Copy all 0 character and apply to this fit:

```
[Sabre, Sabre]

Damage Control II
Nanofiber Internal Structure II

Medium Azeotropic Restrained Shield Extender
5MN Y-T8 Compact Microwarpdrive
Initiated Compact Warp Scrambler
Medium Shield Extender II

Prototype Cloaking Device I
125mm Gatling AutoCannon II
125mm Gatling AutoCannon II
125mm Gatling AutoCannon II
125mm Gatling AutoCannon II
125mm Gatling AutoCannon II
125mm Gatling AutoCannon II
Interdiction Sphere Launcher I, Warp Disrupt Probe

Small Hyperspatial Velocity Optimizer II
Small Hyperspatial Velocity Optimizer II


Warp Disrupt Probe x46
```

This is the skill list: 

![image](https://user-images.githubusercontent.com/3904767/30623710-4d89ec56-9d87-11e7-8bd6-7c40d9f6c54a.png)

Grant missing skills, these skills are still needed:

![image](https://user-images.githubusercontent.com/3904767/30623719-6346e0d0-9d87-11e7-8b2e-8acc2cf4b194.png)

This is all assuming strict skills are enabled

burnsypet:
I added some traces to characterSelection.py and it looks like grantMissingSkills is not considering skills deeper in the tree, eg. 11579 "Cloaking" is considered however it's requirement 3426 "CPU Management" is not.

_buildSkillsTooltipCondensed seems to traverse all the skill requirements, and I have made an amendment on a local version to re-use the skillsMap generated by this to populate the skillsMap used by grantMissingSkills. Will hopefully have something ready to submit tomorrow.

PR #1303, opened 2017-09-26
Change how skillsMap is generated

burnsypet:
Changes to characterSelection.py:
grantMissingSkills now uses _buildSkillsTooltipCondensed to
generate a list of skills to update
_buildSkillsTooltipCondensed now outputs level and ID rather than
just level so that it can be used by grantMissingSkills
Anything else using _buildSkillsTooltipCondensed like exportSkills
and fitChanged have been amended to extract just the level from
the skills map rather than level and ID

---

#1289 

Granting missing skills (without strict skills set) now always provides the pre-requisite skills because _buildSkillsTooltipCondensed is used to generate the skills map (which contains the requirement skills). This does not appear to have any performance impact compared to the master version which only applied the top level skills and I couldn't really think of any scenario where only applying the top level skills would be beneficial.

If you have the strict skills set in the preferences there appears to be a performance hit because we are changing a large amount of skills from level 0 to unlearned. For example setting spaceship command to level 4 changes all of these skills to unlearned:

16591, 28656, 23950, 37615, 20524, 20526, 20527, 20528, 20532, 28374, 24312, 24311, 24313, 24314, 3344, 3345, 3346, 3347, 24606, 20525, 20530, 20531, 20533, 28374, 34327, 29029, 20342, 28667, 22761, 28609, 28374, 29637, 3327

It might be worth looking at exactly what strict skills is doing to try and reduce the performance impact... but I think this submission resolves the original issue.

### DIFF
--- a/gui/characterSelection.py
+++ b/gui/characterSelection.py
@@ -108,16 +108,10 @@ class CharacterSelection(wx.Panel):
         charID = self.getActiveCharacter()
         sChar = Character.getInstance()
 
-        skillsMap = {}
-        for item, stuff in self.reqs.iteritems():
-            for things in stuff.values():
-                if things[1] not in skillsMap:
-                    skillsMap[things[1]] = things[0]
-                elif things[0] > skillsMap[things[1]]:
-                    skillsMap[things[1]] = things[0]
+        skillsMap = self._buildSkillsTooltipCondensed(self.reqs, skillsMap={})
 
-        for skillID, level in skillsMap.iteritems():
-            sChar.changeLevel(charID, skillID, level, ifHigher=True)
+        for index in skillsMap:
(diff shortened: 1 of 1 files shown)


blitzmann:
Thanks for looking into this, will review and merge as soon as I'm able :)

> For example setting spaceship command to level 4 changes all of these skills to unlearned:

If it's only changing them form 0 to unlearned (`None`), it's probably due to an issue where both evaluate to falsey or something. I don't think this is exactly the cause of the performance hit - from what I remember, it's simply the fact that generating the skill dependencies can take quite a bit of time - quick for one skill, but it adds up. Performance tuning for that stuff is an area I've wanted to look into since developing the feature.

blitzmann:
This looks good, sorry for taking a bit to get back. Thanks for the bug fix, much appreciated! :D

Issue #1315, opened 2017-10-19
energyWarfareResistance applied twice


ibudnikov:
## Bug Report
energyWarfareResistance parameter calculated correctly, but applied twice.
One place is eos/saveddata/fit.py, line 1197. Have no idea, where is other.
### Expected behavior:
Capacitor "-" should be 46.1
### Actual behavior:
Capacitor "-" is 23.2
### Detailed steps to reproduce:
energyWarfareResistance of one battery is 25%. Ship's energyWarfareResistance: 1-0.75*(1-0.25*0.869)*(1-0.25*0.571)=49.6 %. That value is calculated correctly.
Apply 1 Capital Energy Neutralizer II against this fit. Capital Energy Neutralizer drains 4400/48=91.7 GJ/s, and after the resist it should be 91.7 * 0.503 = 46.1. 
But looks like resist applied twice: 91.7 * 0.503 * 0.503=23.2 is shown.
### Fits involved in EFT format (Edit > To Clipboard > EFT):
[Revelation, Revelation]

[Empty Low slot]
Imperial Navy Capacitor Power Relay
Imperial Navy Capacitor Power Relay
Imperial Navy Capacitor Power Relay
Centum B-Type Energized Adaptive Nano Membrane
Centum B-Type Energized Adaptive Nano Membrane
Centum B-Type Energized Adaptive Nano Membrane
Centum B-Type Energized Adaptive Nano Membrane

Capital Cap Battery II
Capital Cap Battery II
Capital Cap Battery II
Imperial Navy Cap Recharger

Siege Module II
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]

Capital Capacitor Control Circuit II
Capital Capacitor Control Circuit II
Capital Capacitor Control Circuit I

Projected:
Capital Energy Neutralizer II
### Release or development git branch? Please note the release version or commit hash:
Latest release (1.32)
### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
Windows 8.1

burnsypet:
Looking at this now, there's definitely something going on here.

When entering eos.saveddata.fit.addDrain the capNeed has already been modified by the resistance and reduced from 4400 to 2114, but it is then multiplied by the resistance again reducing it to 1114 which is the amount used in the calculation and displayed on the capacitor view. 

[...]

Issue #1139, opened 2017-04-27
The Electronic Hardening Charge used in Information Command Bursts does not affect projected tracking disruption. 


NamelessNewbie:
# Submit a bug report bug report or feature request
---

## Bug Report

Projecting e.g. a Curse with a Tracking Disruptor fitted onto a turret based ship decreases the tracking and/or range of the turrets. The Electronic Hardening Charge of Information Command Bursts is supposed to lower the effect of Tracking (and Guidance) Disruptors. However, this is not the case.

### Expected behavior:

Adding an Information Command Burst with the Electronic Hardening Script should decrease the effects of Tracking Disruptors. 

### Actual behavior:

The Electronic hardening Script currently only increases the Sensor Str. It has no effect on projected TDs.

### Detailed steps to reproduce:

Fit 250mm Rails to a Vulture. Fit an Information Command Burst to it. Fit a Tracking Disruptor to a Curse. Project the Curse onto the Vulture - the Turret stats go down. Activate the Command Burst with the Electronic Hardening Charge - the Turret stats should go up again, but they stay exactly the same. 

### Release or development git branch? Please note the release version or commit hash:

pyfa 1.28.1

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):

Windows 10

blitzmann:
Thanks for the report, will definitely look into this when I get home tonight.

Without having the data at hand at the moment, this sounds like the EWAR resistance stuff that was never implemented. I have a branch that has it working halfway from when CCP started using it a few months back, but never got around to finishing it. If that is the case, I'll take a look at what's involved to finish it up in time for the next release on the 9th.

blitzmann:
I took a look at this a couple days ago. Indeed, the ewar resist code is included in the codebase - I was thinking about the remote reps impedance stuff that is yet to be implemented. From what I can tell, the calculations are usually applied correctly, just not in this particular case due to the following:

Most ewar modules have a `remoteResistanceID` attribute in order to identify which resist attribute to check against. For example, a Stasis Web has a `remoteResistanceID` of `2115`, which tells the calculations to use the destination's `stasisWebifierResistance` attribute to modify the incoming bonus. However, it seems that Tracking Disruptors don't have a `remoteResistanceID` assigned to them. I've reached out to CCP to see if this is indeed a bug on their end, and if not then how to proceed with this case. Will update when I get more info.

[...]

PR #1322, opened 2017-10-26
Stop applying neut resistance twice

burnsypet:
Energy neut/nos resistance was being applied twice when a nos or
neut is projected onto a fit. I have removed the redundant resistance
application in energyneutralizerfalloff.py and also in
energynosferatufalloff.py as they are already being applied once in
eos.saveddata.fit.addDrain.

---

#1315 

Cap warfare resistance was being applied twice when a nos or neut was projected onto a fit, because it is applied in both the falloff effect and also in the addDrain function which is called by the falloff effect. Removed the resistance calculation from the falloff effect because it calls addDrain anyway.


burnsypet:
@blitzmann looks like the Rifter test case is failing because the ship attributes have changed. I can whip up some amended test cases this week if you think it would add some value?

blitzmann:
> @blitzmann looks like the Rifter test case is failing because the ship attributes have changed. I can whip up some amended test cases this week if you think it would add some value?

no need, I just removed the test yesterday. Those tests aren't properly mocked anyway, so anytime something changes in the database, there's the possibility that it fails. Unit testing shouldn't rely on info straight from the database, and thus pyfa needs to have properly mocked tests.

Will take a look at this PR now, sorry for the delay ;)

blitzmann:
So, i'm thinking you were close, but that it's the other way around - we want to keep the ones in the effects, and get rid of the one in `addDrain()`.

The one in `addDrain()` is older code back when CCP only did resists against cap stuff. Now that they have resists against everything (remote reps, remote caps, etc), they moved over to a more robust system. To top it off, they had, at one point, two different systems for determining the resistance of a bonus - having the module have a `remoteResistanceID` that pointed to a attribute on the module determining how much to resist by, and then moving over to a resist that was on the effect itself. CCP at one time had both systems being utilized so we had to make sure both bases were covered, which might be one of the reasons why this is now a problem? See https://github.com/pyfa-org/Pyfa/issues/1139#issuecomment-298801620 for more details on this.

Gonna keep this PR open while I do some more investigating - will bring it in and make any modifications that I need, but hopefully it's simple keeping the effects the same and removing the resist stuff from `addDrain()`

burnsypet:
That's pretty interesting, just felt a bit odd that the resistance was bundled in with the falloff effect. Although to be honest now that I look at it that falloff effect doesn't really seem to be doing much else.

blitzmann:
>  just felt a bit odd that the resistance was bundled in with the falloff effect

TBH, a lot of eos feels odd because it's been the calc engine behind pyfa since it's inception lol. A lot of it is bolted on, including this, which is the only way that we have of passing in an effect that we can access in the calcs to determine resists for everything that has it. I'm just surprised that no one else has noticed the double resist bug until now... <_<

PR #1328, opened 2017-10-29
 Display warp core strength on warp speed tooltip

burnsypet:
Changes to antiwarpscramblingpassive.py:
Fixed a typo of "warmScrambleStatus" to "warpScrambleStatus"

Changes to targetingMiscViewMinimal.py:
Extended existing tooltip to include warp core strength. Warp core
strength is none for ships that do not have a hull bonus (venture etc.)
or do not have any warp core stabs fitted so logic is added to
default value to 0.

Changes to targetingMiscViewFull.py:
As above.

---

Add ship warp core strength onto the existing warp speed tool tip:

image

Fixed an obscure bug in the warp scramble effect too. I would like to expand this to make it so that projected warp disruption effects modify the fit's warp core strength as they currently do not appear to do that.


gui/builtinStatsViews/targetingMiscViewFull.py
```diff
@@ -197,7 +197,12 @@ def refreshPanel(self, fit):
                     elif labelName == "labelFullSigRadius":
                         label.SetToolTip(wx.ToolTip("Probe Size: %.3f" % (fit.probeSize or 0)))
                     elif labelName == "labelFullWarpSpeed":
-                        label.SetToolTip(wx.ToolTip("Max Warp Distance: %.1f AU" % fit.maxWarpDistance))
+                        maxWarpDistance = "Max Warp Distance: %.1f AU" % fit.maxWarpDistance
+                        if fit.ship.getModifiedItemAttr("warpScrambleStatus"):
+                            warpScrambleStatus = "Warp Core Strength: %.1f" % abs(fit.ship.getModifiedItemAttr("warpScrambleStatus"))
```
blitzmann: Is there a reason for the abs() function here? I would think that these would always be positive, and if we were to somehow get a negative here that could lead to incorrect data.

blitzmann:
I like this idea. Just had one question, once that's sorted I can merge this in hopefully with tonights release

gui/builtinStatsViews/targetingMiscViewFull.py
burnsypet: Warp core stabs and hull bonuses appear to apply a negative modifier, so having a +2 strength bonus comes out as -2. I thought it would be confusing to people to see -2 as all the in game descriptions imply +2.

Like I mentioned the projected warp disruption does not seem to amend the warp core strength so there isn't currently a way to get erroneous results yet, but if I expand it to include that it would need to be looked at.

blitzmann: Ah, yeah, because `warpScrambleStatus` would indicate something has you scrammed, so the stregnth is the inverse. I would recommend changing this then to `fit.ship.getModifiedItemAttr("warpScrambleStatus") * -1` to flip the sign.

And yes, indeed, I've never implemented scram/disruption modifiers for these since we never had a UI need for it. But looking at the effects shows that they do indeed modify the attribute (adds positives). Its up to you if you want this merged in now after the change above or to wait untilt he other effects are fixed up first :)

burnsypet: Let's sit on this for now and I'll push a patch set this week once I've read into how projections work a bit.

blitzmann: Rightio, just remove the in progress label when you're ready for merging :) (that is, if you have access to removing it. I can never remember)

burnsypet:
Fiddled with the projected effects on scramblers and fixed a bug with disruptors. Warp disruption field generator will need a bit more thinking since the warp scramble effect should only be applied if it has a super point script loaded, at the moment it doesn't seem to have any projected effects at all.

blitzmann:
@burnsypet for the Field generator, that stuff would probably go on an effect for the script itself. I just checked and the scramble one looks like it has something akin to:

```
def handler(fit, src, context):
    fit.ship.increaseItemAttr("warpScrambleStatus", src.getModifiedItemAttr("warpScrambleStrength"))
    fit.modules.filteredItemIncrease(lambda mod: mod.item.requiresSkill("High Speed Maneuvering"), "activationBlocked", src.getModifiedItemAttr("activationBlockedStrenght"))
    fit.modules.filteredItemIncrease(lambda mod: mod.item.requiresSkill("Micro Jump Drive Operation"), "activationBlocked", src.getModifiedItemAttr("activationBlockedStrenght"))
    fit.ship.increaseItemAttr("gateScrambleStatus", src.getModifiedItemAttr("gateScrambleStrength"))
```

probably not 100% accurate (especially the `activationBlocked` thing - I think we use a different method for that, check the current MWD)

burnsypet:
@blitzmann I've added some new logic to produce the WDFG disruption and scramble effect but unfortunately I wasn't able to find the code you posted anywhere in my repository. I considered adding the MWD scrambling to an effect on the script but the scrambling should only be applied when the scram script is used and not the disruptor script and they both seem to have exactly the same effects. I have used the code the logic from the MWD scram effect and added it to the WDFG warpdisruptsphere.py instead:

```
    if "projected" in context:
        fit.ship.increaseItemAttr("warpScrambleStatus", module.getModifiedItemAttr("warpScrambleStrength"))

        if module.charge is not None and module.charge.ID == 45010:
            for mod in fit.modules:
                if not mod.isEmpty and mod.item.requiresSkill("High Speed Maneuvering") and mod.state > State.ONLINE:
                    mod.state = State.ONLINE
```

In this case it should only do the MWD scrambling if it is a projected effect and the module has the scrambling script charge fitted. I can always change it if you think there is a better way to do it. I think this feature is complete now anyway but I can't remove the label myself.

blitzmann:
> unfortunately I wasn't able to find the code you posted anywhere in my repository. 

Sorry, should have been more clear. That's code that I get from my effect generation script. It's not something that is in the pyfa codebase, it's a PITA to set up, not even close to 100% accurate, but it does help determine what an effect does without tracing through the raw data and is accurate about 80% of the time. If you're interested I can walk you through getting it set up, just shoot me an email :)

> I considered adding the MWD scrambling to an effect on the script but the scrambling should only be applied when the scram script is used and not the disruptor script and they both seem to have exactly the same effects. 

Both scripts have two unique effects relating to what they should do: `shipModuleFocusedWarpScramblingScript` and `shipModuleFocusedWarpDisruptionScript`. It would be best to have the effects defined there as it's the proper place if you can get it working. Keep in mind that when a module has it's effects run, it first looks to see if it has a charge - if it does, it runs the charge's effects first. Many times these charges apply modifications to the module, which then applies modifications when the mods' effects run. Try this - if there's any problems let me know and I'll take a closer look, but so far this looks good :)

blitzmann:
@burnsypet so I looked into this a bit more, I think I see the issue that you ran into. We don't really support charges on projected modules directly modifying the target fit (at least when the module is attached to a fit - if it's a literal projected module, we could make it work). In this case, they always modify the local module, which is then supposed to modify the target fit.

I still think the proper way would be for the charges' effect to apply this bonus. However, that would require a bit of touch up to the projected logic, which is already somewhat fragile, so I don't think I want to touch that right now. Thus, I'm okay with moving forward with this :)

blitzmann:
👍 Great work!

Issue #1332, opened 2017-11-02
Komodo's [shipbonustitang2alldamagebonus.py] should not have thermalDamage assignments


oaao:
There are currently two files that assign missile damage bonus to Komodo:

-  [eos/effects/shipbonustitang1kinthermdamagebonus.py](https://github.com/pyfa-org/Pyfa/blob/8647fe460c17448ade4df8bc694005c1a8dde4f4/eos/effects/shipbonustitang1kinthermdamagebonus.py) for kin/therm damage types (+50%), and

- [eos/effects/shipbonustitang2alldamagebonus.py](https://github.com/pyfa-org/Pyfa/blob/8647fe460c17448ade4df8bc694005c1a8dde4f4/eos/effects/shipbonustitang2alldamagebonus.py) for what **should** be all else (expl/EM, +40%).

However, thermalDamage assignment exists in both, so pyfa is currently double-dipping sources for thermal XL missile damage bonus and displaying values of ~triple what they should be.

L9/10, L15/16, L23/24 should be deleted in [eos/effects/shipbonustitang2alldamagebonus.py](https://github.com/pyfa-org/Pyfa/blob/8647fe460c17448ade4df8bc694005c1a8dde4f4/eos/effects/shipbonustitang2alldamagebonus.py).

[e:] PR included below.

blitzmann:
Hey there,

Good catch! However, this is a known issue and is actually (probably) an EVE bug as well. See https://github.com/pyfa-org/Pyfa/issues/1320#issuecomment-340227162 for details. In hindsight, I probably shouldn't have closed that issue lol.

I can't actually confirm if this happen in-game, but according to the data directly from EVE this is the case. I have a bug report out for it without any activity on it yet - I'll give it another couple of weeks before I ping one of the devs.

As for the PR, I appreciate the fix, however I'm going to leave it be until I hear back from CCP or notice a fix deployed. It is technically a thing that happens in game, so pyfa should be accurate (again, can't test... does anyone have this ship? :P)

PR #1333, opened 2017-11-02
[fix] eos - remove Komodo thermalDamage double-dipping

oaao:
Addresses erroneous thermal damage double-dipping for Komodo titan as outlined in [#1332](https://github.com/pyfa-org/Pyfa/issues/1332).


blitzmann:
Thanks for the PR. When CCP comes out and fixes this bug on their end, I will go ahead and merge this in :) Until then, the calcs pyfa is doing is most likely correct.

See https://github.com/pyfa-org/Pyfa/issues/1332#issuecomment-341426507 for details :)

PR #1353, opened 2017-12-19
improved DPS Simulator graph

taleden:
## Changes

The primary purpose of this pull request is to introduce a new "DPS Simulator" in the Graph window which functions much like EFT's damage graph:

* Multiple attackers and targets may be graphed simultaneously; plots of DPS by distance are calculated for each pair and rendered on the same axes for easy comparisons.
* An attacker must be a saved fit; any fit can be dragged onto the Attackers list to add it, and currently opened fits can be added using the right-click context menu on the Attackers list.
* A target may be either a saved fit or a target profile (previously called "target resists"); target fits can be added in the same ways, but target profiles can only be added by right-click (as there is no good place to drag them from).
* Attackers and targets can both be removed by double-clicking them in their respective list.
* Left-click on a line in the plot to select it; that line then becomes thicker and a caption appears to identify the corresponding attacker and target. The line's color can also be changed by clicking on the matching colored box beside the line caption.
* Right-click on the plot to set a marker at a specific range; each plot will also show an annotation of its exact DPS value at that range.
* Set attacker and target velocity vectors using the graphical vector widgets above the attacker and target lists; left-click to set to any values, right-click to snap to 15&deg;/5% increments, or mouse wheel to change only the speed (as a percentage of maximum).
* For target fits, damage resistances can be ignored, calculated only for a specific layer (hull/armor/shield), or weighted by layer hitpoints (the default). Target profiles have only one set of resistances, so they can only be ignored or considered (any other setting).

A few small changes are also made elsewhere to support the new DPS Simulator graph:

* "Target resists" are now called "target profiles" and may optionally specify the target's signature radius and velocity in addition to its damage type resistances.
* Existing profiles will default to 0 (or blank) for both new values, which will ignore those attributes for applied damage calculations (a signature radius of 0 is treated as infinite).
* When a target profile is selected in an open fit's Firepower panel, the displayed effective firepower will reflect a "best-case scenario" in which range and transversal are zero, but the target's signature radius is still considered.
* Consequently, effective turret damage will appear slightly higher than before since signature radius does not impact turret damage with zero transversal, but the signature radius calculation still factors in wrecking shots which were previously ignored.
* Conversely, effective missile damage may appear lower than before unless the target's signature radius is larger than the missile's explosion radius. Any activated target painters or missile guidance computers on the fit will be factored in, however.

## Screenshots

DPS Simulator

Target Profile Editor

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1,0 +1,32 @@
+# pyfa-taleden
+
+## Changes In This Fork
+
+The primary purpose of this fork is to introduce a new "DPS Simulator" in the Graph window which functions much like EFT's damage graph:
+
+* Multiple attackers and targets may be graphed simultaneously; plots of DPS by distance are calculated for each pair and rendered on the same axes for easy comparisons.
+* An attacker must be a saved fit; any fit can be dragged onto the Attackers list to add it, and currently opened fits can be added using the right-click context menu on the Attackers list.
+* A target may be either a saved fit or a target profile (previously called "target resists"); target fits can be added in the same ways, but target profiles can only be added by right-click (as there is no good place to drag them from).
+* Attackers and targets can both be removed by double-clicking them in their respective list.
+* Left-click on a line in the plot to select it; that line then becomes thicker and a caption appears to identify the corresponding attacker and target. The line's color can also be changed by clicking on the matching colored box beside the line caption.
+* Right-click on the plot to set a marker at a specific range; each plot will also show an annotation of its exact DPS value at that range.
+* Set attacker and target velocity vectors using the graphical vector widgets above the attacker and target lists; left-click to set to any values, right-click to snap to 15&deg;/5% increments, or mouse wheel to change only the speed (as a percentage of maximum).
+* For target fits, damage resistances can be ignored, calculated only for a specific layer (hull/armor/shield), or weighted by layer hitpoints (the default). Target profiles have only one set of resistances, so they can only be ignored or considered (any other setting).
+
+A few small changes are also made elsewhere to support the new DPS Simulator graph:
+
+* "Target resists" are now called "target profiles" and may optionally specify the target's signature radius and velocity in addition to its damage type resistances.
+* Existing profiles will default to 0 (or blank) for both new values, which will ignore those attributes for applied damage calculations (a signature radius of 0 is treated as infinite).
+* When a target profile is selected in an open fit's Firepower panel, the displayed effective firepower will reflect a "best-case scenario" in which range and transversal are zero, but the target's signature radius is still considered.
+* Consequently, effective turret damage will appear slightly higher than before since signature radius does not impact turret damage with zero transversal, but the signature radius calculation still factors in wrecking shots which were previously ignored.
+* Conversely, effective missile damage may appear lower than before unless the target's signature radius is larger than the missile's explosion radius. Any activated target painters or missile guidance computers on the fit will be factored in, however.
+
+## Screenshots
+
+![DPS Simulator](https://i.imgur.com/HVwtnba.png)
+
+![Target Profile Editor](https://i.imgur.com/ugKq1Pm.png)
+
+# pyfa-org (upstream)
+
+Refer to (README.md)[https://github.com/taleden/Pyfa/blob/master/README.md] for information about the original upstream project. To test this fork, please refer to the [FAQ](https://github.com/pyfa-org/Pyfa/wiki/FAQ#requirements-for-running-pyfa-from-source) and [Setting Up Development Environment](https://github.com/pyfa-org/Pyfa/wiki/Setting-Up-Development-Environment) for notes on how to run from source.
--- a/README.markdown
+++ b/README.markdown
@@ -29,4 +29,4 @@ A few small changes are also made elsewhere to support the new DPS Simulator gra
 
 # pyfa-org (upstream)
 
-Refer to (README.md)[https://github.com/taleden/Pyfa/blob/master/README.md] for information about the original upstream project. To test this fork, please refer to the [FAQ](https://github.com/pyfa-org/Pyfa/wiki/FAQ#requirements-for-running-pyfa-from-source) and [Setting Up Development Environment](https://github.com/pyfa-org/Pyfa/wiki/Setting-Up-Development-Environment) for notes on how to run from source.
+Refer to [README.md](https://github.com/taleden/Pyfa/blob/master/README.md) for information about the original upstream project. To test this fork, please refer to the [FAQ](https://github.com/pyfa-org/Pyfa/wiki/FAQ#requirements-for-running-pyfa-from-source) and [Setting Up Development Environment](https://github.com/pyfa-org/Pyfa/wiki/Setting-Up-Development-Environment) for notes on how to run from source.
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ A few small changes are also made elsewhere to support the new DPS Simulator gra
 * Consequently, effective turret damage will appear slightly higher than before since signature radius does not impact turret damage with zero transversal, but the signature radius calculation still factors in wrecking shots which were previously ignored.
 * Conversely, effective missile damage may appear lower than before unless the target's signature radius is larger than the missile's explosion radius. Any activated target painters or missile guidance computers on the fit will be factored in, however.
 
+Below, for reference, is upstream's original README in its entirety. To test this fork, please refer to the [FAQ](https://github.com/pyfa-org/Pyfa/wiki/FAQ#requirements-for-running-pyfa-from-source) and [Setting Up Development Environment](https://github.com/pyfa-org/Pyfa/wiki/Setting-Up-Development-Environment) for notes on how to run from source.
+
 # pyfa
 
 [![Join us on Slack!](https://pyfainvite.azurewebsites.net/badge.svg)](https://pyfainvite.azurewebsites.net/) [![Build Status](https://travis-ci.org/pyfa-org/Pyfa.svg?branch=master)](https://travis-ci.org/pyfa-org/Pyfa)
--- a/README.md
+++ b/README.md
@@ -1,36 +1,3 @@
-# pyfa-taleden
-
-## Changes In This Fork
-
-The primary purpose of this fork is to introduce a new "DPS Simulator" in the Graph window which functions much like EFT's damage graph:
-
-* Multiple attackers and targets may be graphed simultaneously; plots of DPS by distance are calculated for each pair and rendered on the same axes for easy comparisons.
-* An attacker must be a saved fit; any fit can be dragged onto the Attackers list to add it, and currently opened fits can be added using the right-click context menu on the Attackers list.
-* A target may be either a saved fit or a target profile (previously called "target resists"); target fits can be added in the same ways, but target profiles can only be added by right-click (as there is no good place to drag them from).
-* Attackers and targets can both be removed by double-clicking them in their respective list.
-* Left-click on a line in the plot to select it; that line then becomes thicker and a caption appears to identify the corresponding attacker and target. The line's color can also be changed by clicking on the matching colored box beside the line caption.
-* Right-click on the plot to set a marker at a specific range; each plot will also show an annotation of its exact DPS value at that range.
-* Set attacker and target velocity vectors using the graphical vector widgets above the attacker and target lists; left-click to set to any values, right-click to snap to 15&deg;/5% increments, or mouse wheel to change only the speed (as a percentage of maximum).
-* For target fits, damage resistances can be ignored, calculated only for a specific layer (hull/armor/shield), or weighted by layer hitpoints (the default). Target profiles have only one set of resistances, so they can only be ignored or considered (any other setting).
-
-A few small changes are also made elsewhere to support the new DPS Simulator graph:
-
-* "Target resists" are now called "target profiles" and may optionally specify the target's signature radius and velocity in addition to its damage type resistances.
-* Existing profiles will default to 0 (or blank) for both new values, which will ignore those attributes for applied damage calculations (a signature radius of 0 is treated as infinite).
-* When a target profile is selected in an open fit's Firepower panel, the displayed effective firepower will reflect a "best-case scenario" in which range and transversal are zero, but the target's signature radius is still considered.
-* Consequently, effective turret damage will appear slightly higher than before since signature radius does not impact turret damage with zero transversal, but the signature radius calculation still factors in wrecking shots which were previously ignored.
-* Conversely, effective missile damage may appear lower than before unless the target's signature radius is larger than the missile's explosion radius. Any activated target painters or missile guidance computers on the fit will be factored in, however.
-
-## Screenshots
-
-![DPS Simulator](https://i.imgur.com/HVwtnba.png)
-
-![Target Profile Editor](https://i.imgur.com/ugKq1Pm.png)
-
-# pyfa-org (upstream)
-
-Below, for reference, is upstream's original README in its entirety. To test this fork, please refer to the [FAQ](https://github.com/pyfa-org/Pyfa/wiki/FAQ#requirements-for-running-pyfa-from-source) and [Setting Up Development Environment](https://github.com/pyfa-org/Pyfa/wiki/Setting-Up-Development-Environment) for notes on how to run from source.
-
 # pyfa
 
 [![Join us on Slack!](https://pyfainvite.azurewebsites.net/badge.svg)](https://pyfainvite.azurewebsites.net/) [![Build Status](https://travis-ci.org/pyfa-org/Pyfa.svg?branch=master)](https://travis-ci.org/pyfa-org/Pyfa)
--- a/README.md
+++ b/README.md
@@ -21,6 +21,14 @@ A few small changes are also made elsewhere to support the new DPS Simulator gra
 * Consequently, effective turret damage will appear slightly higher than before since signature radius does not impact turret damage with zero transversal, but the signature radius calculation still factors in wrecking shots which were previously ignored.
 * Conversely, effective missile damage may appear lower than before unless the target's signature radius is larger than the missile's explosion radius. Any activated target painters or missile guidance computers on the fit will be factored in, however.
 
+## Screenshots
+
+![DPS Simulator](https://i.imgur.com/HVwtnba.png)
+
+![Target Profile Editor](https://i.imgur.com/ugKq1Pm.png)
+
+# pyfa-org (upstream)
+
 Below, for reference, is upstream's original README in its entirety. To test this fork, please refer to the [FAQ](https://github.com/pyfa-org/Pyfa/wiki/FAQ#requirements-for-running-pyfa-from-source) and [Setting Up Development Environment](https://github.com/pyfa-org/Pyfa/wiki/Setting-Up-Development-Environment) for notes on how to run from source.
 
 # pyfa
--- a/README.md
+++ b/README.md
@@ -1,3 +1,26 @@
+# pyfa-taleden
+
+## Changes In This Fork
+
+The primary purpose of this fork is to introduce a new "DPS Simulator" in the Graph window which functions much like EFT's damage graph:
+
+* Multiple attackers and targets may be graphed simultaneously; plots of DPS by distance are calculated for each pair and rendered on the same axes for easy comparisons.
+* An attacker must be a saved fit; any fit can be dragged onto the Attackers list to add it, and currently opened fits can be added using the right-click context menu on the Attackers list.
+* A target may be either a saved fit or a target profile (previously called "target resists"); target fits can be added in the same ways, but target profiles can only be added by right-click (as there is no good place to drag them from).
+* Attackers and targets can both be removed by double-clicking them in their respective list.
+* Left-click on a line in the plot to select it; that line then becomes thicker and a caption appears to identify the corresponding attacker and target. The line's color can also be changed by clicking on the matching colored box beside the line caption.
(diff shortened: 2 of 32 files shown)


taleden:
I've resolved all the automated style issues, but I don't understand the meaning of the codecov test failures or what to do to address them.

blitzmann:
Wow, @taleden, major kudos to you and this pull request. I haven't had a chance to take this for a spin, holidays and all, but I will definitely check it out soon. One thing, as it looks like a pretty major change, I may add it alongside the current graph for a release while it's evaluated by the community and default to this one, (unless it's already implemented as such). :)

taleden:
Thanks @blitzmann, and happy holidays!

It is indeed already implemented as an additional graph rather than a replacement, because the new one can only model DPS over distance as the variable (in order to simplify the user experience for that common use case), while the original graph can be configured to vary by signature radius, velocity, or any of its other supported fields.

Right now the Graph window will initially pop up to the old graph which I've renamed "DPS Calculator", and you have to switch to this new "DPS Simulator" using the drop-down selector. But the original graph also benefits from some upgrades to the parent graphing window: you can click to select and identify individual lines on the plot and change their colors, in lieu of the old legend which obscured the right side of the plot area; you can right-click the fits list to add fits in addition to the old drag-drop method; and you can right-click the plot area to set a marker and annotate exact DPS numbers for each line at that marker.

If people like the new DPS Simulator it could be made the default selection in the graph window, but I think there's no need to remove the old DPS Calculator since it covers a few additional (though probably less common) use cases.

blitzmann:
I haven't forgotten about this, I plan to take a look at it this week as life morphs back to normalcy after the holidays lol.

blitzmann:
@taleden 

Took a look at this tonight, ran into a few errors:

## Adding target resist profile

I had two fits on the left side. On the right, I right clicked and added a target profile. This popped up:

```
OS version: 	Windows-8-6.2.9200
Python: 	2.7.10
wxPython: 	3.0.2.0
SQLAlchemy: 	1.0.5
Logbook: 	1.0.0
pyfa version: 1.34.0 Stable -  Arms Race 1.3
pyfa root: C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa
save path: C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\saveddata
fs encoding: mbcs

EXCEPTION: unsupported operand type(s) for *=: 'NoneType' and 'float'

  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\builtinContextMenus\targetResists.py", line 55, in handleResistSwitch
    self.mainFrame.graphFrame.addTargetProfile(pattern)
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\graphFrame.py", line 309, in addTargetProfile
    self.calculate(tgtID)
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\graphFrame.py", line 438, in calculate
    x_success, y_status = self.currentView.getPoints(values, fit, tgt)
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\builtinGraphs\fitDpsSim.py", line 177, in getPoints
    data = self.prepareData(values, fit, tgt)
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\builtinGraphs\fitDpsSim.py", line 157, in prepareData
    data["tgtSpeed"] *= tgtThrottle
```

After removing the target profile via double click, one of the fits disappeared from the graph.

#### Clicking a line in the graph throws error

This only happens if you reproduce the above error. After that error happens, if you click on one of the line graphs, this error pops up

```
OS version: 	Windows-8-6.2.9200
Python: 	2.7.10
wxPython: 	3.0.2.0
SQLAlchemy: 	1.0.5
Logbook: 	1.0.0
pyfa version: 1.34.0 Stable -  Arms Race 1.3
pyfa root: C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa
save path: C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\saveddata
fs encoding: mbcs

EXCEPTION: dictionary changed size during iteration

  File "C:\Users\holme\Documents\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\site-packages\matplotlib\backends\backend_wx.py", line 1125, in _onLeftButtonDown
    FigureCanvasBase.button_press_event(self, x, y, 1, guiEvent=evt)
  File "C:\Users\holme\Documents\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\site-packages\matplotlib\backend_bases.py", line 1887, in button_press_event
    self.callbacks.process(s, mouseevent)
  File "C:\Users\holme\Documents\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\site-packages\matplotlib\cbook.py", line 540, in process
    proxy(*args, **kwargs)
  File "C:\Users\holme\Documents\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\site-packages\matplotlib\cbook.py", line 415, in __call__
    return mtd(*args, **kwargs)
  File "C:\Users\holme\Documents\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\site-packages\matplotlib\backend_bases.py", line 1773, in pick
    self.figure.pick(mouseevent)
  File "C:\Users\holme\Documents\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\site-packages\matplotlib\artist.py", line 369, in pick
    a.pick(mouseevent)
  File "C:\Users\holme\Documents\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\site-packages\matplotlib\axes\_base.py", line 3237, in pick
    martist.Artist.pick(self, args[0])
  File "C:\Users\holme\Documents\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\site-packages\matplotlib\artist.py", line 369, in pick
    a.pick(mouseevent)
  File "C:\Users\holme\Documents\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\site-packages\matplotlib\artist.py", line 355, in pick
    self.figure.canvas.pick_event(mouseevent, self, **prop)
  File "C:\Users\holme\Documents\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\site-packages\matplotlib\backend_bases.py", line 1853, in pick_event
    self.callbacks.process(s, event)
  File "C:\Users\holme\Documents\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\site-packages\matplotlib\cbook.py", line 540, in process
    proxy(*args, **kwargs)
  File "C:\Users\holme\Documents\Sync\WinPython-32bit-2.7.10.1\python-2.7.10\lib\site-packages\matplotlib\cbook.py", line 415, in __call__
    return mtd(*args, **kwargs)
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\graphFrame.py", line 583, in <lambda>
    self.canvas.mpl_connect('pick_event', lambda event: self.onPick(event))
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\graphFrame.py", line 662, in onPick
    for lineID, line in self._lines.iteritems():
```

Will continue testing on my end. :)

blitzmann:
Additionally, I'm not too sure I like the legend being removed, at least from the calculator graph. It makes it much more difficult to see at a glance how various fits stack up against one another, forcing the user to click on the line to see what it represents. I get why it's like that for the simulator graph, because you have a line for each combination of atk/tgt and it could get messy.

How about we add the color to a column on the displays on the bottom? That way you can see at a glance which fit is which color. Additionally, is there a way to make lines of alternating color? Like a dashed line, but dashed with two different colors. That could play very well for the simulator if we move the colors to columns - each atk/tgt would have a color, and the alternating line color would better represent that combination I think (if it's possible, you've touched more graph code than I have recently lol, so I defer to you).

Additionally, I'm not fond of having inward number lines - I much prefer the style of the old graph. Is there a reason this was changed?

blitzmann:
@taleden About the color column thing, this is what I was talking about:

image

I can continue to fiddle with it and push my changes to a branch, where you can take a look over it and include it (or I can make a PR into your branch as well).

On the alternating color dashed line for the simulator, this kind points to what I was suggesting:

https://stackoverflow.com/questions/31540258/how-to-get-alternating-colours-in-dashed-line-using-matplotlib

taleden:
I'm still traveling today and tomorrow but your points are well taken; when I get home I'll try to reproduce those exceptions and take a look at your changes. Thanks!

blitzmann:
No rush at all, looking forward to later this week then! :)

taleden:
I wasn't able to reproduce the exception on adding a target profile, is there some other step required? I set my current fit's target profile to None, opened the Graph window, right-clicked to add a second fit (still in the Calculator view), then switched to Simulator and had just the Dummy target profile. I right-clicked to add another target profile, and that worked fine; I double-clicked to remove both and re-added a target profile, no problem. Tried a few more times adding and removing target profiles and attacker fits and couldn't get it to pop up.

Regarding line colors, I do see your point but I'm not sure there's any perfect solution, there will be pros and cons to either approach. I would really like your color-in-the-list scheme if it were just going to be the one list as in the Calculator, but as you pointed out, that scheme just won't work for the dual-list Simulator unless we do something like dual-color stripes as you suggested. But I personally would not like that at all because with even a modestly large number of plotted lines I think striped lines would get really visually noisy and confusing. (For example, in my use case I will often have 12+ lines up at once: 3+ mission fits vs profiles for cruisers, battlecruisers, battleships and sentry guns of some NPC faction. If all of those were multi-color striped, it would just be a mess.)

One way that the current design mitigates color confusion is to remember the color assigned to each atk/tgt combination (or atk alone, for the Calculator) even if that combination is removed, so that if you later re-add it, it will re-use the same color as before. So my thinking is that after a user adds all their relevant attackers and targets, they aren't going to stick with the automatically assigned colors no matter what they are, they're almost always going to want to click through the lines to assign colors according whatever scheme makes sense to them (such as, for example, red for an attacking Typhoon and blue for an attacking Raven, with light/medium/dark shades of those hues for each attacker vs a set of targets). Then they can start playing with vectors and tweaks to the fits to see how the plot changes, and since they chose all the colors, they won't have too much trouble keeping track of what's what, even without a persistently visible legend for all lines at once.

Would it be a reasonable compromise to just bring back the on-plot legend from your trunk for the Calculator, and leave the Simulator without? Or I guess there could be a toggle for the legend, but that takes more UI space.

On the inward number lines, that was mostly an attempt to reclaim some wasted screen space especially when expanding the Graph window size. With the old style, the bigger you made the window the more empty space was wasted surrounding the plot; I figured most users would prefer to have as much space as possible for the actual plotted lines, so I re-aligned the axes to fill the entire UI space, which required the ticks to be flipped inward. (I tried at first to just reduce the margin so that there was just enough room for the axis labels without all the wasted extra, but that turns out to be basically impossible to do in matplotlib; zero margin and inward ticks was the second best solution.)

Let me know what you think of all this, and thanks for taking a look!

blitzmann:
Hi there @taleden,

Sorry it's taken me a while - I've been busy the past two weeks moving to a new abode :) Catching up now that we have most things unpacked lol

I see why it might be kinda ehh to have striped lines, could get really visually confusing. It's just really a pain to have to click the lines to figure out what they are and I would like to avoid that if at all possible. Users should be able to see that information at a glance.

> So my thinking is that after a user adds all their relevant attackers and targets, they aren't going to stick with the automatically assigned colors no matter what they are, they're almost always going to want to click through the lines to assign colors according whatever scheme makes sense to them

You would probably be correct for some users, especially power users, and I agree that's how users *should* use this, but I feel like most users would keep the colors as-is. Especially if they have something like 15 combinations (5 and 3): click the line, click the color, change the color, repeat 15 times is a lot to assume for the average end user just to get a graph (and in this situation, I would be an average end user :P).

I have three ideas, the last one being the one I'm leaning towards as a possible solution. Each one would utilize the Color column as described and detailed above and in the PR I sent, and the DPS calculator can use that just like it uses the legend today - maps a fit to a color, but also gets the legend out of the plot so that there is more room for plotting. I actually like this idea a lot for the calculator and would like to implement it regardless of what happens with the simulator. The following are my ideas for the simulator. Each one would avoid the striped lines and provide at-a-glance colors for the combos.

### Mergetool-like design

Hear me out, this might sound / look a bit weird, but how about another list that lists the combinations and has the color for that combination. Kind of like a merge tool - left has attackers, right has targets, middle has the combinations of them with their color. Basically acting as a legend again, just not drawn in the plot (if it's drawn on the plot, would get extremely crowded and ugly with even a small number of combinations). Middle list would be read-only, can't interact with it, is compiled via the information in the other two lists. 

Cons:
* More UI space (I'm okay with this, even if we need to make the window a bit bigger)
* Might be a bit confusing to end-user at first what it's supposed to be?

### Single list with manual attack / target combinations
If it's okay to drop support for running every atk/tgt combo (I don't know the use cases well enough, so I'm going to defer to you on this) - have a single list that has three columns - color, attacker, and target. User manually sets the attacker / target combination in each row which will get it's own color.

Cons:
* Lose the ability to automatically calculate every possible attack / target combo for 2 given lists, which I'm assuming is one of the highlights of the current system, so this idea might be moot

### Single list with attacker / target combo editor

This one is basically the same thing as above - single list that provides the combinations. However, there would be a button that would open a window for the user to edit the attackers / targets (basically using the same UI in the current iteration of the simulator), so that the user can remove fits from the list, which would be compiled from those two lists in the window.  This allows us to keep the functionality of automatically generating combinations. The context menus can and should still be able to add / remove fits to either side. Hell, we can even get rid of the new window in this idea and just manage it all through context menus:

```
Add Attacker
|-- Currently Open Fits > 
Remove Attacker
|-- Fit 1
|-- Fit 2
(repeat for target)
```

Cons:
* Delegates management to another window (or at the very least context menus)

As stated above, I'm currently leaning towards this being the preferred solution. It separates the management of the fits to another window, and allows the list under the graph to actually represent the graph.

---

Please let me know your thoughts on the items above. Each of these ideas requires more work to be put into this PR, especially the last idea (I believe the other two would be more or less straight forward), but if you need assistance, let me know what I can do :)

You make an interesting example with having attackers have a primary color, while the line to the target is a different shade / hue of that color. This might be something we want to do (eventually) by default. But lets get this thing solved and up and running first. :)

As for the exceptions - I will take another look. I got them simply by pulling your changes down and opening on my existing DB - maybe it's a migration issue or something, but I will double check and get back to you on that. I wont worry about fixing those bugs until it gets closer to final product, especially if we go the route of tweaking the simulator based on discussions above. Would you mind posting your db file as a reference to compare to, as long it doesn't have API keys / secret fits in it?

taleden:
No worries @blitzmann, grats on your new digs!

I think my initial preference is for a version of your first solution, actually, because it seems like the simplest to implement and still serves both usage styles pretty well: the center list would essentially be a legend just like users are accustomed to from the current version, but simply moved off the plot into its own UI space where it doesn't cover anything up; meanwhile, people like me could still use it to set custom colors for each pairing and then ignore the legend and just look at the plot, knowing what all the colors mean.

So the way I imagine that working is that for the Calculator, there would still just be the single list of fits but they'd have your new color column, and I guess clicking on the color patch in the column would let them set the color for that fit. When switching to the Simulator (or any future modes that use separate attacker/target lists), then the center legend would also appear, but the color column of the attacker list would disappear (because colors would no longer be per-attacker, but per-combination in the legend). Attackers and targets would still be added/removed to the left/right lists as before, while the center list is only for viewing (and changing) colors of plotted combinations (with the extra bonus of being a place to hide or disable specific combinations, in cases where the user doesn't actually want to see every possible combination).

Does that sound reasonable to you? I'll see if I can adapt your color column code into this kind of scheme to try it out, though it occurs to me that I can't think of any other place where columns are dynamically shown or hidden, so hopefully that isn't too hard to do for the color patch in the attacker list.

I wonder also if it would be a good idea to change how deleting entries works. Right now it's a double click on the line because there's no other possible interaction for currently listed fits, but with the addition of the color column I wonder if it would make more sense to also add a "delete" column with a button or text indicator to click there to delete the entry, rather than double clicking anywhere on the line except for the color patch. What do you think?

blitzmann:
> Does that sound reasonable to you? I'll see if I can adapt your color column code into this kind of scheme to try it out, though it occurs to me that I can't think of any other place where columns are dynamically shown or hidden, so hopefully that isn't too hard to do for the color patch in the attacker list.

Yep, sounds reasonable, and is exactly how I envisioned the first solution. While I still think I like the third solution a bit better, this is indeed simplier to work out and we don't have to worry about developing a separate window and it's itneractions with the graph window. :)

Not sure what you mean about showing / hiding columns. I don't know if it's possible, but you should just be able to leave it off the column ist for the attackers and targets and only include it with the combination list.

> I wonder also if it would be a good idea to change how deleting entries works. Right now it's a double click on the line because there's no other possible interaction for currently listed fits, but with the addition of the color column I wonder if it would make more sense to also add a "delete" column with a button or text indicator to click there to delete the entry, rather than double clicking anywhere on the line except for the color patch. What do you think?

Negative - convention within pyfa is to double click to remove items. If the user clicks the color column, it should ignore the delete and go to the color picker. This is similar to how modules work on the fitting window - double click to remove module, however if you're clicking on the State column (offline/online/active/overheat) it toggles the state instead. You can take a look at that code (`gui/builtinViews/fittingView.py` or something like that) to see how it functions.

I think this is a good discussion and solution, looking forward to seeing how it turns out. Let me know if you need anything! :D

taleden:
> Not sure what you mean about showing / hiding columns. I don't know if it's possible with the current codebase, but you should just be able to leave it off the column list for the attackers and targets and only include it with the combination list.

In Calculator mode, only the left-hand "attacker" list is visible (using the whole bottom space) so that list will need the color column in order to also serve as the legend. But in Simulator mode with its right-hand "target" list and the new center "pairs" list, the color column should be only in the "pairs" list since that becomes the legend, which means the color column in the "attacker" list should be hidden. Hopefully I can figure out how to do that dynamically, otherwise I guess the alternative is to maintain two separate "attacker" lists (one with color and one without) and swap between them, but that'd be a little more kludgy I think.

> Negative - convention within pyfa is to double click to remove items. If the user clicks the color column, it should ignore the delete and go to the color picker. This is similar to how modules work on the fitting window - double click to remove module, however if you're clicking on the State column (offline/online/active/overheat) it toggles the state instead. You can take a look at that code (gui/builtinViews/fittingView.py or something like that) to see how it functions.

Ah, okay, fair enough. I'll look there for an example of overriding the click handler for the color column.

We'll have guests this weekend but hopefully I can get this at least partly implemented sometime next week so we can see how it looks and feels. Thanks for your feedback!

blitzmann:
Oh, the both graphs are using the same base list? I thought each graph had it's own gui panel that it displayed. That might be the best route instead of trying to juggle toggling things on/off depending on which graph we have. ofc, it's been so long since I looked at the graph code, it might very well be a PITA to do that lol

blitzmann:
@taleden just curious as to where this might be. I'm currently busy with the 2.x codebase and getting it up to snuff before an official stable release (hopefully in the next couple months), but it would be awesome if this could be included as a new feature for it :)

If time is limited, let me know and I can start poking at it as well with what we discussed above.

nickPietrass:
Is this currently being worked on?
I would very much like to see this in pyfa and would just throw my hat in and say, that i can start working on it if this is abandoned.

PeterGabriel2:
Guys, this takes you soooooo freaking long.... amazing feature but "two years" later we still don't have it in official release... I just wish pyfa was coded in any other language (java, kotlin, C#, Delphi) so I could help... but when I saw how much work (code) has to be done for such simple GUI enhancements I realized that python is just "no go zone" for me. Very sorry.

I just have suggestion maybe?
Maybe creating "code modules" less dependent on existing modules (or maybe better no dependent at all) would help next time to deliver features sooner?
Because we are in 2019. Time is essence. I don't believe that code cleanness must be perfect. I believe that better approach would be put modules in "BETA" state in release if they are independent "windows" and let community use them even if they are not perfect. (Users doesn't care about code) 
Code can be made better in next version... I believe that put small panel with label "This is feature in BETA state... use it on your OWN risks..." would be better in comparison to don't have feature at all. 
What do you think?

PS: I am very sorry for my criticism - It is not my attention. I just want to figure out how can it be done faster - I very appreciate amazing work both of you.

DarkFenX:
> Because we are in 2019.

Exactly, and pyfa is 10+ year old software. It carries alot of legacy and many of its modules are interconnected since its inception (and neither me nor blitzmann did that - majority of original code is written by Darriele and cnc). Many people prefer to just pump the feature out as soon as possible instead of spending on it 3x time to have clean isolated code. And my experience says that this approach has some merit, especially in the beginning - better to have the program out before you burn out. Same for "big features". When you have some spare time and you look into how you can implement e.g. mutaplasmids - you see that current architecture doesn't let you to implement it cleanly, and you again face the choice of ignoring many "best practices" and releasing it ASAP, or rewriting everything to support it properly (which can easily take month+), possibly burning out in process for a month or two, and delaying initial feature release significantly.

I know that very well - because I did and still (passively) doing my OCD project, next gen eos, where I follow all these practices with modularization, huge test set etc.

I can't give any promises to rework/merge it anytime soon (especially since I've implemented all features I wanted in a few recent releases and wanted to spend my time on other projects), but i will look into it, and if it doesn't take much time - I might do that.

DarkFenX:
Well, I've got no idea why my comment does not appear as the last message, but just in case - scroll up a little for my response.

DarkFenX:
@taleden, if you still have some mojo for updating it, please contact me (e.g. via pyfa slack or by any other means). That would help alot for pushing these improvements out.

DarkFenX:
I am closing this pull request as I just merged my version of dps sim. I've reused your vector picker (with some modifications), color picker and hsl to hsv converter. And, obviously, alot of inspiration from your pull request.

Issue #1360, opened 2018-01-10
Nos +cap amount is reduced by fit cap batteries


alelazar:
## Bug Report
Nos +cap amount is reduced by fit cap batteries

### Expected behavior:
Nos amount should not be impacted by fitting a cap battery. Nos amount is only impacted by a cap battery on the fit of the ship being targeted. 

### Actual behavior:
Having a cap battery fit reduces the amount a nos gives you. It does not impact the drain amount stated, just the gain amount. 

### Detailed steps to reproduce:
See above. 

### Fits involved in EFT format (Edit > To Clipboard > EFT):
[Bhaalgorn, nos test]

[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]

Republic Fleet Large Cap Battery
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]

Corpus X-Type Heavy Energy Nosferatu
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]

[Empty Rig slot]
[Empty Rig slot]
[Empty Rig slot]

### Release or development git branch? Please note the release version or commit hash:
Pyfa 1.34.0

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
Windows 7, OSX 10.12.6

blitzmann:
Possible duplicate #1144, #1304

Last time I took a look at it, it wasn't an easy fix (IIRC). I'll have to take a look at it again to refresh my memory. Can you give me exactly the values that you get vs what you are expecting? The more data, the better :)

alelazar:
The nos +cap amount (but not the -cap to target) is reduced by exactly the % stated by the cap battery. 

I would guess that wherever is doing the nos calculations is accidentally pulling the cap battery field even though that field is for nos targeting you, not your own nos.

burnsypet:
In eos/effects/energynosferatufalloff.py the capacitor need calculation was using the amount that had already been modified by the energy warfare resistance. I believe I introduced this defect when working on #1322  😨 

Fix is to only apply the resistance modifier when dealing with a projected module. Will submit a pull request soon to resolve.

blitzmann:
> I believe I introduced this defect when working on #1322 😨

[...]

PR #1367, opened 2018-01-26
Remove energy warfare resistance on capNeed calc

burnsypet:
When calculating the capactitor need of an energy nosferatu the
amount was being modified incorrectly by the energy warfare
resistance of the ship. Change to only apply energy warfare
resitance modifier if this is a projected effect.

---

#1360 

When fixing #1322 I introduced a new bug that caused the capacitor need of nosferatus to be modified by the energy warfare resistance of the ship. The correct approach is to only modify this amount in the case where this is a projected effect.

### DIFF
--- a/eos/effects/energynosferatufalloff.py
+++ b/eos/effects/energynosferatufalloff.py
@@ -12,7 +12,7 @@ def handler(fit, src, context, **kwargs):
     amount = src.getModifiedItemAttr("powerTransferAmount")
     time = src.getModifiedItemAttr("duration")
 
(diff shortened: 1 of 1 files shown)


blitzmann:
A quick check on sisi shows that this indeed does fix the issue with cap batteries and nos. I'm not 100% it fixes everything related to them - I remember looking into this and it being a fairly PITA problem, but I may have overlooked something simple like this. I'll go ahead and merge this into master and flag the issues surrounding them as done.

Thanks! :)

PR #1391, opened 2018-02-15
Character Editor: accept skill plans in Roman numerals too

BlueShanks:
Needs code review - I'm a python and git developer noob.

Added imports of "re" and "roman" modules to characterEditor.py too, not sure if these need adding someplace else so pyfa builds fine

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,3 +118,4 @@ ENV/
 .idea
 eos.iml
 gitversion
+*.swp
--- a/gui/characterEditor.py
+++ b/gui/characterEditor.py
@@ -43,6 +43,7 @@ import re
 
 pyfalog = Logger(__name__)
 
+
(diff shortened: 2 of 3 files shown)


blitzmann:
Hi there! I saw your posts on tweetfleet Slack, but wasn't able to respond today due to busy day at work lol. Home now, code looks good. I will test this out soon and probably merge it :)

Thanks!

blitzmann:
Merged! Even though new development is focused on the new `py-3` branch and master is supposed to be for bug fixes and data update only, I went ahead and modified this PR a bit to add the `roman` module to pyfa so as to avoid legacy issues with our current build process and merged it into master. These issues aren't present in the new 2.x cycle on `py-3`, so it's been added to that branch as it was originally developed :)

Thanks so much for your contribution! I'm sure there's a few folks that could find this useful :D

PR #1444, opened 2018-02-27
Ensure graph frame unbinds its event handlers.

Neugeniko:
This is a partial reversion of d4df989 which seems required for the graphFrame to unbind event handlers.


blitzmann:
Don't worry about the failed Travis checks - I've haven't been making sure this branch is up to style snuff. I'll fix those as we get closer to release :)

Issue #1456, opened 2018-03-02
Warp Disruption Field Generators & Standup Focused Warp Disruptor working incorrectly.


Neugeniko:
### Expected behavior:
These modules only turn off MWD when they have a scrambling script loaded.

### Actual behavior:
These modules turn of MWD all the time.

### Detailed steps to reproduce:
Self project the following fit and note the mwd is turned off irrespective of it being loaded
with disruption, scrambling or none scripts.
Also project the Standup Focused Warp Disruptor and note the same.

### Fits involved in EFT format (Edit > To Clipboard > EFT):
[Broadsword, AC IP PS MW]

IFFA Compact Damage Control
Gyrostabilizer II
Gyrostabilizer II
Tracking Enhancer II

50MN Y-T8 Compact Microwarpdrive
Large F-S9 Regolith Compact Shield Extender
Adaptive Invulnerability Field II
Large F-S9 Regolith Compact Shield Extender
Adaptive Invulnerability Field II
Adaptive Invulnerability Field II

425mm AutoCannon II
425mm AutoCannon II
Warp Disruption Field Generator II
425mm AutoCannon II
425mm AutoCannon II
425mm AutoCannon II

Medium Core Defense Field Extender II
Medium Core Defense Field Extender II

### Release or development git branch? Please note the release version or commit hash:
py3

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
Win 10

Neugeniko:
https://support.eveonline.com/hc/en-us/articles/115004925705-Warp-Scrambling-and-Warp-Disruption

Neugeniko:
warpscrambleblockmwdwithnpceffect.py and structurewarpscrambleblockmwdwithnpceffect.py are the places where the mwd is turnoff.

Neugeniko:
warpdisruptsphere.py effects the Warp Disruption Field Generators. But it has some garbage? that is causing another bug. They also turn off MJD so we need to turn those off too. I am unable to reproduce the original bug that started this issue.

All this stuff I dont think it belongs here.

fit.ship.boostItemAttr("mass", module.getModifiedItemAttr("massBonusPercentage"))
fit.ship.boostItemAttr("signatureRadius", module.getModifiedItemAttr("signatureRadiusBonus"))
fit.modules.filteredItemBoost(lambda mod: mod.item.group.name == "Propulsion Module",
[...]

PR #1471, opened 2018-03-07
Redefine operation of numerous effects.

Neugeniko:
Added the code to turn off MJDs for various modules. Also ensure they operate correctly depending on script loaded. Various other corrections to operation. See Issue #1456 for further details.

### DIFF
--- a/eos/effects/doomsdayaoebubble.py
+++ b/eos/effects/doomsdayaoebubble.py
@@ -1,0 +1,9 @@
+# doomsdayAOEBubble
+#
+# Used by:
+# Module: Standup Warp Disruption Burst Projector
+type = "projected", "active"
+
+
+def handler(fit, module, context):
+    return
--- a/eos/effects/structuremissileguidanceenhancer.py
+++ b/eos/effects/structuremissileguidanceenhancer.py
@@ -1,14 +1,15 @@
 # Not used by any item
-type = "active"
+type = "passive"
 
 
 def handler(fit, container, context):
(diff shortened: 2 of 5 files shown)


burnsypet:
Looks good.

I think I was the last person to touch this when I was adding the warp scramble strength thing, guess I just totally forgot that MJDs existed when I did that.

Neugeniko:
One thing that is nice now is that we can see things in the in game simulator. It doesn't help with projections thou, and we still need someone to test those on the test server usually with two accounts.

Yeah looks like the travis check is due to line length. The appveyor fail seems outside the scope of the pr. I'll let blitzmann handle any changes until I get familiar with the requirements.

Neugeniko:
Corrected the structure missile guidance enhancer effect.

Neugeniko:
Added a effects file for the Standup Warp Disruption Burst Projector (shared with the super based equivalent).

blitzmann:
>Edit: Just noticed the code check is kicking off, I think it's because you've included surplus extra lines or exceeded the character limit of 165 in warpdisruptsphere.py

The py3 branch will remain in a failing state in travis since I'm not worried about keeping code guidelines in placed right now as development on this branch is fast going. AppVeyor also breaks since there's new build processes with this branch, and I haven't had the time or motivation to fix it yet. both of these will be back to normal by the time 2.x goes stable and is released.

@burnsypet weren't you working on a cap delta thing? how's that coming along?

PR #1481, opened 2018-03-11
Attribute disallowRepeatingAction has been renamed.

Neugeniko:
Make appropriate modification to instances of this attribute
name in code.

### DIFF
--- a/eos/saveddata/module.py
+++ b/eos/saveddata/module.py
@@ -752,7 +752,7 @@ class Module(HandledItem, HandledCharge, ItemAttrShortcut, ChargeAttrShortcut):
 
     @property
     def disallowRepeatingAction(self):
-        return self.getModifiedItemAttr("disallowRepeatingAction", 0)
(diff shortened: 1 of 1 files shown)


blitzmann:
Hmm, that's weird. So, that property was added with b3c727. I went back to that commit and took a look at `eve.db`, and that attribute isn't available there either. I'm thinking that the original property was typo'd, and it's been broken for a little over a year - I think it's always been `disallowRepeatingActivation`? Nice catch!

blitzmann:
Just fyi, for next time, if working on a bug that affects both master and another branch, base your fix branch off master and make a PR to master. As it stands right now, I can't merge this into master without bringing over all the py3 stuff.

Not a big deal - I can always (and will) [cherry pick](https://git-scm.com/docs/git-cherry-pick) this commit from py3 to master. Just a heads up for next time. Thanks!

blitzmann:
Done and done!

Neugeniko:
Ok if it effects both a branch and master, base the fix branch off master. Gotcha.

Issue #1490, opened 2018-03-14
CapSim uses a fixed reload of 10 secs.


Neugeniko:
### Expected behavior:
CapSim uses the reload of the module when asked.

### Actual behavior:
CapSim uses a reload of 10sec for all modules when asked.

### Fits involved in EFT format (Edit > To Clipboard > EFT):
Fits that use AAR(60sec reload), Remot AAR(60sec reload), Hybrid weapons(5sec reload)
are effected. There may be others.

### Release or development git branch? Please note the release version or commit hash:
master

PR #1491, opened 2018-03-14
CapSim changes to make reloadtime a variable.

Neugeniko:
This allows the module attribute reloadtime to be passed to the capsim. This addresses #1490. Without reload preference true, this capsim, sims exactly the same as previously. With reload preference true the capsim is only different for fits with modules that have a reload other than 10secs.


blitzmann:
Makes sense to me. I think the 10sec reload was a very, very old assumption way back when. Thanks!

PR #1493, opened 2018-03-15
Beam type doomsday DPS - Resolves Issue#1486

Neugeniko:
This code is lifted directly from Ebag333. Please set him as the contributor.


blitzmann:
It's been a while since I've done it, but I was successful with changing the commits involved to show you as the contributor and @Ebag333 as the author, for proper attribution :D Since it was a rewriting of the commits, the dates a are a bit funky, and I used my own branch for them, so I'll be closing this one.

image

https://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit 

Thanks to both!

PR #1498, opened 2018-03-17
Create basic effects files for various modules.

### LINKED ISSUE #1484: Numerous modules have no active state but should show cap use.
### Expected behavior:
Modules like the reaper, lance, gtfo, bosonic, mjfg etc use cap.

### Actual behavior:
Modules dont use cap since they can't be activated.

### Release or development git branch? Please note the release version or commit hash:
1.35.2 or py3

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
win 10

Neugeniko:
Partially resolves #1484. We still have the following modules that use cap but have no active state.

Jump Portal Generators.
Cargo Scanners
Ship Scanners
Survey Scanners
Passive Targeters

### DIFF
--- a/eos/effects/doomsdaybeamdot.py
+++ b/eos/effects/doomsdaybeamdot.py
@@ -1,0 +1,9 @@
+# doomsdayBeamDOT
+#
+# Used by:
+# Module: Lance type modules
+type = "active"
+
+
+def handler(fit, src, context):
+    pass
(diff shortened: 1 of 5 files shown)


Neugeniko:
Yay I passed a travis check...booo what is a codecov check and how do I pass those?

blitzmann:
> booo what is a codecov check and how do I pass those?

Code coverage. It's a way to check to make sure that a project has unit testing that touches as much of the logic as possible. There was a brief stint to try to get unit testing going for some things in pyfa, but a lot of them are currently disabled due to them not being properly mocked in a lot of the instances (db updates have changed values, which causes assumptions to change, etc)

I really should disable CodeCov because it just isn't really used right now. :)

PR #1513, opened 2018-04-02
Issue/1369

fsufitch:
Depends on https://github.com/pyfa-org/Pyfa/pull/1510 since local development would have been broken without it.


blitzmann:
This is amazballs, nice work!

PR #1549, opened 2018-05-09
Add Drone Damage Amplifier to defaults

flashspys:


flashspys:
I don't know why all these checks fail, but they shouldn't fail on this single change.

blitzmann:
> I don't know why all these checks fail, but they shouldn't fail on this single change.

When 2.0 released, travis shit the bed. I have to reconfigure it, and I've been tight on time :)

Thanks for the addition! :)

PR #1578, opened 2018-05-15
Use proxies from Pyfa's "Network settings" for ESI requests

minlexx:
Requests session objects already can use proxies for all requests it makes. Just need to provide them to session object

### DIFF
--- a/service/esiAccess.py
+++ b/service/esiAccess.py
@@ -18,7 +18,7 @@ import base64
 
 import datetime
 from eos.enum import Enum
-from service.settings import EsiSettings
+from service.settings import EsiSettings, NetworkSettings
 
 from requests import Session
 from urllib.parse import urlencode, quote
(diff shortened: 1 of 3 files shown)


minlexx:
It looks strange to me that Pyfa already has service/Network class that (ideally?) should do all networking, but esiAccess is using its own method. This works too, but results in some code duplication between Network service and esiAccess (specifically code used to construct proxies in format suitable for `requests` module).

This PR already works as is, but I have one idea on how to reduce this code duplication a little bit (involving also some refactoring in NetworkSettings, Network) - that's why WIP.

blitzmann:
>With this PR, pyfa itself uses that proxy too, for esi requests, and I could finally import my character's skills successfully

Just an aside to make sure this is clarified, the only thing pyfa.io is used for is initial sso login and to refresh access tokens. Once you have a valid access code, pyfa uses direct calls to ESI using that code.

Glad to hear that it seems to work though, excellent news! Please let me know when this is ready to be merged, the code itself looks fine

> which is not intuitive at all

>_> Will be fixing this with the next release! :P

blitzmann:
oh shit, forgot about this for 2.0.2. Should come with the next release now :)

Issue #1545, opened 2018-05-08
Possibility of Add Multi-language Support


BikerGitHub:
Many chinese players are difficult to read English.
So,"Air Dragon" did this chinese version by edit the "eve.db".
[eve.zip](https://github.com/pyfa-org/Pyfa/files/1984374/eve.zip)
I want to ask is there any possibility to add multi-language support.
"Air Dragon"said:"Add multi-language support did not need too much change. "

blitzmann:
It is not too difficult to swap out the english language for another language for EVE data like ships and modules. I simply have to tell my data script which language to use.

The difficult part is to have both English and Chinese (or Russian, Spanish, etc) available in the same database - that would be a big undertaking. A bigger undertaking would be to also have the pyfa UI translated.

These are things that I wish to look into eventually, something like having the ability to download and "install" a language pack.

I will also look into stop gap measures such as making available for download an `eve.db` for a specific language.

blitzmann:
Also, I would be careful using a database that was translated. There are certain things within pyfa that are hardcoded in english that might not work with a translated database (things like Group names, certain item names, etc) :)

icyskyair:
Hello, I'm the "Air Dragon" he mentioned.
I want to help you to finish the multi-language function.
<hr/>
I think it is not suitable to translate `eve.db` to specific version. It would cause some problem:
1. The translation of Special Edition Ship will cause a crash on start.
2. The translation of Skills would cause a crash on adding a fitting.
3. Translated items can not be searched.
4. Other problems I have not find.
The translation of `eve.db` is just a experiment, not a eventually way to address it.
<hr/>
I suggest to seprate `process` and `show`. `process` in English and `show` in specific language. The language pack is just a table of translation of strings. When we want to display a string, we search in the table, if exsit, translate it.
Although it is a hard coding, we can change it little by little. The unmodified part will display in English instead of a total crash.

blitzmann:
> I suggest to seprate `process` and `show`. `process` in English and `show` in specific language. The language pack is just a table of translation of strings. When we want to display a string, we search in the table, if exsit, translate it. Although it is a hard coding, we can change it little by little. The unmodified part will display in English instead of a total crash.

[...]

PR #1581, opened 2018-05-15
Multilanguage

HaoxuanGuo:
This is a simple edition for multilanguage.
**Multilingual features may reduce some performance**
I added language packs in the pyfa. Of course, only some simple translate. I have to admit that this function is a huge work. I will do it little by little until all the part can be translate by an extra language pack. 
Up to now, I only translated some UI. Because of the huge work, I decide to merge it when I did a part to escape some unexpected problems.
So, this is the first part. I will do it little by little.


blitzmann:
@icyskyair 

Neato, thanks for this initial work!

However, I wish to stress that localization is going to be a huge project (as I'm sure you're aware). I want to know more about it before we make a huge effort for it.

### UI Localization

I do have a suggestion. Instead of rolling your own translation database, where you have a `from` and `to` column, and then writing functions that fetch that, I would recommend that you instead use the packages that python itself gives you, which is standardized:

https://docs.python.org/3/library/i18n.html

It's nearly the same thing that you're doing, however instead of a sqlite database, it runs on flat files. Additionally, this is a standard for software development, so I would much prefer this route.

The way that it works is that you provide language packs that's basically a key value pair. For example:

English:"
```"additional_panel_projected": "Projected"
"additional_panel_command": "Command"```

Simplified Chinese:
```"additional_panel_projected":	"预计"
"additional_panel_command": "指挥"```

And to call it in python;

`_('additional_panel_projected')`

Pretty simple. It's just an example, so I'm not too sure on the proper syntax (I'm sure the actual language pack file is different). But this is the standard way of doing localization for the UI.

### EVE data localization
The EVE Data is probably not going to be handled by the python localization. I don't quite know what I'm going to do about it, I first need to run some benchmarks and see just how much extra data having multiple languages would incur, and if it's too great, devise a way to support a plug-able system maybe? not sure, but this is going to take much more time and research to get right

blitzmann:
In fact, I remember there was some guy that was trying to localize the UI...

Yep! https://github.com/liuyuran/Pyfa

Note how his localization files are under a structure like `locale/en_US/LC_MESSAGES/`. And here's the mapping for it: https://github.com/liuyuran/Pyfa/blob/master/locale/en_US/LC_MESSAGES/lang.po.

There's also a `lang.mo`. i'm not exactly sure what that is, maybe a compiled version?

This is utilizing the i18n standard for localization, and is the direction we should take. You can look at how that repo gets it set up if needed

Looks like that repo went dead a while ago though, no updates since Aug 2016. So do keep in mind it's using older code, python 2.7. I can load it on my machine, but not sure if you have the dependencies.

@liuyuran you still around?

icyskyair:
@blitzmann Okay, I will have to try.

blitzmann:
Closing in favor of recent discussions and work surrounding localization, see #1545, 2198

PR #1591, opened 2018-05-19
Copy text from item traits pane

### LINKED ISSUE #1529: Bug/FR: Unable to copy text from info dialogs
You can select text just fine, but Ctrl+C won't copy text on Windows in the latest version of the Pyfa 1.x.

image

minlexx:
Popup menu with "Copy" command (can work more reliable that shortcurs)
Select all on Ctrl+A
Copy to clipboard on Ctrl+C
Fixes #1529

### DIFF
--- a/gui/builtinItemStatsViews/itemTraits.py
+++ b/gui/builtinItemStatsViews/itemTraits.py
@@ -37,6 +37,9 @@ class ItemTraits(wx.Panel):
         # Ctrl + C
         if keyCode == 67 and event.ControlDown():
             self.copySelectionToClipboard()
+        # Ctrl + A
+        if keyCode == 65 and event.ControlDown():
+            self.traits.SelectAll()
 
     def copySelectionToClipboard(self):
         selectedText = self.traits.SelectionToText()
--- a/gui/builtinItemStatsViews/itemTraits.py
+++ b/gui/builtinItemStatsViews/itemTraits.py
@@ -13,5 +13,24 @@ class ItemTraits(wx.Panel):
         self.traits = wx.html.HtmlWindow(self)
         self.traits.SetPage(item.traits.traitText)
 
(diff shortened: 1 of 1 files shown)


blitzmann:
Excellent, found a bug though:

Open up traits for a ship, select some text, and copy. It copies fine.
Close the item window, find another, different ship, open it, and DON"T select, just right click > copy. it copies the old selected values.

I don't think it's a critical bug, think we could still include this even with it, but I'll let you investigate first to see if there's a quick fix:)

minlexx:
> ... and DON"T select, just right click > copy. it copies the old selected values.

Could it be that previously selected text just remained in clipboard? Because if nothing is selected, it does not perform clipboard copy operation.
I guess I could just disable "Copy" menu item..?

blitzmann:
This is wonderful, thanks!

since you've poked around the code a bit, would is be possible to add the Copy context menu to the description of an item as well, not just the traits? We can still select and ctrl+c, but I think having a copy context would work there as well. But other than that, this looks excellent, and thanks so much for picking this task up :)

minlexx:
> would is be possible to add the Copy context menu to the description
> of an item as well, not just the traits?

Do you mean, not only context menu, but also shortcuts handling (Ctrl+A, Ctrl+C)? Because IMO description pane also doesn't have this functionality.

blitzmann:
I was thinking just the copy button, however looking at it it seems to use the same widget as the traits

`wx.html.HtmlWindow(self)`

So we should be able to perform the same thing here, yeah?

If it's gonna take a while, don't worry about, just let me know and we can plan to look into it later. But I know people are gonna ask "I can copy traits, why can't I copy description" :P

Gonna merge this in this weekend regardless of the description pane though, in prep for next weeks release :)

Issue #1521, opened 2018-04-12
DPS graph problem


dyh135:
## Bug Report

Graph does not show the dps after changing "Target Velocity (m/s)" with Carriers(fighters loaded)

### Expected behavior:
Show the dps graph after changing the "Target velocity"

### Actual behavior:

Nothing shows in the graph, and bottom line shows "Invalid values in 'Nidnoggur fit'

### Detailed steps to reproduce:

1. Create new Nidhoggur fit
2. fit 3 squad of Einherji, with active ability missile attack, Turret attack
3. open graph
4. change the "Target Velocity (m/s) to any value, e.g. 12
5. dps line disappear, "Invalid values in 'Nidnoggur fit"

### Fits involved in EFT format (Edit > To Clipboard > EFT):
[Nidhoggur, Nidhoggur fit]

[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]

[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]

[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]

[Empty Rig slot]
[Empty Rig slot]
[Empty Rig slot]

Einherji I x9
Einherji I x9
Einherji I x9

### Release or development git branch? Please note the release version or commit hash:

pyfa v2.0.0b5

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):

Windows 10

blitzmann:
Thanks for the bug report! A quick peak displays this exception message:

`gui.graphFrame: math domain error`

Hell if I know what that mean though, but I'll look into it :)

blitzmann:
Got this fixed with the help of @watchmecalculate pointing me in the right direction :D Thanks!

PR #1612, opened 2018-05-27
Changed fighter missile prefix to correctly reflect database attribute

WatchMeCalculate:
After investigating #1521 I found the issue that was occurring was due to in gui\builtinGraphs\fitDps.py#160  that  damageReductionFactor and damageReductionSensitivity were returning 0 which causes the log function to fail. It seems the value for this is no longer stored with the prefix fighterAbilityMissiles but instead fighterAbilityAttackMissile. After making this change velocityFactor was correctly calculated.

### DIFF
--- a/eos/effects/fighterabilitymissiles.py
+++ b/eos/effects/fighterabilitymissiles.py
@@ -7,7 +7,7 @@ effects, and thus this effect file contains some custom information useful only 
 displayName = "Missile Attack"
 
 # Attribute prefix that this ability targets
-prefix = "fighterAbilityMissiles"
+prefix = "fighterAbilityAttackMissile"
 
 type = "active"
 
--- a/eos/effects/fighterabilitymissiles.py
+++ b/eos/effects/fighterabilitymissiles.py
@@ -7,7 +7,7 @@ effects, and thus this effect file contains some custom information useful only 
 displayName = "Missile Attack"
 
 # Attribute prefix that this ability targets
-prefix = "fighterAbilityAttackMissile"
+prefix = "fighterAbilityMissiles"
 
 type = "active"
 


blitzmann:
I don't believe this is the proper fix.

Fighters (at these these fighters, the Einherji), have three modules to activate (along with their prefix):

Turrets -- `fighterAbilityAttackMissile`
Microwarpdrive - `fighterAbilityMicroWarpDrive`
Missile -- `fighterAbilityMissile`

As you can see, the turrets are technically `fighterAbilityAttackMissile` (doesn't make much sense to me either lol, way to go with the naming CCP. I think it's because they both act like missiles in the game? I don't remember). If we implement this, I think you effectively turn the missile attack into the turret attack, since it's now using the Turret's prefix.

That said, I haven't looked into this issue much at all. But it looks to me that, in the case of turrets, the value is under `fighterAbilityAttackMissileReductionFactor` instead of  `fighterAbilityAttackMissileDamageReductionFactor`. You can probably do something like getting the first one and then attempting to get the second one if the first one fails...

WatchMeCalculate:
Ah yep I agree with your assignment, their naming conventions are whack. I confirmed that this is only happening on evaluating fighters missile attack.  I did some queries against ESI and it is in fact being populated:
id 2128: fighterAbilityMissilesDamageReductionSensitivity
id 2127: fighterAbilityMissilesDamageReductionFactor
```
{
  "capacity": 0,
  "description": ... ,
  "dogma_attributes": [
    ...
    {
      "attribute_id": 2128,
      "value": 5.5
    },
    {
      "attribute_id": 2127,
      "value": 5
    },
....
  ],
  "graphic_id": 2899,
  "group_id": 1652,
  "market_group_id": 840,
  "mass": 1000,
  "name": "Templar II",
  "packaged_volume": 1000,
  "portion_size": 1,
  "published": true,
  "radius": 35,
  "type_id": 40556,
  "volume": 1000
}
```
I also found that it is populated correctly in the eve database as well. So for some reason ability.fighter.getModifiedItemAttr for these two attributes is returning 0

blitzmann:
> So for some reason ability.fighter.getModifiedItemAttr for these two attributes is returning 0

It's probably because for the turret attack, it's `fighterAbilityMissilesReductionSensitivity` rather than `fighterAbilityMissilesDamageReductionSensitivity` (same for the other attribute). We probably assum in the code that it's going to contain the work "damage" when that's not entirely the case.

I'll take a look at this later today - if my hunch is correct, should be an easy fix

blitzmann:
Took a look at the code:

```
damageReductionFactor = ability.fighter.getModifiedItemAttr("{}ReductionFactor".format(prefix))

        # the following conditionals are because CCP can't keep a decent naming convention, as if fighter implementation
        # wasn't already fucked.
        if damageReductionFactor is None:
            damageReductionFactor = ability.fighter.getModifiedItemAttr("{}DamageReductionFactor".format(prefix))

        damageReductionSensitivity = ability.fighter.getModifiedItemAttr("{}ReductionSensitivity".format(prefix))
        if damageReductionSensitivity is None:
            damageReductionSensitivity = ability.fighter.getModifiedItemAttr(
                    "{}DamageReductionSensitivity".format(prefix))

        targetSigRad = explosionRadius if targetSigRad is None else targetSigRad
        sigRadiusFactor = targetSigRad / explosionRadius
```

So this seems like a problem with our recent conversion form py2 to py3, in that None and 0 aren't really the same thing anymore (`getModifiedAttr` used to return `None` if the attribute didn't exist, but now returns 0)

This seems like a proper fix:

9f560584cedd52e0ea388fb1e1271bd59ff13bb1

Default to None if the attribute can't be found

Issue #1599, opened 2018-05-20
asterisk / star indicating character was modified doesn't go away after sync


IslayTzash:
## Bug Report

I modified some character skills using the right click "change module skills" menu under a module.   I rolled back using the green arrow synch character button.   The star remains next to my name when I expect it to clear since the character is no longer modified.  The skills do rollback, I see the DPS numbers go back to their original values.

### Release or development git branch? Please note the release version or commit hash:

Rel 2.0.1.

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):

Win 10.

### Other relevant information:

Using ESI with local SSO and my own dev app key.

![image](https://user-images.githubusercontent.com/18371221/40282255-6fb26b56-5c21-11e8-82b5-db408c6327f2.png)

blitzmann:
Thanks for the bug report! I've known about this for a while, haven't had a chance to really look into it. >.<

Should be a pretty simple fix though, I just need to sit down and deal with it lol

WatchMeCalculate:
Looks like when refresh API button is pressed in and the UpdateAPIThread is called in [service\character.py](https://github.com/pyfa-org/Pyfa/blob/master/service/character.py#L483) The clearSkills function doesn't also clear dirtySkills [eos\savedata\character](https://github.com/pyfa-org/Pyfa/blob/master/eos/saveddata/character.py#L121)

PR #1616, opened 2018-05-28
Remove asterisk next to character name after API refresh

WatchMeCalculate:
For #1599 , When API Refresh is pressed and clearSkills on the character is called in the Update thread, also clear dirty skills.


blitzmann:
Huh... interesting. Makes sense though, thanks!

This is a great fix for the issue, but I believe there's other issues regarding the character's name keeping an asterisk. (like, when copying character I believe).

I'll have to do some research on that and open up another issue regarding it.

Issue #1519, opened 2018-04-10
Disable/Enable fitting restrictions


SubjectXXX:
## Bug Report
Modules get unfit on re-enabling fitting restrictions

### Expected behavior:
Nothing happens

### Actual behavior:
First Rocket launcher and Assault Damage Control II gets removed

### Detailed steps to reproduce:
Load this fit. Disable fitting restrictions, enable them right back.

### Fits involved in EFT format (Edit > To Clipboard > EFT):
[Jaguar, SB]

Navy Micro Auxiliary Power Core
Assault Damage Control II
Caldari Navy Power Diagnostic System

Pithum B-Type Medium Shield Booster
Thukker Medium Cap Battery
Initiated Compact Warp Scrambler
1MN Y-S8 Compact Afterburner
[Empty Med slot]

Rocket Launcher II, Nova Rage Rocket
Rocket Launcher II, Nova Rage Rocket
Rocket Launcher II, Nova Rage Rocket
[Empty High slot]

Small Anti-Kinetic Screen Reinforcer II
Small Ancillary Current Router II

Warrior II x2

### Release or development git branch? Please note the release version or commit hash:
1.36.0

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
win10

blitzmann:
Thanks for the bug report! Indeed, this is no expected behavior. I'll take a look into why this might be happening!

WatchMeCalculate:
Initial investigation of at least the rocket launcher disappearance. This occurs for all shipTypes as well as modules being Turret or Missile. Basically, when you toggle the function it triggers [sevice\fit.py](https://github.com/pyfa-org/Pyfa/blob/master/service/fit.py#L974) to this code block: 
```
 if not fit.ignoreRestrictions:
        for m in fit.modules:
            if not m.isEmpty and not m.fits(fit):
                self.removeModule(fit.ID, m.modPosition)
```
In [eos\savedata\module.py](https://github.com/pyfa-org/Pyfa/blob/master/eos/saveddata/module.py#L479) as it's validating each module on the first Hardpoint module it finds the hardpoint check fails
```
 if fit.ship.getModifiedItemAttr('launcherSlotsLeft') - fit.getHardpointsUsed(Hardpoint.MISSILE) < 1:
``` 
[...]

PR #1626, opened 2018-06-01
Issue #1519, on fitting restriction toggle, accurately keep valid modules

WatchMeCalculate:
Ensures on toggling fitting restrictions we no longer check if a fit module could be fitted to ship it's already fit on, but instead that it is a valid module for given ship type. When module gets colored red for being one-too-many, ensure that when re-enabled that background gets reset. There is more in #1519 to illustrate the issues


WatchMeCalculate:
side note: generally I wouldn't use `type` as a variable name in python because it masks the built-in function even if just for the scope of the function, but I went with conformity with the functions nearby.

Issue #1627, opened 2018-06-01
Crash when attempting a char import on OSX version


cliveflint:
## Bug Report
Character import on OSX version fails when pressing the Add Character button

### Expected behavior:
Not done it before but I expect it should allow me to import a character

### Actual behavior:
OS version: Darwin-17.5.0-x86_64-i386-64bit
Python version: 3.6.3 (default, Oct 22 2017, 22:24:50) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.72)]
wxPython version: 4.0.0b2 (wxWidgets 3.0.4)
SQLAlchemy version: 1.1.10
Logbook version: 1.0.0
Requests version: 2.18.4
Dateutil version: 2.6.1

####################

Traceback (most recent call last):
  File "/Users/ryan/pyfa/gui/characterEditor.py", line 800, in addCharacter
  File "/Users/ryan/pyfa/service/esi.py", line 111, in login
  File "/Users/ryan/pyfa/service/esi.py", line 129, in startServer
  File "/usr/local/Cellar/python3/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/socketserver.py", line 453, in __init__
  File "/Users/ryan/pyfa/service/server.py", line 113, in server_bind
  File "/usr/local/Cellar/python3/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/socketserver.py", line 467, in server_bind
socket.gaierror: [Errno 8] nodename nor servname provided, or not known

####################

Traceback (most recent call last):
  File "/Users/ryan/pyfa/gui/esiFittings.py", line 331, in addChar
  File "/Users/ryan/pyfa/service/esi.py", line 111, in login
  File "/Users/ryan/pyfa/service/esi.py", line 129, in startServer
  File "/usr/local/Cellar/python3/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/socketserver.py", line 453, in __init__
  File "/Users/ryan/pyfa/service/server.py", line 113, in server_bind
  File "/usr/local/Cellar/python3/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/socketserver.py", line 467, in server_bind
socket.gaierror: [Errno 8] nodename nor servname provided, or not known

####################

Traceback (most recent call last):
  File "/Users/ryan/pyfa/gui/esiFittings.py", line 331, in addChar
[...]

PR #1632, opened 2018-06-06
Adds exception handler for when HttpServer fails on init

WatchMeCalculate:
This will catch errors that happen on init of StoppableHttpServer, such as in Issue #1627, and display helpful prompt that suggests trying manual server settings. This is probably a bit too broad of a catch, but hopefully some tracebacks get posted to help sift through the noise


blitzmann:
Sorry about the wait on this! Thanks for the PR! :)

Issue #1597, opened 2018-05-20
Mutaplasmids and Abysaal modules


blitzmann:
WARNING: this feature is being developed on the `mutaplasmids` branch. If you're interested in checking it out, feel free, however be warned that this has changes to the user database, It is unknown if downgrading back to a stable release would work

---

Initial discussion was started here: #1530

Mutated items seem to be a completely new item (different item ID). These items are not populated with the required attributes (see below)... CCP keeps track of what the base item was, and I assume will be using the base item and simply substituting the dynamic attribute value over the "mutatable" ones.

```
[+] 10MN Abyssal Afterburner
  [+|n] medPower
  [+|y] moduleBonusAfterburner
  [+|n] online
  [+|y] overloadSelfSpeedBonus
  [+] capacity: 0
  [+] mass: 1.0
  [+] radius: 1.0
  [+] requiredSkill1: 3450.0
  [+] requiredSkill1Level: 4.0
  [+] volume: 25.0
```

At it's most basic form, I think this would be a relatively easy thing to accomplish. pyfa could have a `mutators` table that is linked to a `module.ID` and references an attribute with it's new value, and when the module loads, it loads any mutators that it has with it. Should look into the current attribute override functionality to see if there's anything in there that we can reuse, or if this is going to require something new.

The biggest problem right now is the fact that Multiplasmid definitions are not available from CCP yet. I've requested access to this data, and also want it implemented via ESI (see https://github.com/esi/esi-issues/issues/928). But without this data, I'm not able to programatically look up which items have multiplasmids, and which attributes are enabled to be "mutatable". 

I was able to find the RES file that I think contains these definitions bu looking at the resfileindex, however it's in FSD format and I don't have the skills to decode it. So until this data is available for us, this feature is going to be on hold. Technically I could work on certain assumptions, but I don't want to get caught in a situation where my assumptions are completely wrong and have to rework the feature once the data becomes available. :/

## Import / Export

The mutated modules are literally a new type, so if you have a mutated afterburner on your rupture, pyfa will currently assume that it's the literal item. That being said, sounds like CCP are working on an ESI endpoint that allows you to get the base item for an `itemID`, as well as the attributes for that mutated module. This means that we should be able to import, look up the item to get and convert to the base module, and then add the mutators. CCP says that sometime this week (week of 5/20) they should have an idea on what this endpoint looks like.

blitzmann:
Need to delve into this a bit, but thanks to @inomares:

[...]

PR #1643, opened 2018-06-13
Added reset and randomize buttons to mutaplasmid view

burnsypet:
#1597 

---

Added new methods for resetting and randomizing the attributes of an Abyssal item. Two new methods resetMutatedValues() and randomMutatedValues are included in itemMutator.py

It's randomly generating uniformly distributed floats at the moment, but it can easily be changed to ints if needed.


blitzmann:
Thanks! Will check this out and merge :) As for uniformity, that's fine for now. i think I read somewhere that it's actually a bell curve (so, `random.triangular()` I think), however that hasn't been confirmed

Issue #1631, opened 2018-06-05
Cannot import EVE API XML character files


fsufitch:
### Expected behavior:

Using `File > Import Character File` and giving it an XML character file successfully imports it as a character to use with fittings.

### Actual behavior:

Nothing happens. The logs show nothing.

### Detailed steps to reproduce:

1. Get the XML file inside this ZIP: [Minmatar Starter.zip](https://github.com/pyfa-org/Pyfa/files/2073337/Minmatar.Starter.zip).
2. Start Pyfa
3. `File > Import Character File` and select the XML file.
4. Observe as nothing happens.

### Release or development git branch? Please note the release version or commit hash:

![image](https://user-images.githubusercontent.com/4542769/40991913-390e1312-68c3-11e8-87f5-b1e3d40f8827.png)

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):

Windows 10 Professional, Version 1803, Build 17134.48

blitzmann:
![image](https://user-images.githubusercontent.com/3904767/40992344-8a97f490-68c4-11e8-9f19-1dacd5772f48.png)

lol

I did this when stripping out EVE XML stuff, commented this out so that I could evaluate later. later never came.

Should be easy enough to fix this though

WatchMeCalculate:
Was looking into this a bit and saw that import is also failing on this particular xml file with `xml.parsers.expat.ExpatError: not well-formed (invalid token): line 1, column 1`

WatchMeCalculate:
Well barring the return statement in run(), there were some issues with the xml. Loading it into an interpreter it had some funky unicode going on `'ï»¿<?xml version="1.0"?>`  After sanitizing that it appeared that saveddata/character.py was missing apiUpdateCharSheet.

blitzmann:
Merged in the fix from @WatchMeCalculate, thanks!

fsufitch:
Closing this since it's fixed.

blitzmann:
I actually like to keep these things open until a release is made. helps with writing up the changelog, and also shows people on current release that this known issue has been addressed :D

fsufitch:
Roger, my bad :)

PR #1652, opened 2018-06-19
Address Issue where evemon xml was not being imported

WatchMeCalculate:
This addresses issue #1631

### DIFF
--- a/eos/saveddata/character.py
+++ b/eos/saveddata/character.py
@@ -125,6 +125,12 @@ class Character(object):
 
         return all0
 
+    def apiUpdateCharSheet(self, skills, secStatus=0.00):
+        self.clearSkills()
+        for skillRow in skills:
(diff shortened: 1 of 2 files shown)


blitzmann:
Thanks!

I feel like there was a reason I removed `apiUpdateCharSheet`... think it was to make room for however the ESI one works, since the XML api is no longer. Will have to take a deeper look at it :)

WatchMeCalculate:
Yeah seems reasonable, I thought it might be useful for evemon imports as
that project has been updated for ESI, but perhaps they don’t support
exports into xml format anymore.

blitzmann:
Haven't really had a chance to take that "deeper look" lol, but it should be benign enough to add this back into the fray. Should re-evaluate the code behind this at some point and clean it up sometime, but... *kicks can down the road*

PR #1660, opened 2018-07-02
Make pyfa retina ready

flashspys:
This should help pyfa to look better on macOS 📺


flashspys:
Hey @blitzmann ! 
I added the entry to the Info.plist locally, and this helped at least a bit. The font is much better but the icons are still in a very bad resolution. Better than nothing. 
I never built pyfa with this changes. But if you can build pyfa and the high resolution entry is in the info plist, then it should work

blitzmann:
The next release will include this change, so we'll see if we get any reports of funkyniess :D

PR #1663, opened 2018-07-07
Fixes the Helios's drone damage bonus.

MaruMaruOO:
The required skill should be Light Drone Operation as light drones don't require Drone Avionics.
This causes the damage bonus to apply correctly. Thus giving a max skill Hobgoblin II 29.7dps rather than 19.8dps.


blitzmann:
Who are you!? You seemed to have randomly shown up recently and started fixing bugs, I love it! :P

Issue #1662, opened 2018-07-06
Bug for new drone implant


AlanTemis:
cant add new drone implant (CreoDron 'Yellowjacket' Drone Tuner D5-10T

OS version: Windows-7-6.1.7601-SP1
Python version: 3.6.1 (v3.6.1:69c0db5, Mar 21 2017, 17:54:52) [MSC v.1900 32 bit (Intel)]
wxPython version: 4.0.0b2 (wxWidgets 3.0.4)
SQLAlchemy version: 1.1.10
Logbook version: 1.0.0
Requests version: 2.18.4
Dateutil version: 2.6.0

####################

Traceback (most recent call last):
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\builtinAdditionPanes\implantView.py", line 79, in OnRadioSelect
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\service\fit.py", line 1003, in toggleImplantSource
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\service\fit.py", line 1253, in recalc
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\eos\saveddata\fit.py", line 856, in calculateModifiedAttributes
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\eos\saveddata\implant.py", line 98, in calculateModifiedAttributes
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\eos\effects\skillbonusdroneinterfacing.py", line 9, in handler
AttributeError: 'Implant' object has no attribute 'level'

PR #1665, opened 2018-07-09
Allow drone skill effects to process implants

MaruMaruOO:
This should provide a fix for #1662 and allow similar drone implants to function without additional changes.
I tested the Bumblebee implant in game found it does impact fighters.
Given the descriptions only mention drones not fighters CCP may change that as a bug fix.
For now however the effects are applied to fighters to match actual gameplay.

### DIFF
--- a/eos/effects/skillbonusdronedurability.py
+++ b/eos/effects/skillbonusdronedurability.py
@@ -1,12 +1,14 @@
 # skillBonusDroneDurability
 #
 # Used by:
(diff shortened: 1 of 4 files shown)


blitzmann:
> Given the descriptions only mention drones not fighters CCP may change that as a bug fix.

I can file a bug report with CCP to see if this is the actual case.

Thanks a bunch for your bug fixes @MaruMaruOO! Really appreciated!

PR #1671, opened 2018-07-12
Added missing data to fit copying

MaruMaruOO:
This should cover everything currently not copied when using the copy button.
Specifically it should now copy booster side effects, fighter abilities and command fits along with the active state for fighters, projected fits and command fits.


blitzmann:
Oh wow, I didn't realize copying was missing a bunch of these things. Yay!

Code seems concise and makes sense, will test this out and merge it in when I

Issue #1673, opened 2018-07-16
Cap battery mutaplasmid error


NomadThanatos:
### Expected behavior:
Open mutaplasmid stats UI

### Actual behavior:
Error

### Detailed steps to reproduce:
1. Get any cap battery fit on a ship.
2. Apply mutaplasmid.
3. Module stats.
4. Oops, error.

### Fits involved in EFT format (Edit > To Clipboard > EFT):
[Vargur, Test]

Republic Fleet Gyrostabilizer
Republic Fleet Gyrostabilizer
Republic Fleet Gyrostabilizer
Republic Fleet Gyrostabilizer
Tracking Enhancer II

Gist X-Type 500MN Microwarpdrive
Adaptive Invulnerability Field II
Sensor Booster II, Scan Resolution Script
X-Large Abyssal Shield Booster
Large Abyssal Cap Battery
Tracking Computer II, Optimal Range Script

800mm Repeating Cannon II, Republic Fleet Fusion L
800mm Repeating Cannon II, Republic Fleet Fusion L
800mm Repeating Cannon II, Republic Fleet Fusion L
800mm Repeating Cannon II, Republic Fleet Fusion L
Bastion Module I
[Empty High slot]
[Empty High slot]
[Empty High slot]

Large Capacitor Control Circuit II
Large Capacitor Control Circuit II

Warrior II x5
Hammerhead II x5

### Release or development git branch? Please note the release version or commit hash:
Version v2.3.0

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
Windows-10-10.0.17134-SP0

### Other relevant information:
OS version: Windows-10-10.0.17134-SP0
Python version: 3.6.1 (v3.6.1:69c0db5, Mar 21 2017, 17:54:52) [MSC v.1900 32 bit (Intel)]
wxPython version: 4.0.0b2 (wxWidgets 3.0.4)
SQLAlchemy version: 1.1.10
Logbook version: 1.0.0
Requests version: 2.18.4
Dateutil version: 2.6.0

####################

Traceback (most recent call last):
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\contextMenu.py", line 138, in handler
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\builtinContextMenus\itemStats.py", line 69, in activate
[...]

PR #1677, opened 2018-07-16
Have missing icons display as question mark icons

MaruMaruOO:
Fix for #1673 
This will cause attempts to get non-existent icons via BitmapLoader.getStaticBitmap to return a question mark icon (0.png) rather than fail. This is effectively the same behavior seen under the attributes tab already.
This specific issue was caused by the 'Capacitor Warfare Resistance Bonus' having no icon.
There's a blank space in place of the icon when hovering over an Abyssal Cap Battery in game.

### DIFF
--- a/gui/bitmap_loader.py
+++ b/gui/bitmap_loader.py
@@ -49,7 +49,7 @@ class BitmapLoader(object):
     @classmethod
     def getStaticBitmap(cls, name, parent, location):
         static = wx.StaticBitmap(parent)
-        static.SetBitmap(cls.getBitmap(name, location))
+        static.SetBitmap(cls.getBitmap(name or 0, location))
         return static
 
     @classmethod


blitzmann:
So normally if they don't have an icon, they are NULL, but this one is actually just an empty string, correct?

hrm...

Seems like a quick fix, but want to keep it open for now to remind myself when I get home to check to see if this is an EVE data error or an error with our scripting to compile the data

MaruMaruOO:
Not quite, the iconID is null in the database.
There's a check in gui.builtinItemStatsViews.itemAttributes to confirm the iconID is not None before calling getBitmap. If it's None it instead calls it with a "0".
We could put the null/None check in gui.builtinItemStatsViews.itemMutator instead but I see no advantage in doing so.

blitzmann:
Ah, got it, it was the icon for the attribute, not the item itself. 👍

Issue #1624, opened 2018-05-31
Ability to export ship loadout to clipboard with calculated stats


molbal:
Hello,

I would like to request a feature, if possible, please. I would like to add an export function (based on the XML or JSON exporter) that exports the fit and the calculated stats as well (tank/dps/etc)

Thank you,
Veetor

blitzmann:
This would be pretty difficult to do, I think. Well, not difficult, but time consuming, and a bit of thought to get right. The stats are from various different places and all aggregated into one location (usually). Then there's the fact that maintaining this mapping - if we ever add a stat, we would need remember to add it to the export as well.. I wonder if there's a python decorator we can use to mark a property as an exportable property... Hrm...

You basically want the information from the Stats Panels, correct? Or are you thinking something much deeper, such as the specific damage of a charge?

molbal:
Hey, thanks for the answer! 

I was thinking of just first glance stats, like EHP and DPS and speed, no need for all the stats. (at least for me) Unfortunately I can't comment on the decorator, because my python knowledge is nonexistent :(

Cheers

blitzmann:
I'll give this some thought, but I can't promise anything, especially soon : )

molbal:
Thank you if you can squeeze it in, no worries if you can't!

blitzmann:
So, haven't reaally started with getting an actual system down to export calculated stats... however, I did use `jsonpickle` which converts python objects to json. This produces some interesting output.

https://gist.github.com/blitzmann/f690773185dd00b0f4cd4d10d5ca642e

This is definitely not the right way to do this. But might be useful? It basically gives you everything about the fit, internal information and all

![image](https://user-images.githubusercontent.com/3904767/42743024-2e6ebd2c-888e-11e8-94a9-1b51fa0f9e13.png)

Could be interesting to use this as a "debug" export

blitzmann:
lol just realized this might not actually work lol. A lot of the thigns we display are using getter functions, which this obv won't access.

Still interesting to see the object laid out. It waas a very simple fit, not very complicated, and most online json viewer  / formatter had difficulty loading it lol

molbal:
Thanks for checking it. It's a good idea and certainly good enough to work, but I doubt it the 5+ megabyte JSON files will be quick to parse so I'll stick with the standard EFT export and ask the others to copy the fields manually. (Or maybe experiment with an RPA based solution like uipath community edition)

blitzmann:
[...]

PR #1688, opened 2018-07-18
Adds copy to clipboard for Eve Fleet Simulator

MaruMaruOO:
Allows users to export the Eve Fleet Simulator(efs) data for the current fit by copying it to the clipboard.
efs_port
It's then pasted into efs like so:
efs_example1

Happy to make any needed changes and answer any questions.
This can also be discussed via an issue or elsewhere if preferred.

Preemptive answers to some possible questions:

What is efs?
Primarily a simulator to determine the pvp effectiveness of user created fleets. 
It can also provide a data visualization tool to quickly compare different ship types.

Why haven't I heard of it?
It's not released yet however it will launch soon and is intended as an open source project.

Why are you adding export functionality to pyfa for something that's not even released?
Efs can't create new fits internally and instead relies on pyfa to give it fit data. This leaves efs 
fairly inaccessible unless there's an easy way to export fits from pyfa.

Why such a heavy dependency on pyfa?
Simulating fleet combat requires many calculated stats for each fit. There are already multiple fitting 
tools/assistants that do this for eve and I don't think there's merit in reinventing the wheel.
Using pyfa also makes it easier to maintain and avoids splitting dev resources.

A picture of an in progress simulation for context:
efs_example2


blitzmann:
Neato! Can't review or even process much of this right now, but will take a look! More export formats are awesome, and I would be happy to include support for a new service as it releases :D

I'm guessing this is a separate project that you're working on @MaruMaruOO?

And taking a brief look at the format in that picture, seems it could be a good contender to solve for #1624 as well. Two birds with one stone, etc :D

MaruMaruOO:
That's great to hear, and of course I expect this type of thing to take time to get reviewed. Especially given it's a fairly large patch for an export format and I just uploaded it out of nowhere.

Yep this is a separate project I've been working on for awhile now.

As for #1624 yeah actually I think it might fill that role quite nicely.
I intentionally want to keep it human readable and editable anyway.
Being JSON it's a bit painful to read as plain text but it's easily/widely processed.
Given the way it's set up a variant for that specific purpose would also be very easy to create if the base wound up not being suitable.

service/efsPort.py
```diff
@@ -0,0 +1,630 @@
...
+        return {
+            "usingMWD": False,
+            "unpropedSpeed": fit.maxSpeed,
+            "unpropedSig": fit.ship.itemModifiedAttributes["signatureRadius"]
+        }
+
+    @staticmethod
+    def getOutgoingProjectionData(fit):
+        # This is a subset of module groups capable of projection and a superset of those currently used by efs
+        modGroupNames = [
```
blitzmann: Not sure if it's helpful, but you're able to check to see if a module can be projected by using `item.isType("projected")` (I think)(so you would use that in your `for mod in projectedMods:` loop). Maintaining a hardcoded list like this won't be easy. Granted, doing it that way means you might pick up projected modules that EFS can't process, but in that case you might want to implement a checks on the EFS to skip the entry if it's a type you're not expecting.

service/efsPort.py
```diff
@@ -0,0 +1,630 @@
...
+        projectedMods = list(filter(lambda mod: mod.item and mod.item.groupID in modGroupIds, fit.modules))
+        projections = []
+        for mod in projectedMods:
+            stats = {}
+            if mod.item.groupID in [modGroupData["Stasis Web"]["id"], modGroupData["Stasis Grappler"]["id"]]:
+                stats["type"] = "Stasis Web"
+                stats["optimal"] = mod.itemModifiedAttributes["maxRange"]
+                EfsPort.attrDirectMap(["duration", "speedFactor"], stats, mod)
+            elif mod.item.groupID == modGroupData["Weapon Disruptor"]["id"]:
+                stats["type"] = "Weapon Disruptor"
```
blitzmann: You might be able to use `mod.item.group.name` here to avoid hardcoded values, depending on how it applies to your application :)

service/efsPort.py
```diff
@@ -0,0 +1,630 @@
...
+            if mod.itemModifiedAttributes["maxRange"] is None:
+                pyfalog.error("Projected module {0} has no maxRange".format(mod.item.name))
+            stats["optimal"] = mod.itemModifiedAttributes["maxRange"] or 0
+            stats["falloff"] = mod.itemModifiedAttributes["falloffEffectiveness"] or 0
+            EfsPort.attrDirectMap(["duration", "capacitorNeed"], stats, mod)
+            projections.append(stats)
+        return projections
+
+    @staticmethod
+    def getModuleNames(fit):
```
blitzmann: Think this was mentioned in the related issue, might be better to use a list of objects that represent the module, rather than a list of names. So, something like

	[{
		typeID: 1234,
		typeName: "Module Name",
		slot: 'high',
		position: 3,
		...	
	},
	...
	]

That would avoid having blank lines,m or lines labeling the rack. The data above could be used to reconstruct the same order, with knowledge of what rack it's in, and would be more universally applicable for other's who might look to utilize this format.

service/efsPort.py
```diff
@@ -0,0 +1,630 @@
...
+    def exportEfs(fit, groupID):
+        includeShipTypeData = groupID > 0
+        fitID = fit.ID
+        if includeShipTypeData:
+            fitName = fit.name
+        else:
+            fitName = fit.ship.name + ": " + fit.name
+        pyfalog.info("Creating Eve Fleet Simulator data for: " + fit.name)
+        sFit = Fit.getInstance()
+        sFit.recalc(fit)
```
blitzmann: Don't think you need to recalc... once you get to this spot, the fit should have already been loaded and calc'd in pyfa. That being said, it doesn't really hurt.

service/efsPort.py
```diff
@@ -0,0 +1,630 @@
...
+    @staticmethod
+    def getPropData(fit, sFit):
+        fitID = fit.ID
+        propGroupId = getGroup("Propulsion Module").ID
+        propMods = filter(lambda mod: mod.item and mod.item.groupID == propGroupId, fit.modules)
+        activePropWBloomFilter = lambda mod: mod.state > 0 and "signatureRadiusBonus" in mod.item.attributes
+        propWithBloom = next(filter(activePropWBloomFilter, propMods), None)
+        if propWithBloom is not None:
+            oldPropState = propWithBloom.state
+            propWithBloom.state = 0
```
blitzmann: Probably best to import the `State` Enum and use that instead of the hardcoded value here, for explicitness

service/efsPort.py
```diff
@@ -0,0 +1,630 @@
...
+        sFit.removeModule(fitID, mwdPosition)
+        sFit.recalc(fit)
+        fit = eos.db.getFit(fitID)
+        return mwdPropSpeed
+
+    @staticmethod
+    def getPropData(fit, sFit):
+        fitID = fit.ID
+        propGroupId = getGroup("Propulsion Module").ID
+        propMods = filter(lambda mod: mod.item and mod.item.groupID == propGroupId, fit.modules)
```
blitzmann: Could just do `mod.item.group.name == 'Propulsion Module'` instead of having to get the group first, but doesn't really matter. Six of one and half a dozen of the other.

Additionally... I'm seeing a lot of `filter/lambda` combinations for narrowing lists down. The more conventional / pythonic way would be to use list comprehension. So, instead of

`propMods = filter(lambda mod: mod.item and mod.item.groupID == propGroupId, fit.modules)`

it would be something like

`propMods = [mod for mod in fit.modules if mod.item and mod.item.groupID == propGroupId]`

service/efsPort.py
```diff
@@ -0,0 +1,630 @@
...
+
+    @staticmethod
+    def getT2MwdSpeed(fit, sFit):
+        fitID = fit.ID
+        propID = None
+        shipHasMedSlots = fit.ship.itemModifiedAttributes["medSlots"] > 0
+        shipPower = fit.ship.itemModifiedAttributes["powerOutput"]
+        # Monitors have a 99% reduction to prop mod power requirements
+        if fit.ship.name == "Monitor":
+            shipPower *= 100
```
blitzmann: Looks like we're attempting to do some fitting calculations here for the Monitor...? Is this something that isn't implemented in the regular pyfa calcs?

blitzmann:
Took a quick look and did a review of the code, noticed a few things that I'm interested in learning more about / clarifying.

I know this format is in flux, please let me know when you would consider it "done" and we can do a final review. :)

service/efsPort.py
```diff
@@ -0,0 +1,630 @@
...
+from service.market import Market
+from eos.enum import Enum
+from eos.saveddata.module import Hardpoint, Slot, Module
+from eos.saveddata.drone import Drone
+from eos.effectHandlerHelpers import HandledList
+from eos.db import gamedata_session, getItemsByCategory, getCategory, getAttributeInfo, getGroup
+from eos.gamedata import Category, Group, Item, Traits, Attribute, Effect, ItemEffect
+from logbook import Logger
+pyfalog = Logger(__name__)
+eos.db.saveddata_meta.create_all()
```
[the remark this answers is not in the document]
MaruMaruOO: A separate script runs exportEfs with an unfitted copy of every ship in eve as part of the EFS build process.
I'll move it out of efsPort as you've correctly noticed that it's really in the wrong place.

service/efsPort.py
```diff
@@ -0,0 +1,630 @@
...
+            {"name": "Misc", "groupIDs": map(lambda s: getGroup(s).ID, miscGroupNames)}
+        ]
+        for size in shipSizes:
+            if groupID in size["groupIDs"]:
+                return size["name"]
+        sizeNotFoundMsg = "ShipSize not found for groupID: " + str(groupID)
+        return sizeNotFoundMsg
+
+    @staticmethod
+    def exportEfs(fit, groupID):
```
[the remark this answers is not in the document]
MaruMaruOO: It's non-zero when we want data about the ship as a hull not a specific fit.
It's normally used with empty fits when building EFS.
I'll rename it and get fits to export an actual groupID correctly.

service/efsPort.py
MaruMaruOO: Well groups that aren't handled explicitly won't have a useful set of values anyway.
pyfa considers a lot of stuff to be "projected" that really don't belong here. Missile launchers for example.
It also doesn't cover all the stuff I actually do want like Phenomena Generators and MJDs.
Overall I'm not sure if there's much of a maintenance advantage to using isType("projected") when I'd still need to add any new cases and keep a doc list of what it deals with.

service/efsPort.py
MaruMaruOO: Good point, that cleans the whole thing up and massively simplifies getting the projectedMods list itself.

service/efsPort.py
MaruMaruOO: Same situation as create_all(), I'll move it out of EfsPort.

service/efsPort.py
MaruMaruOO: Will do, I intended to use them where possible but I missed this one.

service/efsPort.py
MaruMaruOO: Yeah I was looking at the db itself and wasn't aware that was a thing for most of my time making this. I'll change it accordingly.

I'll look into swapping some of the lambdas out.

service/efsPort.py
MaruMaruOO: The shipPower is being compared to the base power of MWDs to work out if the ship could plausibly fit them.
We would have to actually equip each option for pyfa's calcs to apply which would be much, much slower.
If the bonus becomes widespread I'll certainly want to change the implementation.

service/efsPort.py
MaruMaruOO: I'm not sure if you'd read my response to this in #1624 when you wrote this?
Due to performance constraints this is almost laughably bad from an EFS standpoint.
For context I'll mention that EFS runs in the browser.

I would much prefer to make a secondary format with another name if need be.

MaruMaruOO:
Thanks for the review.

On the file structure I did consider adding it to port.
I left it alone because as you say port is already a bit crowded and I didn't feel it warranted the largest python file in the project.
I think the ideal solution is probably to split port up and use a ./service/port/ directory to house the various formats. 
I'm not really concerned either way however and I can arrange it however you like.

Noted I'll swap the mod.itemModifiedAttributes references to use mod.getModifiedItemAttr.

Yep that sounds good to me, I'll let you know when I've finished patching stuff up.

service/efsPort.py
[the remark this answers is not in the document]
blitzmann: Ah, so you're incorporating this script as part of EFS build as well. That kind of explains a few things. :)

As a general rule of thumb, EFS-specific things should probably have a build script in that project that handles those things. It's totally cool to import bits and pieces of pyfa to bootstrap if it helps get EFS set up, but going forward I would be a bit reluctant to include code that doesn't have any functional purpose within pyfa proper. 😄

service/efsPort.py
blitzmann: Fair points :)

service/efsPort.py
blitzmann: I did not read the comment before (they didn't show up on my GitHub notifications???), however still a bit confused.

Obviously I don't know how EFS is handling that stuff at all, so this is all based on assumptions :P As I understand it, you're using the current mod names array as a simple reminder for the user. My confusion is the assumption that you can still build that same thing fairly easily if the export was in an array-of-objects format, regardless of if it's browser-based or not - just loop though the array, ordered by slot and position, and then build the mod name array from it. It indeed would increase the size of the export, but that isn't too much of a concern (at least for pyfa and I would assume the end user, unsure if there's some limitations on EFS's side)

I'm genuinely curious as to why doing it that way would carry huge performance impacts on EFS? You would know better than I, so of course I defer to you as to what's best in this case. I simply feel that having a list of objects would make the format more widely accessible -- but you might be right; I might have my head wrapped around the idea that this format should carry all this info, and that's probably just not it's intended purpose 😛

service/efsPort.py
blitzmann: kk, I think I'm picking up what you're putting down. It's just another case of me not fully understanding how EFS is using this data. 😛  Thanks for the clarification

blitzmann:
> I think the ideal solution is probably to split port up and use a ./service/port/ directory to house the various formats.

I've been thinking the same thing. No worries, was just wondering what the considerations were :)

Thanks for the clarifications on my notes, think I have a better understanding of the format as a whole and how it's being used.

Unrelated to pyfa, when are you planning on launching EFS? What technologies does it use? You mentioned it's browser-based, and my day job includes AngularJS, so I'm simply curious  :P

MaruMaruOO:
Slightly slow reply as I wanted to actually get some of this patched first.

To answer your first question I'm hoping to launch in 2-3 weeks.
Keep in mind it could run long, especially if it proves difficult to get good default fits by scraping zkill.

In terms of tech the EFS stack looks something like this:
react / jsx (no flux or immutable.js)
flow
semantic-ui
less (Required to sensibly manipulate some semantic-ui values. Actual page themes use css vars)
and eslint (airbnb settings with a couple of small changes)
I'm leaving off managers/bundles, smaller ui stuff, plugins and various other bits and pieces. 


For the patches I still have to do a lambda purge and write you an explanation on why EFS wants the format to stay smaller if practical.
Outside of that I think I covered all the needed changes.

Also a quick note on the getters for *ModifiedAttributes (eg `mod.getModifiedItemAttr`).
The returns are all structured as: `return return_value or default`
This means 0, False, 0.0, null/None ect to all return the default.
You can get the information via attributes anyway but I figured it might be unintentional given they have quite different meanings.
For example a ship that isn't cloaked and a random mod will both return 0 from `ship.getModifiedItemAttr('cloaked', 0)` and `mod.getModifiedItemAttr('cloaked', 0)`.
This is for actual values of False and None respectively.

service/efsPort.py
MaruMaruOO: Overall the typical use case for EFS involves far more fits than pyfa. To reduce the barrier to usefulness, make it viable on mobile and minimize tedium I intend for EFS to have several default fits for each ship.

Eve has a lot of ships so even a very modest number per ship makes for a fairly high total. For most situations increasing the size of the format by 50% has a very similar performance impact on EFS to adding 50% more fits. If those are my options I'll pick more default fits unless the number is already really high. If the change is smaller or it enhances the simulation it's not really a big deal but this is 50-60% and mostly formatting fluff.

In terms of design you more or less hit the nail on the head. It's just not it's intended purpose. Ultimately the main purpose of this object is to construct virtual ships. Thousands of these virtual ships have to be able to fight each other at up to a thousand times normal speed. There's no point in the exports having too much detail when the simulation can't ever use it without slowing to a crawl. I intentionally abstracted out the concept of modules in the first place for that exact reason.

Hopefully this clears things up. I can go into a more technical explanation if you want but I feel like most of it will be telling you things you know alongside long winded justifications for implementation choices that wouldn't change my overall opinion even if they weren't the case.

blitzmann:
@MaruMaruOO o/ sorry it's been a while, was away for a week at a conference and then was sick for the week after x_x. All the while focusing on the undo/redo functionality and the refactoring that comes with that (will be getting a beta build out for that very soon).

I feel like I've neglected this PR a bit, so wanted to pop in and say I'm still here, and actually currently reviewing it for inclusion into the release tonight - I would assume it's at least somewhat final since the last commit was 26 days ago, and if it's not, well, that's why I suggested a version field :P

Issue #1683, opened 2018-07-17
2.3.1 is missing all the boosters like standard exile and blue pill from the implants & boosters section.


taltracyn:

blitzmann:
Huh... 

You can search these boosters directly, but attempting to right-click> show market group produces this error:

```OS version: Windows-10-10.0.16299-SP0
Python version: 3.6.1 (v3.6.1:69c0db5, Mar 21 2017, 17:54:52) [MSC v.1900 32 bit (Intel)]
wxPython version: 4.0.0b2 (wxWidgets 3.0.4)
SQLAlchemy version: 1.1.10
Logbook version: 1.0.0
Requests version: 2.18.4
Dateutil version: 2.6.0

####################

Traceback (most recent call last):
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\contextMenu.py", line 138, in handler
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\builtinContextMenus\marketJump.py", line 51, in activate
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\marketBrowser.py", line 100, in jump
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\gui\builtinMarketBrowser\marketTree.py", line 92, in jump
wx._core.wxAssertionError: C++ assertion "item.IsOk()" failed at ..\..\src\msw\treectrl.cpp(1032) in wxTreeCtrl::GetItemParam(): invalid tree item
```

Seems like it could be an issue with the underlying data, almost like these were removed from the market tree somehow. Will check when I get home :)

taltracyn:
no its just 2.3.1

blitzmann:
Eh, it's also an issue in 2.3.0. I'll have to back through and look at what data changed between the Abyss expansion and the july patch ;/

MaruMaruOO:
The bug is caused by the unneeded "# Implants & Boosters > Booster" entries in service.market.ITEMS_FORCEDMARKETGROUP.
Fixing it causes boosters to correctly have multiple market groups but also exposes a bug in the way item variations are obtained.

This is primarily a note to say I'm working on this and should have a patch up in the next day.
I think I've got a working solution but it needs proper testing and cleaning up.
As a nice bonus it also looks like we should be able to remove ~60 lines of dead code.

Sadario:
[...]

PR #1689, opened 2018-07-19
Update market.service to reflect new booster market groups.

MaruMaruOO:
Fix for #1683 

Cause of the bug:
service.market.marketGroupHasTypesCheck is used to check if a market group has items and by extension if it has child market groups. 
If ITEMS_FORCEDMARKETGROUP contains a market group's ID then marketGroupHasTypesCheck will return true.
gui.builtinMarketBrowser.marketTree.expandLookup assumes groups with items cannot contain child groups.
Thus the hard coded marketGroupIDs corresponding to the 'Booster' group caused it's children to be ignored after changes by CCP moved boosters to child market groups based on slot and type.

Changes:
Updated booster IDs for ITEMS_FORCEDMARKETGROUP and removed entries that already have identical marketGroupIDs in the database.

Updated getVariationsByItems to correctly handle boosters for the new market groups. Note the variations_limiter system was not suitable for this, I can explain at length why if desired.

Conservatively changed marketGroupHasTypesCheck so similar database changes will only impact the items in ITEMS_FORCEDMARKETGROUP.

### DIFF
--- a/service/market.py
+++ b/service/market.py
@@ -527,7 +527,7 @@ class Market(object):
         categories = ['Drone', 'Fighter', 'Implant']
 
         for item in items:
-            if True and item.category.ID == 20 and item.group.ID != 303:  # Implants not Boosters
+            if item.category.ID == 20 and item.group.ID != 303:  # Implants not Boosters
                 implant_remove_list = set()
                 implant_remove_list.add("Low-Grade ")
                 implant_remove_list.add("Low-grade ")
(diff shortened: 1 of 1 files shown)


blitzmann:
Awesome work, and well documented. :) Will probably be the weekend before I get to testing this and other PRs, but on the surface look great.

blitzmann:
twas a little longer than a weekend <_<

Issue #1695, opened 2018-08-15
Dragging module onto fitting window over a mismatched slot type erases module


feyyd:
While working on issue #1690 I discovered this issue.  Have a patch written for it but wanted to open the issue here so we have reference number for the commit.

Issue #1690, opened 2018-07-21
Moving module from market does not append correctly


blitzmann:
Moving a module from the market to the fitting window (on a slot that it doesn't fit in) should be the exact same as double clicking it in the market browser. That is, it should append it to the next available, legal slot. However, seems to not be working correctly, looks like it might be doing some sort of shift in positions.

![pyfabug](https://user-images.githubusercontent.com/3904767/43040729-3dd2d818-8d19-11e8-9904-4912c909aa13.gif)

feyyd:
Doing a little research of my own and the code I've isolated the issue to:
module.py : 458

The fits() function does not take into account the slot we are dragging onto, so it is causing unexpected behavior.  

Dragging a mismatched slot will remove whatever you drag onto regardless of whether it is the same type or not.  It will also remove the module (int the mismatched slot we drag over) even if the type of slot we are trying to fill is full.  If we drag over the wrong slot (ie mid onto a high slot), it shouldn't remove the equipped item, but it does currently.

There are a few spots I believe this could be fixed but I am new to the project so I thought I would just do a writeup here before making any changes.  I have a few proposals:
1) Add another check to changeModule in fit.py so that it checks where it is being dragged to, and if it is improper slot, do nothing and return None.  Then fittingView will call the append function on changeModule when None is returned.
2) Change addModule in fittingView.py , do the check before we call change, and only call change if we are modifying the correct slot, otherwise we just append.
3) Overload fits() function in fit.py with another parameter that is the slot we are dragging onto, so that it will return false if the types mismatch.

Getting tired so I'm going to call it, but I hope this information helps, I'll try to write a fix for this but I'd prefer some input on the direction to go.  I noticed there is a __calculateSlot function that could be used to compare the dragged onto slot and the module that is being dragged's slot, so the implementation shouldn't be too hard.  Cheers.

PR #1696, opened 2018-08-15
Issue #1695 Prevents modules from being destroyed on change if module type mismatch

feyyd:
While working on #1690 I discovered that dragging a module on top of a mismatched item slot (ie High onto a Mid) would result in the item that was dragged onto being deleted no regardless.


feyyd:
working to resolve these issues, first submission here and I wan't very thorough.

feyyd:
closed and reissued on #1697, apologies if this is improper way of doing things.

PR #1697, opened 2018-08-16
Bug/1695, Bug/1690

feyyd:
#1690 and #1695.  Retrying pull request after running tox tests and fixing other pull issues.


feyyd:
I commited the 1690 fix along with this pull request, but not sure if it needs to re-run the automated checks.   Let me know if something else needs to be done on my end.

blitzmann:
excellent work @feyyd! I've tested both and the code looks solid, thanks!

Issue #1706, opened 2018-08-22
Price columns not loaded until forcing fittingView refresh.


feyyd:
@blitzmann edit:
To clarify, this happens when there is not pricing information int the database at all. The prices are fetched from the web, but are never displayed until the next GUI refresh

---

## Bug Report
The first time a module is dragged onto a fit, imported onto a fit, loaded from fit, the price columns in fittingView are not updated.  This problem fixes itself on next refresh, ie swapping module, adding module (with the exception of the module dragged, other prices update), activating/deactivating/heating a module, deleting a module.  Seems like anything that refreshes, but these were the cases I tested that caused the price columns to update.

One can see the price statView update upon retrieval of item price, however, the column price does not get refresh after prices are retrieved until next fittingView refresh.

### Expected behavior:
Upon loading a fit from db/clipboard, price columns should be updated after prices are retrieved.
Upon loading single module, price column should be updated after price is retrieved.

### Actual behavior:
Price columns remain empty until next fittingView refresh.

### Detailed steps to reproduce:
Load any fit, price will update but all columns remain empty until next fittingView update.
Load any module (that hasn't been loaded prior) and price column will remain empty.

### Release or development git branch? Please note the release version or commit hash:
Release 2.4.0

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
Windows 10

### Other relevant information:
I believe what happens is when price data isn't available (we first load the fit, first load an individual module) we Thread out the market price retrieval.  The price updates, but the price callback probably needs to call a refresh on fittingView once prices are retrieved.  

### Workaround
Force a fittingView refresh by toggling module state,  swapping a module, deleting module

feyyd:
Haven't looked into the code yet, but we should be able to add a fittingView refresh call to the statsView->priceViewFull and statsView->priceViewMinimal.

feyyd:
Been looking over this issue and it looks like there are a couple of ways I can fix this.  

Method one:  Call the FIT_CHANGED global event in the price callback.  This is very easy to do, will require minimum changes to the code, but I'm not sure if using FIT_CHANGED global event will have side effects (logging, marking fits as dirty for saves, etc).

Method two:  Bind a new event to fittingView PRICE_CHANGED.  Call this new event in price callback, PRICE_CHANGED and call the refresh.

[...]

PR #1709, opened 2018-08-23
Bug/1706

feyyd:
#1706 Adds a new global event, PRICE_CHANGED to use in the Price callbacks so that fittingView knows the price columns need to be updated.   This was used in lieu of FIT_CHANGED so that log messages would be more accurate.  

Changed the call signature of fitChanged to take in extra value which is the log label.  The fitChanged and priceChanged functions were identical so just went for code reuse, I don't think this will have any side effects as its an optional parameter.

I didn't modify port.py, but my tox tests were failing (i branched off upstream master) so I resolved this.

### DIFF
--- a/gui/builtinStatsViews/priceViewFull.py
+++ b/gui/builtinStatsViews/priceViewFull.py
@@ -24,6 +24,8 @@ from gui.bitmap_loader import BitmapLoader
 from gui.utils.numberFormatter import formatAmount
 from service.price import Price
 from service.settings import PriceMenuSettings
+import gui.mainFrame
+import gui.globalEvents as GE
 
 
 class PriceViewFull(StatsView):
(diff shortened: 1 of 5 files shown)


feyyd:
Sounds good.  As it stands the workaround is so easy because ANYTHING refreshing fittingView resolves the issue.

feyyd:
Been looking over this, I didn't know these viewColumns existed else I would have started there.  It seems the cause of this issue is that delayedText for Price is never being run.  This is because we return `''` instead of `False` when we don't have a price.  I fixed this, and got delayedText function running, but it has other issues I'm working on resolving.

blitzmann:
> seems the cause of this issue is that delayedText for Price is never being run.

Yepp! Sorry, I had a fix in place and forgot to push and make PR, but yeah, essentially some refactoring a while ago caused this whole callback thing to break.

Will create a PR with my fix and we can discuss on there. As I said in the review, eventually I want to look at utilizing events in a smarter way for pricing updates instead of relying on callbacks in callbacks in callbacks >.<

feyyd:
Awesome, great work as usual, i'll delete this PR once the other gets merged.

blitzmann:
Closing in favor of #1715

PR #1724, opened 2018-08-31
Abyssal export

blitzmann:


blitzmann:
@DarkFenX fixed a small issue in the xml import and some other formatting thing, but all in all looking good! EFT cfg works (at least my simple test files work)

Issue #1720, opened 2018-08-27
Possible to activate multiple prop-mods


Kalaratiri:
## Bug Report
More than one propulsion module can be activated/does not automatically deactivate the first prop-mod on adding the second

### Expected behavior:
Fitting an mwd while an afterburner is fitted should turn the afterburner into the "off" state. 
It should not be possible to activate one type of prop-mod while another is already activated.
It should not be possible to activate two of the same prop-mod.

### Actual behavior:
Fitting a second propulsion module does not deactivate the one already fitted.
It is possible to have both an afterburner and an mwd active on the fit. https://i.imgur.com/E9HYM1i.png
It is possible to have two of the same propulsion modules active on the fit. https://i.imgur.com/7F7Eour.png

### Detailed steps to reproduce:
Fit any two prop-mods.

### Fits involved in EFT format (Edit > To Clipboard > EFT):
Any fit.

### Release or development git branch? Please note the release version or commit hash:
2.5.0b1

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
Windows 10

### Other relevant information:
Amusingly, the speed appears to be capped at 1,000,000m/s, although an 8 mwd scorpion can get considerably faster than this. https://i.imgur.com/1MchmhQ.png

blitzmann:
Merp. Yeah, this is a thing. A byproduct of the new Command pattern that enables undo/redo.

The reasoning behind it is due to the fact that, before, when we added a module, we were able to add it then run it through a `checkStates` function, which checks the validity of states. If we found an invalid states, we would turn it off (based on another module).

We can no longer do that with the Command pattern because we don't want to do a recalc within the command. The reasoning behind this is because it's not part of the command's scope - there may be situations where we need to do three or four commands, and we don't want to recalc after each one (usually), so the recalc happens at a higher level than the command itself.

I meant to look into this bug and find a better way of fixing states - one that doesn't require a recalculation. I don't think it would be difficult, just never got around to actually looking into it.

All in all, I think this is a minor bug in comparison to the usefulness of what we get with the command pattern (undo/redo). If it needs to live around for a while until it gets fixed, I'm okay with that. That being said, I will try to look into this issue again before the actual release.

😄

blitzmann:
This was sort of fixed with recent commits to the 2.5.0 release. There's still some side effects to the undo/redo support, but this particular issue is fixed :)

PR #1727, opened 2018-09-03
Undo/Redo (refactoring into command pattern)

blitzmann:
This PR introduces a refactoring of the fit service into "Commands", specifically to facilitate undo/redo functionality.

Basics were done first - modules, states, projections, etc. First-order fitting things. Second order fitting modifications (abyssal modifications, fighter abilities, etc) will come later, but this is a good first iteration.

Gonna keep this open for a couple of days to facilitate comments - the bugs that have so far been reported have been fixed, sans #1720. But this is a huge change, not only in number of file affected, but also an evolution in the basic design principles that we've used so far in pyfa. I'll be going through the diff line by line to see if I notice anything funky that I may have forgotten about, and address (or at the very least comment on)


config.py
```diff
@@ -229,19 +229,6 @@ def defLogging():
...
-        try:
-            sys.stdout = LoggerWriter(pyfalog.warning)
-        except:
-            pyfalog.critical("Cannot redirect.  Continuing without writing stdout to log.")
-
-        # Output all stderr (stacktrace) messages as critical
-        try:
-            sys.stderr = LoggerWriter(pyfalog.critical)
-        except:
-            pyfalog.critical("Cannot redirect.  Continuing without writing stderr to log.")
```
blitzmann: I don't quite remember why I removed this bit; had to do with needing this stuff to not redirect. I think it had something to do with the deprecated decorator I'm using to flag things as old now. Dunno if we actually need it ¯\_(ツ)_/¯

eos/saveddata/fighter.py
```diff
@@ -53,6 +53,20 @@ def __init__(self, item):
...
+                ability.active = True
+                standardAttackActive = True
+            else:
+                # Activate all other abilities (Neut, Web, etc) except propmods if no standard attack is active
+                if ability.effect.isImplemented and \
+                                standardAttackActive is False and \
+                                ability.effect.handlerName != 'fighterabilitymicrowarpdrive' and \
+                                ability.effect.handlerName != 'fighterabilityevasivemaneuvers':
+                    ability.active = True
+
```
blitzmann: Moved this bit from the add fighter function to the actual fighter initialization

gui/builtinShipBrowser/fitItem.py
```diff
@@ -17,10 +17,29 @@
 from gui.bitmap_loader import BitmapLoader
 from gui.builtinShipBrowser.pfBitmapFrame import PFBitmapFrame
 from service.fit import Fit
-
+import gui.fitCommands as cmd
 pyfalog = Logger(__name__)
 
 
+class FitRenameCommand(wx.Command):
```
blitzmann: Need to move this, this was the very first test case of the command stuff, and I just stuck it here. >_>

blitzmann:
Merging this in - further testing hasn't provided any additional bugs to check out, so going all in :D

PR #1733, opened 2018-09-08
Implemented copying the currently open fit stats to the clipboard.

m-sasha:
Added a menu item in the "Edit" menu, which copies a compact form of the currently selected fit stats to the clipboard.

### DIFF
--- a/eos/saveddata/damagePattern.py
+++ b/eos/saveddata/damagePattern.py
@@ -74,6 +74,15 @@ class DamagePattern(object):
     }
 
     @classmethod
+    def oneType(cls, damageType, amount=100):
+        pattern = DamagePattern()
+        pattern.update(amount if damageType == "em" else 0,
+                       amount if damageType == "thermal" else 0,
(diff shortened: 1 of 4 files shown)


IndictionEve:
would love to see this in pyfa for base stats.
can be enhanced step by step if needed.

Gochim:
Hi

Is this pull request worth looking into to solve it's issues and make it into master? Or there are some serious problems that I'm not seeing with that PR so that it's not worth bothering?

I there is nothing serious with it - I could pretty it up and finish...

DarkFenX:
@Gochim this pull request was submitted during my absence so I haven't really checked it. No idea why @blitzmann ended up not accepting it. But i can see following things which have to be done before accepting it:

1) Stat format has to be discussed and agreed on
2) Export could be moved to regular ctrl-c menu (where EFT, multibuy and other exports are)

DarkFenX:
@Gochim please create a ticket for discussion of format. I think it's more appropriate than doing it in here.

PR #1735, opened 2018-09-11
The text drawn in chrome_tabs is now crisp on Retina screens

m-sasha:
Before:
screen shot 2018-09-11 at 14 45 30

After:
screen shot 2018-09-11 at 14 44 38

(You might need to view these images on a Retina screen to see the difference).

### DIFF
--- a/gui/chrome_tabs.py
+++ b/gui/chrome_tabs.py
@@ -552,24 +552,6 @@ class _TabRenderer:
                     bmp,
                     self.left_width + self.padding - bmp.GetWidth() / 2,
                     (height - bmp.GetHeight()) / 2)
-            text_start = self.left_width + self.padding + bmp.GetWidth() / 2
-        else:
-            text_start = self.left_width
-
-        mdc.SetFont(self.font)
-
-        maxsize = self.tab_width \
-            - text_start \
-            - self.right_width \
-            - self.padding * 4
-        color = self.selected_color if self.selected else self.inactive_color
-
-        mdc.SetTextForeground(color_utils.GetSuitable(color, 1))
-
-        # draw text (with no ellipses)
-        text = draw.GetPartialText(mdc, self.text, maxsize, "")
-        tx, ty = mdc.GetTextExtent(text)
-        mdc.DrawText(text, text_start + self.padding, height / 2 - ty / 2)
 
         # draw close button
         if self.closeable:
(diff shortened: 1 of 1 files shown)


m-sasha:
@blitzmann Can you point me to them? ShipItem and FitItem look fine on my Mac. All the icons are blurry everywhere, but fixing that would require high-res icons, and I don't know where to find them.

blitzmann:
> @blitzmann Can you point me to them? ShipItem and FitItem look fine on my Mac. All the icons are blurry everywhere, but fixing that would require high-res icons, and I don't know where to find them.

Then they may be fine, I just know that they are custom drawn and all our custom drawn widgets have scaling issues

m-sasha:
By the way, is there a mapping between pyfa's image names and the ones in CCPs image dump at https://developers.eveonline.com/resource/resources ?

blitzmann:
As for the icons - yeah, most of ours are 16x16. To fix the icons is a bit of an involved process, and involves us shipping two sets - the normal set, and the 2x set (basically icons that are twice as big, but shown as half their size - so a 16x16icon would be 32x32 on the filesystem, but still displayed as a 16x16).

wxPython supposedly has support for loading these kind of icons using the `@2x` convention (regular icons is called `icon.png` whereas the hide one is `icon.png@2x` or something like that). Or maybe that's just the convention and it still needs to be properly supported in the applications code, dunno.

It would probably require a bit of tweaking with our bitmap loader (possibly a rewrite of that) to find and provide the correct image. There's scarce resources I've been able to find on this, and the fact that I don't have a UHD display hinders the amount development work I can devote to it.

If you're interested in researching this, googling various terms such as wxPython, wxWidgets, UHD, HiDPI, Retina (of course), etc should get you started. I've also heard that it might not have decent support in wxWidgets 3.0, which is what wxPython is based off of (for now), so it may be futile?

blitzmann:
> By the way, is there a mapping between pyfa's image names and the ones in CCPs image dump at https://developers.eveonline.com/resource/resources ?

The icons in pyfa are named after the icon ID. You might be able to find some interesting info in the `icon.yaml` file in the SDE - although I don't know off the top of my head how that translates to the actual files (the file structure is totally fucked)

if you want to play around with it, I would start with a very simple demo / test bed, such as the icon for stasis webifier (or maybe even something outside pyfa completely, a very small stripped down demo). if we can get it working with one icon really well, then we can look into incorporating it into the script that gets all the icons, and go from there :)

m-sasha:
It appears doable. wx.Window.GetContentScaleFactor() to get the scale factor, then load the appropriate wx.Image, convert it to wx.Bitmap and SetSize(w/scale, h/scale) on it.
Before:
scorp_1x

After:
scorp_2x

PR #1737, opened 2018-09-11
BitmapLoader.getBitmap will now load a @2x (or whatever the main window's content scale is) bitmap, if one is available.

m-sasha:
This should solve the problem for all images loaded with getBitmap(). You can test it by grabbing CCP's high-res images from https://developers.eveonline.com/resource/resources and adding a 2x version of whatever icon/image you find.
Before:
scorp_1x
After:
scorp_2x

Code that uses getImage(), unfortunately, must either be converted to use getBitmap() (the preferable way), or carry the scaling factor up until it actually draws the image, and downscale it there (with this patch, getImage() will return a prescaled-down-to-regular-size version of the 2x image, so code using it shouldn't be affected).

### DIFF
--- a/gui/bitmap_loader.py
+++ b/gui/bitmap_loader.py
@@ -27,6 +27,8 @@ import wx
 
 import config
 
+import gui.mainFrame as mainFrame
+
 from logbook import Logger
 logging = Logger(__name__)
 
@@ -55,9 +57,7 @@ class BitmapLoader(object):
     @classmethod
     def getBitmap(cls, name, location):
         if cls.dont_use_cached_bitmaps:
-            img = cls.getImage(name, location)
-            if img is not None:
-                return img.ConvertToBitmap()
+            return cls.loadBitmap(name, location)
 
         path = "%s%s" % (name, location)
 
@@ -65,11 +65,7 @@ class BitmapLoader(object):
             cls.cached_bitmaps.popitem(False)
 
         if path not in cls.cached_bitmaps:
-            img = cls.getImage(name, location)
-            if img is not None:
-                bmp = img.ConvertToBitmap()
-            else:
-                bmp = None
+            bmp = cls.loadBitmap(name, location)
             cls.cached_bitmaps[path] = bmp
         else:
             bmp = cls.cached_bitmaps[path]
@@ -78,8 +74,29 @@ class BitmapLoader(object):
 
     @classmethod
     def getImage(cls, name, location):
-        filename = "{0}.png".format(name)
+        return cls.getBitmap(name, location).ConvertToImage()
 
+    @classmethod
+    def loadBitmap(cls, name, location):
+        filename = "{0}.png".format(name)
+        scale = int(mainFrame.MainFrame.getInstance().GetContentScaleFactor())
+        if scale == 1:
+            img = cls.loadImage(filename, location)
+        else:
+            filenameScaled = "{0}@{1}x.png".format(name, scale)
+            img = cls.loadImage(filenameScaled, location)
+            if img is None:
+                img = cls.loadImage(filename, location)
+                scale = 1
+        if img is None:
+            return None
+        bmp: wx.Bitmap = img.ConvertToBitmap()
+        if scale > 1:
+            bmp.SetSize((int(bmp.GetWidth()/scale), int(bmp.GetHeight()/scale)))
+        return bmp
+
+    @classmethod
+    def loadImage(cls, filename, location):
         if cls.archive:
             path = os.path.join(location, filename)
             if os.sep != "/" and os.sep in path:


blitzmann:
Holy hell, this is huge for uhd screens! Will test as soon as I can and evaluate the code, then fix up the icon getter script to process them and add to the repo. 

Great work! I honestly didn't think it would be this simple =D

blitzmann:
I'm going to merge this into an official branch on the pyfa repo - I've done a few tweaks (this PR has a broken import) as well as re-ran the icons script to populate it with the new, scaled images

Just tried on Windows - doesn't work. `GetContentScaleFactor()` always returns 1 for me after setting my scaling to 200%, though I suspect this has more to do with the underlying support for windows than it does with the code. I'm going to have a few OSX folks test this tho and hopefully we can get this moved into `master`./

Please try the `hd-images` branch with all the scaled images saved and let me know if it all looks kosher! I know there's going to be a few icons that are GUI-specific (right stats panel, some third-party icons, etc) that would need to be re-worked, but I want to know that the bulk of the module icons and ship renders look good.

@DarkFenX if you could also test `hd-images` on linux with your snazzy new UHD screen, that would be amazing.

minlexx:
What about fractional scaling? Qt's `QT_SCALE_FACTOR` supports non-integer values ( http://doc.qt.io/qt-5/highdpi.html ), and according to https://wiki.archlinux.org/index.php/HiDPI#Fractional_Scaling , Gtk apps can do it, too. See also https://developer.gnome.org/gtk3/stable/gtk-x11.html `GDK_SCALE`, `GDL_DPI_SCALE` . Qt docs also mentions in Windows section DPI awareness that must be set by an application `An application on Windows can assume one of the following levels of "DPI Awareness": https://msdn.microsoft.com/ru-ru/C9488338-D863-45DF-B5CB-7ED9B869A5E2`. In our case GUI framework should do that.. In the end true fix ideally would go into wxWidgets, and while wx is not fixed, it may need some workarounds. Like in Linux, manually read env variables (`GDK_SCALE` / `QT_SCALE_FACTOR`), and in WIndows, I don't know yet where to get scaling info from :cry: .

P.S. I also have 4k display and I probably can test in Linux KDE Plasma desktop.

m-sasha:
If this looks good on those systems, we can look for the image with the closest (rounding up, I guess) available scaling and use that. So with 1.25x scaling, it would use the 2x image and scale it down.

minlexx:
> Add a user setting to use high-res images.

I like this option

> look for the image with the closest (rounding up

Sounds good, too

m-sasha:
> Please try the `hd-images` branch with all the scaled images saved and let me know if it all looks kosher! I know there's going to be a few icons that are GUI-specific (right stats panel, some third-party icons, etc) that would need to be re-worked, but I want to know that the bulk of the module icons and ship renders look good.

The ship and module renders look good. Everything else is still blurry, obviously.

m-sasha:
By the way, I would recommend using larger versions of icons. 16x16 is too small for rich icons and is really only OK for simple geometrical ones; At 24x24 they would be much easier to read.

blitzmann:
@minlexx does the image scaling work correctly for you on Linux?

> By the way, I would recommend using larger versions of icons. 16x16 is too small for rich icons and is really only OK for simple geometrical ones; At 24x24 they would be much easier to read.

That would be a much larger change, and more of a design discussion. Increasing the icon size would lead to larger list views and the like wherever the icons are used (module listing, for example. I can play around with them and see how it looks though, as I don't think I've ever tried it

minlexx:
@blitzmann It does not look to me like the image scaling is woking. Judge by yourself, hd-images branch screens (beware 4k PNG images):
* [with GDK_SCALE=1](https://eve-wh.space/static/img/pyfa%20hd-images%20scale%201.png)
* [with GDK_SCALE=2](https://eve-wh.space/static/img/pyfa%20hd-images%20scale%202.png)

what about second thought in my [comment](https://github.com/pyfa-org/Pyfa/pull/1737#issuecomment-420562591)  about working around broken GUI toolkit and manually reading `os.environ['GDK_SCALE' / 'QT_SCALE_FACTOR''] `?

blitzmann:
> what about second thought in my comment about working around broken GUI toolkit and manually reading os.environ['GDK_SCALE' / 'QT_SCALE_FACTOR'']?

Sorry, didn't fully process that at the time.

Is it simply that on Linux `GetContentScaleFactor` just doesn't return the correct value (much like on Windows where it seems to always be `1`)?

If you hardcode the scale, does it look normal or still pixelated? If it looks normal, then I'm okay with adding these environmental checks and maybe an option in preferences to force loading of HD images.

I'm going to need to put together a small test application that I can take to robin and say "fix this" or at least get confirmation that it's not a thing with the wxWidgets version wxPython is running off of. >_>

I'll test a windows fix with that manifest tweak and see if that works for at least built pyfa versions.

minlexx:
@blitzmann I'm here! You want me to write this application? In python or C++? 😉 

What I think: the whole HIDPI problem is not simple, especially in Linux case, where different desktop envs, GUI toolkits do not play nice with each other. If this PR solves problem at least at Mac, you can accept it. Or even better find a solution for Windows, I guess there are more users on Windows. And Linux users are used to suffer anyaway 😆

blitzmann:
> @blitzmann I'm here! You want me to write this application? In python or C++? 😉

lol no, I guess I was more asking about if you actually tested any of the fixes you suggested and if they are working well on Linux. For example, since wx isn't picking up the scale automatically, if you simply hardcode it to `2` does it get the correct images and do those images look correct in the application?

The plan is to get this merged even if full support isn't there. I'm just needing a bit more confirmation on the Linux side of things since I can't test easily (I only have a 1440 monitor with a VirtualBox Mint VM that isn't properly set up for my monitors / GPU, not ideal for testing with scaled things 😛 ). If it seems to work in linux fine, it's just a fact of not being able to get the scale, then I can easily work around that using environment variables or even a literal checkbox that says "Use HD images" in the preferences.

minlexx:
https://cgit.kde.org/kscreen.git/tree/kcm/src/scalingconfig.cpp#n84 

It turns out that KDE's screen config is using `QT_SCREEN_SCALE_FACTORS` envvar (not `QT_SCALE_FACTOR` that I was thinking about), as comment in code says. So many different variables (several for Qt, several for Gtk) so maybe user-configured checkbox "Use HD images" is the easiest way?

> if you actually tested any of the fixes you suggested

I haven't tried yet, I'll try later today. I can only test in KDE Plasma now, I don't have Gnome to test

blitzmann:
> I haven't tried yet, I'll try later today. I can only test in KDE Plasma now, I don't have Gnome to test

That's totally fine, and would be much appreciated. I believe @DarkFenX uses Gnome, so I'll ask him to test it on that. Thanks!

blitzmann:
@DarkFenX reports that setting a hardcoded value of `2` for the scaling factor doesn't seem to work in GTK:

image

Even thought the bitmap seems to scale down correctly to 16x16 for the icons (and I assume 32x32 for the renders).

Very interested in seeing the test results on the KDE side :)

minlexx:
I don't understand how it was tested, harcoded 2 was where? Was `GDK_SCALE=2` exported?

blitzmann:
> I don't understand how it was tested, harcoded 2 was where? Was GDK_SCALE=2 exported?

`2` was hardcoded as the scale in the loader (this line https://github.com/pyfa-org/Pyfa/blob/7f02a078fd64d363e681a21c2479972568a4b37a/gui/bitmap_loader.py#L49) - this results in the image above. 

Exporting `GDK_SCALE=2` seemed to have no effect on images, but drastic effects on the text. Whether or not this is actually indented to happen, I have no idea. I don't know enough about this shit >.<

image

¯\_(ツ)_/¯

minlexx:
hardcoded `scaling_factor=2`
* [with no GDK scaling, GDK_SCALE=1](https://eve-wh.space/static/img/pyfa%20hd-images%20scale%202-1.png)
Image is HD (x2 size) but is cropped
* [with GDK scaling, GDK_SCALE=2](https://eve-wh.space/static/img/pyfa%20hd-images%20scale%202-2.png)
Image is HD, but scaled once again (x4 size and little blurry)

blitzmann:
hrm, kk, thanks for checking for me. Still some work to be done to support other platforms, and it seems it wont be as easy as a checkbox to enable HD images.

blitzmann:
wxwidgets 3.1 might have better support, but wxpython uses the 3.0 branch. Unsure when, if ever, 3.1 will be supported :(

blitzmann:
I've slated this for a 2.5 release, nice work all around and thanks to everyone for helping test it on various systems 👍 I'll close this PR once the release is out

m-sasha:
Unfortunately, it appears that
```
bmp.SetSize((bmp.GetWidth() // scale, bmp.GetHeight() // scale))
```
fails with wxPython 4.1.0 with:
```
pyfa v2.24.1
EVE Data Version: 1777281 (2020-07-28 11:14:33)

OS version: Darwin-19.5.0-x86_64-i386-64bit
Python version: 3.7.7 (default, Mar 10 2020, 15:43:33) 
[Clang 11.0.0 (clang-1100.0.33.17)]
wxPython version: 4.1.0 (wxWidgets 3.1.4)
SQLAlchemy version: 1.3.18
Logbook version: 1.5.3
Requests version: 2.24.0
Dateutil version: 2.8.1

####################

  File "pyfa.py", line 143, in <module>
    mf = MainFrame(options.title)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/mainFrame.py", line 172, in __init__
    self.additionsPane = AdditionsPane(self.fitting_additions_split, self)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/additionsPane.py", line 64, in __init__
    self.drone = DroneView(self.notebook)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/builtinAdditionPanes/droneView.py", line 73, in __init__
    Display.__init__(self, parent, style=wx.BORDER_NONE)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/display.py", line 44, in __init__
    self.insertColumnBySpec(i, colName)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/display.py", line 123, in insertColumnBySpec
    col = ViewColumn.getColumn(colSpec)(self, None)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/builtinViewColumns/maxRange.py", line 45, in __init__
    self.imageId = fittingView.imageList.GetImageIndex(iconFile, "icons")
  File "/Users/sasha/Development/Projects/PyfaAT/gui/cachingImageList.py", line 34, in GetImageIndex
    bitmap = BitmapLoader.getBitmap(*loaderArgs)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/bitmap_loader.py", line 74, in getBitmap
    bmp = cls.loadBitmap(name, location)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/bitmap_loader.py", line 108, in loadBitmap
    bmp.SetSize((bmp.GetWidth() // scale, bmp.GetHeight() // scale))
wx._core.wxAssertionError: C++ assertion "GetHeight() == h" failed at /Users/robind/projects/bb2/dist-osx-py37/build/ext/wxWidgets/src/osx/core/bitmap.cpp(1380) in SetHeight(): Changing the bitmap height is not supported
```

m-sasha:
I tried to replace `Bitmap.SetSize` with `Bitmap.CreateScaled`, but then it fails with:
```
  File "pyfa.py", line 143, in <module>
    mf = MainFrame(options.title)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/mainFrame.py", line 172, in __init__
    self.additionsPane = AdditionsPane(self.fitting_additions_split, self)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/additionsPane.py", line 64, in __init__
    self.drone = DroneView(self.notebook)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/builtinAdditionPanes/droneView.py", line 73, in __init__
    Display.__init__(self, parent, style=wx.BORDER_NONE)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/display.py", line 44, in __init__
    self.insertColumnBySpec(i, colName)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/display.py", line 123, in insertColumnBySpec
    col = ViewColumn.getColumn(colSpec)(self, None)
  File "/Users/sasha/Development/Projects/PyfaAT/gui/builtinViewColumns/maxRange.py", line 45, in __init__
    self.imageId = fittingView.imageList.GetImageIndex(iconFile, "icons")
  File "/Users/sasha/Development/Projects/PyfaAT/gui/cachingImageList.py", line 37, in GetImageIndex
    id_ = self.map[loaderArgs] = wx.ImageList.Add(self, bitmap)
wx._core.wxAssertionError: C++ assertion "bitmapSize == m_size" failed at /Users/robind/projects/bb2/dist-osx-py37/build/ext/wxWidgets/src/generic/imaglist.cpp(85) in Add(): All bitmaps in wxImageList must have the same size
```

PR #1760, opened 2018-10-08
Fix cargo items price increase on refit

### LINKED ISSUE #1759: Changing hybrid charges increases stored price of Void S / Null S
## Bug Report
Some prices (in particular, Null S, Void S) were insanely high. Like 5e37 high, which I investigated. Deleting prices forces a refresh, and gathers the right price (about 50ISK). However, whenever I change the charge of one of my blasters, the price of Void and Null in my cargo goes up by a factor of about 1000. Actually, it's exactly 1000, until floating point errors starts showing up. 

### Expected behavior:
Changing charges in a blaster shouldn't change prices.

### Actual behavior:
Changing charges changes prices.

### Detailed steps to reproduce:
Using the attached fit, keep changing the charge in any of the neutron blasters, either with right-click or double-clicking on a new charge. Each time, the cost of Void S and Null S increases.

I've tested on other ships, and it doesn't seem to depend on ship. I've also tested with other ammo, and only Void S and Null S prices seem affected. I've even checked missiles, and while I couldn't spot any missiles increasing in price, changing the missile in a RLML I also increased the cost of Void S/Null S.

I've even uploaded a video I shot at http://strudel-hound.com/pyfa.mkv but note that my screen capture apparently doesn't capture the right-click menus, so the first few charge changes seem to show me clicking on nothing and then the charge changing. You can see this charge change by looking at the actual fitting table though.

### Fits involved in EFT format (Edit > To Clipboard > EFT):
[Enyo, Hunter]

EFFA Compact Assault Damage Control
Magnetic Field Stabilizer II
Tracking Enhancer II
Adaptive Nano Plating II

5MN Y-T8 Compact Microwarpdrive
Initiated Compact Warp Scrambler
Fleeting Compact Stasis Webifier

Light Neutron Blaster II, Uranium Charge S
Light Neutron Blaster II, Guristas Uranium Charge S
Light Neutron Blaster II, Void S
Light Neutron Blaster II, Antimatter Charge S
Small Ghoul Compact Energy Nosferatu

Small Anti-Explosive Pump I
Small Processor Overclocking Unit I

Null S x1000
Void S x1000
Thorium Charge S x1

### Release or development git branch? Please note the release version or commit hash:

I'm running master branch off git, 9ec54c90046231ebb8006e3e9884802c5a53f027

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):

Gentoo

### Other relevant information:

I'm comfortably with Python, I can debug stuff if someone gives me a hint as to where to look, but I don't know where to get started on this.

WPettersson:
The price attribute of an item is linked to its price in the database,
through SQLAlchemy and the ORM map. Updating a price attribute therefore
updates the database. This patch changes the calculation of the cost of
amounts of items, so that the price in the database no longer occurs.

Fixes #1759

### DIFF
--- a/gui/builtinViewColumns/price.py
+++ b/gui/builtinViewColumns/price.py
@@ -51,7 +51,9 @@ class Price(ViewColumn):
             return False
 
         if isinstance(stuff, Drone) or isinstance(stuff, Cargo):
-            price.price *= stuff.amount
+            # Don't set price.price, or it updates the database #1759
+            return formatAmount(price.price * stuff.amount, 3, 3, 9,
+                                currency=True)
 
         return formatAmount(price.price, 3, 3, 9, currency=True)
 


WPettersson:
Closed as this is fixed in release/v2.5.0 anyway.

Issue #1769, opened 2018-10-13
Error when attempting to browse Subsystems on new Tengu fit


brianhultin:
## Bug Report
Error when attempting to browse subsystem on new Tengu fit

### Expected behavior:
Drop down will show available subsystem groups

### Actual behavior:
Error message, drop down does not work, drop down button disappears. Must quit and restart to get it back.

### Detailed steps to reproduce:
Create new Tengu fit
Go to Market tab, click drop down button on Subsystems

### Fits involved in EFT format (Edit > To Clipboard > EFT):
[Tengu, Cloaky PVP]

[Empty Rig slot]
[Empty Rig slot]
[Empty Rig slot]

[Empty Subsystem slot]
[Empty Subsystem slot]
[Empty Subsystem slot]
[Empty Subsystem slot]

### Release or development git branch? Please note the release version or commit hash:
Not sure of branch, version 2.5.0

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
Mac OSX 10.13.6

### Other relevant information:
pyfa vv2.5.0
EVE Data Version: 1399820 (2018-10-09 15:27:02)

OS version: Darwin-17.7.0-x86_64-i386-64bit
Python version: 3.6.3 (default, Oct 22 2017, 22:24:50) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.72)]
wxPython version: 4.0.0b2 (wxWidgets 3.0.4)
SQLAlchemy version: 1.1.10
Logbook version: 1.0.0
Requests version: 2.18.4
Dateutil version: 2.6.1

####################

Traceback (most recent call last):
  File "/Users/ryan/pyfa/gui/builtinMarketBrowser/marketTree.py", line 61, in expandLookup
  File "/Users/ryan/pyfa/gui/builtinMarketBrowser/marketTree.py", line 43, in addImage
  File "/Users/ryan/pyfa/gui/cachingImageList.py", line 37, in GetImageIndex
[...]

PR #1772, opened 2018-10-14
Try loading smaller variants if scaled icons are missing

tron1point0:
References the [discussion on #1769](https://github.com/pyfa-org/Pyfa/issues/1769#issuecomment-429658399).


tron1point0:
AppVeyor is going to fail every time because there's no `requirements_test.txt`:

```
Could not open requirements file: [Errno 2] No such file or directory: 'requirements_test.txt'
241
```

Don't know what should go in there (if anything), so I can't fix it.

Issue #1819, opened 2018-12-12
Spool-up options


DarkFenX:
At present moment, pyfa has full support for spoolup options on the back-end side (for both damage and remote reps). DPS and RR fetchers now have two additional arguments: spoolType and spoolAmount.

SpoolType.SCALE: takes values [0, 1] as amount, with 0 being 0% spoolup and 1 being 100% spoolup. For values in-between, it rounds down to closest module cycle.
SpoolType.TIME: spool amount should be float which defines time in seconds since spoolup modules started firing.
SpoolType.CYCLES: spool amount is amount of cycles modules went through since they started firing.
None: take per-module spoolup settings, stored in module.spoolType and module.spoolAmount

The value they return shows current volley/dps/rr per second for the moment specified by these spoolup arguments.

All fields which need to rely on these options are marked with `# TODO: fetch spoolup option` comment:
- firepower stats panel
- remote reps stats panel (full and minimal)
- fitting panel's misc column (spoolup RRs)
- spoolup RR effect (this will be the trickiest one to support, i suppose)
- EFS stats export

From UI side, it's up to you how to implement these options. I see few possible variants:
- per-module options only (with default being 0% or 100%)
- global options only
- global options with per-module override

DarkFenX:
Another thing to remember: if we implement such options, it would make sense to also update spoolup time for guns/RRs we show in misc column.

DarkFenX:
I completed backend support for these options. Look up for `# TODO: fetch spoolup option` comment, code below it assumes option value = 1 for now.

If you change spoolType and spoolAmount attributes of module, these will set overrides and will be used instead of default values you pass from the above.

PR #1836, opened 2019-01-11
Spoolup

blitzmann:
Code to support modifying spoolup times per module. See #1819

image

Waiting for response from @DarkFenX to make sure it looks  functions proper for linux


DarkFenX:
Seems to be ready for merge after my changes, but context menu needs to be checked on windows.

Issue #1784, opened 2018-10-27
Feature Request - Make used Additions visible


Dawnkeeper:
Right now you have to click on each tab in the Additions section to see if it is used.

It would be nice to see on one glance whether a ship has notes, uses Implants and so on.

This could be done by writing the labels in bold or adding a symbol as soon as this category is used.

blitzmann:
Hi @Dawnkeeper!

This has been mentioned before #1717. I haven't had time to really dig into it as it's kind of a low priority thing, but it seems like at least a few folks might find this helpful. The other issue went in a little bit of a different direction, requesting things like counts and stuff that might be very tricky to do. However, I do like the idea of having the tab bolded to show that there's something there, and this would be a much easier thing to do.

I'm going to try to get something like this implemented by the next release, but again, this would be a low priority thing.

Also tagging this as `help wanted` in case someone drops by looking for something a bit "easier" to do. A bit of info on how this might work:  whenever a fit is changed, an event happens which is captured by all the various addition panes. We can, as part of our event processing, determine if we should flag this panel as having items. We would, with some mechanism, tell the tab to turn bold - this would be the most tricky part, as I don't believe we currently have a way of communicating with the tab or having the tab change font characteristics.

petosorus:
Hi Dawnkeeper, hi Blitz

I'm interested at taking at look at it during today, I'll update if I feel good with it :)

blitzmann:
Awesome, thanks @petosorus!

petosorus:
You were right, I don't see a way to get the tab at the moment. I'm going to fiddle with it for a bit then!

DarkFenX:
Extended version of this was implemented in 2.11.1

PR #1842, opened 2019-01-25
#1784

petosorus:
For issue #1784.
NotesView doesn't work like the others because it does not implement Display.
The notes tab also does not update immediately as the others. It does update at least when a fit is closed, so the idea of the development is still there: opening a fit, seeing there are notes.


DarkFenX:
Without knowing about this PR (didn't check those which were submitted during my absence), I've implemented extended version of this PR in 2.11.1. It now shows amount of active / all items (according to prererences) in tab titles.

Now as feature is there already, I am closing it.

DarkFenX:
Oh and I also did not add support to Notes tab, although due to different reasons (no idea what to show there, just mere presence of text?)

petosorus:
I forgot about this one.
It looks like mere presence of text yeah.

DarkFenX:
Added it in 5f7d9aea893e8120c1a8ecb1052a1ce7fec0a02e - shows amount of text characters

PR #1844, opened 2019-01-29
Merge missing changes from #1755, efs version bump and minor linting.

MaruMaruOO:
#1755 was merged with the development branch in October but never made it to master branch.  
It looks like it's been forgotten so this patch adds the missing changes along with minor linting.  

I've also changed the SpoolOptions to always fully spool up regardless of settings.  
"damageMultiplierBonusMax" and "damageMultiplierBonusPerCycle" are exported already but they can only calculate the spool if the export is consistent and doesn't change with user settings.

### DIFF
--- a/service/port/efs.py
+++ b/service/port/efs.py
@@ -1,13 +1,8 @@
-import inspect
-import os
-import platform
-import re
-import sys
-import traceback
 import json
 import eos.db
 
 from math import log
+from numbers import Number
 from config import version as pyfaVersion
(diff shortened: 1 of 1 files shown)


blitzmann:
> It looks like it's been forgotten
Hrm, yeah we haven't been keeping up with the dev branch. Things have been kind of hectic. Honestly should probably merge that in soonish. Thanks for the heads up :)

PR #1847, opened 2019-02-11
Multibuy export options

### LINKED ISSUE #1840: Multibuy export options
1) export implants and boosters
2) export cargo

Also, it should not export abyssal modules (regardless of options)

DarkFenX:
Functionality to close #1840. Summary of changes:

- removed XML and DNA export options (just from UI)
- added options to export implants+boosters and cargo to multibuy export
- removed option to export loaded charges from global preferences, added corresponding option to EFT and multibuy export (should we have it for ESI too? For now, it exports loaded charges silently)
- reworked how options are stored. It means that previously stored EFT export options will be lost
- fixed pyfa putting export into clipboard even when user clicks "cancel"
- removed abyssal mods from multibuy export altogether (as they cannot be bought anyway)


DarkFenX:
This patch adds ESI option to export loaded charges: https://gist.github.com/DarkFenX/440b66a1243140d810caa99ee384034e

But it extends sizer with 1 checkbox to size of 3 checkboxes:

image

So i am not including it into this pull request until it's clear how to deal with it.

PR #1852, opened 2019-02-15
make the item list colour per slot if colorFitBySlot set

### LINKED ISSUE #1839: Color code the market items
Request from https://forums.eveonline.com/t/pyfa-the-python-fitting-assistant-v2-3-0-mutaplasmids/27156/134?u=sable_blitzmann

We should apply the same color coding found in the fitting window for slots to the market items. I feel like this has been tried before and didn't work out for some reason (possibly because slot is at the Module l;evel, whereas the market items are Items, thus don't technically have slots? Unsure)

Marking this as `help wanted` as I feel this is a good first project fdor someone to pick up if interested in contributing, otherwise will have to wait until I get some time to look into this.

tonycoz:
attempts to resolve #1839 

my first python code ;) let's see how your check tools treat it

### DIFF
--- a/config.py
+++ b/config.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import yaml
+import wx
 
 from logbook import CRITICAL, DEBUG, ERROR, FingersCrossedHandler, INFO, Logger, NestedSetup, NullHandler, \
     StreamHandler, TimedRotatingFileHandler, WARNING
@@ -47,6 +48,11 @@ LOGLEVEL_MAP = {
     "debug": DEBUG,
 }
 
+slotColourMap = {1: wx.Colour(250, 235, 204),  # yellow = low slots
+                 2: wx.Colour(188, 215, 241),  # blue   = mid slots
+                 3: wx.Colour(235, 204, 209),  # red    = high slots
+                 4: '',
+                 5: '' }
 
 def getClientSecret():
     return clientHash
--- a/eos/gamedata.py
+++ b/eos/gamedata.py
@@ -240,6 +240,7 @@ class Item(EqBase):
         self.__assistive = None
         self.__overrides = None
         self.__priceObj = None
+        self.__slot = None
 
     @property
     def attributes(self):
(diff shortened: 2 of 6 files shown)


blitzmann:
Hi @tonycoz!

This is a great first attempt at python, thank you for you contribution! I haven't yet had a chance to actually test it, however have taken a look at the code. I think I want to tweak it a bit by moving the slot into to the Item class and centralizing the definition of the slot colors somewhere so that we don't have two different places for it.

Excellent work though, this seems like it will be a really useful starting point! :)

blitzmann:
I see your changes, good work! I didn't consider putting that stuff in config.py... honestly not sure if that's the proper place. We have to import wx, which might be improper for such a low level config file. Might want to make a guiconfig.py or something. But dont worry about that right now, give me a chance to pull the code and give it a whirl.

Again, excellent work! Will let you know what, if anything, I find /tweak :)

tonycoz:
Yeah, I wasn't sure about adding wx to config.py, but it seemed the only appropriate (existing) place.

I also considered adding tests for item.slot, but when I originally wrote the code, the tests were failing in travis (before I rebased), and now the tests aren't being run in travis (it looks like a packaging check now.)

blitzmann:
Yeah, tests aren't really a thing. They were kinda a thing before, but just
took too much upkeep, and are no longer a thing. CI is simply building
packages these days

blitzmann:
I made the following changes:

* Remove from the Item class slot determination. It totally slipped my mind that the slot calculation on the Module class is a static method, so it's super easy to just call `Module.calulcateSlot(item)` and be done with it (also renamed it to not be a private method)
* Move the `Slot` enum outside of Module file. I wanted to be able to import it into config to use in the lookup dictionary instead of hardcoding the integers, but didn't want to bring in the entire module class and all of its dependencies (which probably would have broken things). This is something we should probably do across the board regardless - move the constants out into their own thing that doesn't have dependencies. 
* There was a bug when you updated the preference without first loading things into the ItemView. Pretty easy fix

All in all, excellent work! Thanks for the contribution! I've gone ahead and merged your commits along with my tweaks :)

Issue #1835, opened 2019-01-10
Pasting mutated item doesn't work on Mac OS X


DarkFenX:
1. Open any blank fit
2. Paste any mutated module exported from mutaplasmid.space

```
Gistii A-Type 1MN Afterburner
  Unstable 1MN Afterburner Mutaplasmid
  power 10.663900, capacitorNeed 10.944000, duration 10000.000000, cpu 13.890000, massAddition 500000.000000, speedFactor 155.136004
```

Actual:
Module is not imported, nothing happens

Expected:
Module is imported

Pyfa 2.7.0, Mac OS X 10.14 (18A391)

DarkFenX:
Fixed after merge of #1856

PR #1856, opened 2019-02-21
Mac Osx clipboard can store items with \r, ensure we split on newline or carriage return

WatchMeCalculate:
This PR addresses #1835 for mutated items split on  `\n` or `\r`  before parsing.

### DIFF
--- a/service/port/port.py
+++ b/service/port/port.py
@@ -208,8 +208,8 @@ class Port(object):
     def importAuto(cls, string, path=None, activeFit=None, iportuser=None):
         # type: (Port, str, str, object, IPortUser) -> object
         # Get first line and strip space symbols of it to avoid possible detection errors
-        firstLine = re.split("[\n\r]+", string.strip(), maxsplit=1)[0]
-        firstLine = firstLine.strip()
+        splitLines = re.split("[\n\r]+", string.strip())
+        firstLine = splitLines[0].strip()
 
         # If XML-style start of tag encountered, detect as XML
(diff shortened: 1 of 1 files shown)


WatchMeCalculate:
I tested this on Mac 10.14.3 as well as Windows 10

DarkFenX:
Thanks for looking into it.

DarkFenX:
@WatchMeCalculate could you provide text string (with \r and \n as they appear) so i could maybe change the fix? The reason I am asking is that if it's \r\n then we can use built-in splitlines(), however if there're \n\r then we have to stick to current solution.
```
>>> 'line1\r\nline2'.splitlines()
['line1', 'line2']
>>> 'line1\n\rline2'.splitlines()
['line1', '', 'line2']
```

WatchMeCalculate:
```
'Gistii A-Type 1MN Afterburner\r  Unstable 1MN Afterburner Mutaplasmid\r  power 10.663900, capacitorNeed 10.944000, duration 10000.000000, cpu 13.890000, massAddition 500000.000000, speedFactor 155.136004'
```
This was what i was seeing with your example string. There wasn't a mix of newlines and carriage returns

DarkFenX:
@WatchMeCalculate could you try copying any module from actual mutaplasmid.space (rather than github) and see what linebreaks are there? Also, could you tell which browser are you using?

DarkFenX:
Changed it in 5d0900337fe4cb89c321c158a7f688ea97424f01

Issue #1841, opened 2019-01-24
Add button to optimize fit price


DarkFenX:
During DB build time for each item type find

1. Direct replacements (same attributes, besides maybe meta lvl, so like domination and dark blood disruptors)
2. Better replacements (using highisgood)

Add some UI elements which can do price optmization by

1) doing replacements for strictly similar module, which is cheaper
2) doing replacements for strictly similar or better module, which is cheaper

Price should be force refreshed for all involved items, if it's older than 1-2 hours. 

I think I can implement backend support pretty easily (except for undo/redo stuff), what do people think about this feature in general and about how it should be handled on UI side?

blitzmann:
I really like this idea. On the ui can be a menu item in the edit menu which displays a list of modules with a list of replacements, if any, and show before and after price ( or maybe isk saved price) Maybe I can mock up an idea soonish

DarkFenX:
It should take a while to optimize as it takes some time to fetch the prices, so I am unsure it you can show it as menu items. I was more thinking about button in prices stats panel.

blitzmann:
Ah yeah, I was envisioning a window that, when opened, would go and fetch the prices of similar items for a fit, while displaying a loading message until prices have been updated (if needed). Basically a triggered fetch. Also, when I say menu item, I mean on the main pyfa window Edit menu for the fit as a whole, not individual module context menus.

Having it done automatically could work too, like you described. Price fetching should already be done in another thread so I think most of the work there is finished. Would just have to include the similar items with the actual items in the price request and they'll be updated if need be. Once prices have been fetch, can have a visual cue somewhere showing that optimizations are available. Cache threshold is 24 hrs instead of proposed 1-2, which I think is fine but we could tweak pretty easily if needed.

The delay that you talk about, is that a delay in the context menu showing up or a delay in the item stats window showing after you click the Stats menu item? I know that the context menu can take some time to show when it has lots of menus in it (variations, affecting skills, etc), it's always been a pain point, but dont think it's due to price fetching as I'm almost positive that prices are always collected in another thread. :)

DarkFenX:
You probably misunderstood me. I think that optimization process should be automatic, but triggering it should be manual. So that user presses button, and then pyfa handles everything automatically - price fetching, replacing etc. And if price is older than 2h by the moment user presses on optimization button, then it has to be re-fetched despite having normal 24h validity period. For some items price is quite volatile, as well as availability (eg TS small emp smartbombs).

[...]

PR #1857, opened 2019-02-21
Price optimize

DarkFenX:
See ticket #1841 for idea overview. Not ready to be merged yet, still work to be done. Creating to keep changelog:

- [x] price column for fighters tab
- [x] attempt to fetch price globally if Jita has no data on specified item
- [x] fix sorting in "compare" tab for price column
- [x] implement multibuy export price optimization main functionality
- [x] show that multibuy export price optimization is in progress (by graying out modal window)
- [x] fix bug with fit export window checkbox sizers
- [x] implement in-place price optimization main functionality
- [x] integrate in-place price optimization with undo/redo
- [x] block fit changes while in-place price optimization is in progress
- [x] decide how to invoke in-place price optimization


DarkFenX:
Added all needed backend changes and enabled price optimization for multibuy export.

DarkFenX:
What works:

- optimization by price in-place
- optimization by price for multibuy export

What still needs to be done:

- multibuy export price optimization: way to show that export is underway (in case network fetches take some time we should gray out export window or something, for now it will just wait until everything is done and close)
- in-place price optimization: way to show that it is in progress (block fit modification while prices are being fetched and fit is being modified)
- decide where to put in-place price optimization invocation (ATM i've added it to edit > optimize price, but this is mostly for testing needs) - would welcome some help on this. I cannot find good location for it.
- integrate in-place price optimization with undo/redo
- in fit export dialog, there is still bug which forces all checkbox blocks to have height of the biggest checkbox block. So if you have 2 blocks of checkboxes - 4 and 1, block with 1 checkboxes will take as much vertical space as 4. Previously it was masked by the fact that EFT export and multibuy export had 3 checkboxes, now multibuy has 4 (and i'd like to add one to ESI - "loaded charges").

On a side-note, while implementing this feature i had to drop idea of replacing items with "better" and cheaper alternative just because finding better item is not a trivial task and is huge maintenance burden due to inconsistencies in CCP data. Consider few examples:

-  durationBonus/durationSkillBonus: high duration on resist mods and prop mods is good while low duration is good for reps. So, there's no way to say if increase in duration is good without knowing context, highIsGood property is meaningless.
- speedFactor: high is good for ABs/MWDs, high is bad for webs.
- aoeCloudSizeBonus: high is good for GDs, high is bad for rigors/crash
- signatureRadiusBonus: high is good for TPs, high is bad for MWDs/x-instinct.

There were many more examples, but you get the gist. Moreover, highIsGood value is fucked up for quite many attributes which have no similar conflicts. By the time i found too many unsolvable inconsistencies (as listed above) I've added ~30 highIsGood overrides to db generation script to fix various hardwirings, modules and drugs (most of these were negative bonuses like -3% cycle time with highIsGood = 1)

So, now we just seek for items which are "the same" - like CN/DG invulns.

blitzmann:
@DarkFenX I added a wait dialog to halt GUI until the callback is called. Let me know if it works for you (seems to work for me on Windows :) )

DarkFenX:
Thanks! It works well.

DarkFenX:
Actually I tested it more thoroughly:

1) alignment of notification window is alright on xorg, but is not on wayland:
modal
2) This window does not prevent fit changes while fetch is in progress. This is not ideal but might be acceptable (introduces risk of various race conditions, but at least user has been warned).
3) It seems to be implemented as separate window, which is visible on alt-tab menu

DarkFenX:
Added window disabler. Now we have to implement UI blocking for multibuy optimization and probably can merge.

Context menu changes can be done outside of scope of this pull request.

Issue #1845, opened 2019-02-02
Feature Request: Support Ctrl+Backspace for deleting whole words in search bar


lunedis:
In many text applications you can use Ctrl+Backspace to delete whole words.

This does not work in Pyfa (tested on OS Win 7), it just writes a small rectangle.

blitzmann:
Hrm, interesting... Gonna mark this as `help wanted` - this is a pretty minor bug, but I think it might be easy enough for anyone to pick up and attempt a fix (or at least research into it). I just don't have a whole lot of time right now :(

WatchMeCalculate:
FWIW:
https://trac.wxwidgets.org/ticket/10344
this is a known issue/wontfix

blitzmann:
Well, that thread is a wild ride. Nice catch! Will review your PR and hopefully get this merged in / provide feedback

PR #1858, opened 2019-02-22
Adding support for ctrl+backspace in searchbars

WatchMeCalculate:
Gif of it working

This was tested on mac and windows and addresses #1845. I have a couple questions though
 1. Should this work on Mac? it's not a feature on OSX and i'm using the event: RawControlDown vs ControlDown because the former only encapsulates ctrl button for Mac and Windows, while ControlDown also looks at Mac command key
2. Was that the correct spot to put the helper function? I wanted to somehow unify the different search bars to not duplicate this logic, but it seemed like more work than it was worth so I centralized the textcontrol manipulation in the event further work needs to be done
3. an elaboration of (2), It would be nice to just add this to the global event of TextControl so this works for all text boxes in pyfa... But I'm unfamiliar with the wx framework and would need a bit of direction.

### DIFF
--- a/gui/builtinAdditionPanes/notesView.py
+++ b/gui/builtinAdditionPanes/notesView.py
@@ -4,6 +4,7 @@ import wx
 from service.fit import Fit
 import gui.globalEvents as GE
 import gui.mainFrame
+from gui.utils.helpers_wxPython import HandleCtrlBackspace
 
 
 class NotesView(wx.Panel):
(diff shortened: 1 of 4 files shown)


blitzmann:
@WatchMeCalculate 

Excellent work! I think the location of the helper function is fine. I agree that it would be nice to be able to handle this globally, however I'm not sure if there is a way. I'll ask on the wxPython group, because this will pop up for other text boxes throughout the application. But for now, fixing these text boxes (and I added fix to notes as well) is good enough

blitzmann:
I merged this, but it didn't auto close :/

Issue #1862, opened 2019-03-02
Move all enums to `eos.const` (and convert to built-in enums)


blitzmann:
Right now Enums are strewn around the various class files (eg: ImplantLocation is in `fit.py`). These can cause issues if you need just the Enum without all the other crust that the file imports.

I've created `eos.cont` and moved `Module.Slots` enum to that to better support a feature that was merged in. We should take some time to continue this process and move all the other definitions into this file.

Additionally, should evaluate using Python 3's built-in enums. This will cause some issues for some definitions (Slot was one of them), so we should be careful and test each application. But this should be a relatively easy code cleanup effort.

Going to mark this as `help wanted` ass it's generally low-level and doesn't require a whole lot of planning or complexity. If someone wants to take a whack at it before I can get around to it, feel free - just state your intention here :)

DarkFenX:
You can use https://github.com/pyfa-org/eos/tree/master/eos/const as reference. Pretty sure similar organization is one of the better (at least, if current eos uses any EVE constants).

blitzmann:
@DarkFenX This is a slightly different request, although the confusion doesn't help since I named the file `const` lol. We don't really have EVE constants anywhere in pyfa, and if we do they are scattered around. Not debating whether or not we should have a const fiile much like what new eos has eventually, but this was more directed towards all the enums that we have that are intertwined with other modules. This makes it difficult to simply import the enum without importing the rest of the things. For example, GUI needs to import ImplantLocation to use it. However, to import it, we must also import all the imports in `eos.saveddata.fit` which is probably something we don't really want. I ran into this problem for the Slot enum, where importing it from `eos.saveddata.module` caused errors because other things had to load that couldn't at that time. :)

IndictionEve:
It's done.
This was way more tricky then expected. From a freaking out code interpreter during rewriteing the code, to figuring out the different behavors of eos.enum to python.enum and make it work again.
I changed the code step by step and made a lot of tests to make sure everything works like before. 
I hope you are fine with the result.

blitzmann:
Thanks @IndictionEve for taking on this task! I haven't had time to really dig through the changes, but will hopefully get to it this week and provide feedback / merge as-is :) From what I've seen though on the PR, everything looks pretty solid.

PR #1869, opened 2019-03-07
Code cleanup enums #1862

IndictionEve:
Basic setup for #1862

- [x] Put all enums into cont.py
- [x] Setup a basic description
- [x] Renamed some of the enums to make their use more visible. 
- [x] Better description
- [x] More restructuring the unit by your whishes

Structure of the renamed enums:
- `Fitting*` for fit related enums
- `Esi*` for ESI related enums
- `Port*` for Export/Import related enums


DarkFenX:
1) PortMultiBuyItemType should be PortMultiBuyOptions (consistency & it can easily have more options not related to item type, see price_optimize branch).
2) @blitzmann does it make sense to move port-related enums into eos const file?

IndictionEve:
@DarkFenX 
regarding 1. Already done
regarding 3. Working on this. I want to get rid of the `esi.Enum` anyway and replace it by the Python 3 enum.

blitzmann:
>@blitzmann does it make sense to move port-related enums into eos const file?

No, lets keep those out. the only enums that should be moves are those already in the eos module. we can talk about where the other ones can go later if need be

> Could use IntEnum and @unique where applicable, but it's fluff.

I agree, but would also caution that this could potentially change the value of certain enums if the order of the values are moved. ddSince some enums are saved as data in the database, this could potentially have consequences (like changing all mid slots to high slots lol). Personally would recommend they remain as explicit values :)

IndictionEve:
Let's move the ESI and PORT enums into the service path. 
So we would have two central sources for enums:
1.  `eos/const.py` for _EOS_ related enums.
2.  `service/const.py` for _ESI_ and _port_ related enums.

IndictionEve:
> EDIT: ...I was confusing it with something else (something that automatically assigns the value)

The only class where `auto()` is used in `FittingAttrGroup()`.
Do we want to change this to exlicit values?

blitzmann:
@IndictionEve I think `FittingAttrGroup` is actually only used for the attribute grouping... ie: GUI stuff. Lets go ahead and move non-eos enums into another module (you suggested service, and I agree 😄)

IndictionEve:
Done

blitzmann:
I made a commit on your branch that fixes the various imports to point to the primary source, instead of where they originally were.

I want to make one more pass on the diff, but otherwise everything tested fine :)

blitzmann:
@IndictionEve please take a look at the new commits on your branch. There were a couple bugs that I fixed, and some imports fixed. If everything looks good to you, I'll merge this in :)

IndictionEve:
thanks! looks good. 

1.  gui/builtinItemStatsViews/itemAttributes.py 
i removed `import config` because it is marked as unused import in my IDE.
2.  gui/builtinViewColumns/state.py
changing the numeric expression by enums make sense :)
3. service/port/efs.py
Looks like my automatic refactoring failed here. Wondering why `PortEftRigSize` was unused, but didn't get any error or warnings. 

when do you plan to make a release? more then 3 month since the last official one.

blitzmann:
>i removed import config because it is marked as unused import in my IDE.

```   if info and config.debug:
            attrName += " ({})".format(info.ID)```

PyCharm indeed shows it as unused. I'm not sure why, but it is used. /shrug

> when do you plan to make a release? more then 3 month since the last official one.

I am hoping for next week. EVE hasn't had  very many data changes recently so we haven't been pressed to update, but we'r getting to that point where I want to get one out soon

Issue #1868, opened 2019-03-07
Empty market group -> move view to first non-empty left tab


SubjectXXX:
## Bug Report
Its not really a bug, but its really annoying and time consuming. When I open item in market, that has Complex version of it and then move to item, that doesnt, item list is empty and I need to manually click on bottom tab to find a populated item window. Mind you, I have large monitor and while clicking on market categories, to switch item types, there is alot to travel with mouse pointer to the bottom of the screen..

### Expected behavior:
When switching items in market, if item does not have selected type (Faction, Complex, Officier), focus selection on first left tab, that has item type.

### Actual behavior:
Moving from Small energy neuts, I selected Corpii B-type small energy neut, then I wanted to add Exile Booster. When I clicked on market tree to select Exile category, bottom item selection was empty, since it remained on Complex tab. I had to move mouse on bottom, to switch tab to Normal.

### Fits involved in EFT format (Edit > To Clipboard > EFT):
No need

### Release or development git branch? Please note the release version or commit hash:
2.7.0

DarkFenX:
At some point in the past, I coded it to work differently (i don't remember how exactly it behaved in this case though). While i was MIAing someone changed it, and i just assumed people were not happy with it - and didn't revert the changes.

Checking commit history, following commits changed logic (get more details via `git log -p -- gui/marketBrowser.py`):

- d6db30b7fccf2d8d4bdc3d087a1e9e872957e867
- a7b6647135326ae41088b2ebce067f33b840193a

Currently, logic of market jumps and switching market group is the same: it just switches market group and doesn't touch meta groups at all, which is very inconvenient in some cases.

How it used to work (i think):

- Market jump to item: switch market group and select meta group which corresponds to meta group of the item (deselect the rest).
- Market group switch: switch market group and keep selected meta groups if possible. If some meta groups become unavailable, buttons gray out but keep their state (so that when another market group selected - items from this meta group is shown). If no items in currently selected meta groups available - add first available meta group to selection to show something.

If there are no objections - I can restore this logic. I find it much more convenient than current implementation.

DarkFenX:
@blitzmann what are your thoughts?

Change requested by @SubjectXXX contradicts to #818 (as it clearly states that metagroup button states should be kept throughout application runtime).

However, to me old behavior was more convenient. I can understand how turning on all metagroups for search was annoying (and i can easily live w/o it).

However, there're some things I am missing:

[...]

PR #1883, opened 2019-03-17
Statspanel/Option: total-price calculation

IndictionEve:
One thing I never felt comfortable with, was the calculation of the total price in the statspanel.
For me (as FC) drones and cargo are absolutely relevant for the total price.
When creating fittings for solo PVP, project-related implants are an issue.
All these different requirements are already covered by the structure of the source code.
I have now made these structures available to the user via options.
python_2019-03-17_10-52-47

In addition, I have extended the hints to provide extended price information discretely, even without using the options.

image
image


DarkFenX:
@IndictionEve these options are already available via RMB menu on price panel, as far as i understand. In fact, I also enable cargo and drones into total price. We can set these two enabled as default.

IndictionEve:
Oh my god, that's one of the hidden features of pyfa.
Never had the idea to use RMB on this panel.
We should think about some kind of indicator, where popup menus for panels are available. 

How should we handle this PR: 
- Implement as additional way to set the prices?
- Just keep the hints?
- Abandon it completely?

blitzmann:
Yeah, I agree, it's not very visible. But I've never heard of a better suggestion on how to let the user know about it.

My opinion is we can keep the preference options as another way of setting which things are included. As for the tooltips, I'm not too sure if theres a use case for knowing the breakdown of cost by slot. Nor do I really thing we need a tooltip for the total as its displayed, since all that information is already displayed in the panel

What do you guys thin

blitzmann:
Actually, I have thought about adding a sort of menu icon to panels that have context menus. But I've never looked into it.

IndictionEve:
I am the kind of person who wants to know exactly what something is made of. 
and I love those statistics and analysis (Yeah, I know... ).
is this function useful? maybe not. 
is it a step backwards? i don't think so. 
is it interesting? for sure.

I was working on a new fleet doctrine the last few days. I found that the costs for the rigs and low slots would have made that fitting much more expensive then it worth.
Would I have recognized that without tooltip? Sure. 
But I always like it, when a tool delivers me information like that directly.
i don't want to push this. I just want you to understand my reasons. 
If you don't like it, I will gladly remove the tooltips. 😄

DarkFenX:
As a theorycrafter myself, I disagree. When you determine effectveness of a doctrine, cost of modules in specific slot doesn't matter. Some ships have 1 module which is more expensive than the rest of fitting combined with a ship (e.g. armor keres with mutated warp disruptor) and it makes sense. What matters is:

- what kind of stats do you get from it
- how much money does it cost upfront
- maintenance cost (ammo, drugs, fuel)
- logistics cost (rigs which are lost during repackaging, some modules which you cannot move due to volume restrictions)
- maybe some other costs i've forgotten

The rest is "for fun" and I rather would not put it anywhere in pyfa.

However, your pull request raises important question about UI: the fact that you did not know about that menu (and I also did not until 2 months ago) shows that it can be improved. I will create seprate ticket for stats panel context menus / main menu / preferences later today.

DarkFenX:
FYI when working on #1868 I decided to introduce set of options for metagroups buttons, and decided to move all market and price-related options there. To do that, I've grabbed commit from your branch. Also I enabled drones and cargo into total price as default.

As I still disagree with detailed popups, I will close this pull request now. "Total" price preferences will be merged into master with "market_metagroups" branch.

IndictionEve:
@DarkFenX thank you very much. I'm fine with that. 
sorry for the late answer, but i was not able to have a look in here for a couple of days.

Issue #1871, opened 2019-03-09
Disable undo/redo menu items when there's nothing to undo/redo


DarkFenX:
Currently these menu items are not disabled.

PR #1904, opened 2019-03-25
Logic to enable/disable redo/undo buttons

burnsypet:
#1871 
---
- Added some logic to enable and disable the undo and redo buttons depending on the contents of the command history as follows:
    - Disable both buttons if no commands in the history
    - Disable undo if there are commands in the history, but the current command is not (this means we are already at the beginning)
    - Disable redo if there are commands in the history, and the current command is the most recent (we are at the end)
    - Otherwise undo and redo are enabled

### DIFF
--- a/gui/mainMenuBar.py
+++ b/gui/mainMenuBar.py
@@ -179,6 +179,21 @@ class MainMenuBar(wx.MenuBar):
         self.Enable(wx.ID_COPY, enable)
         self.Enable(self.exportSkillsNeededId, enable)
 
+        command_history = self.mainFrame.command.GetCommands()
+        current_command = self.mainFrame.command.GetCurrentCommand()
+
+        self.Enable(wx.ID_UNDO, True)
+        self.Enable(wx.ID_REDO, True)
+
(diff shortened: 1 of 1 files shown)


blitzmann:
Code looks good, but we could probably use these methods to make it even more straight forward https://wxpython.org/Phoenix/docs/html/wx.CommandProcessor.html#wx.CommandProcessor.CanRedo

Thanks for the help with this! :smile:

burnsypet:
Oh nice, it can be replaced with this:

```
        if command.CanUndo():
            self.Enable(wx.ID_UNDO, True)
        if command.CanRedo():
            self.Enable(wx.ID_REDO, True)
```

blitzmann:
@burnsypet if you can make those changes and test, we should be all good for this PR :)

burnsypet:
Updated @blitzmann

blitzmann:
Awesome, thanks! Will need to check in os x and linux before merging - will do that this week :)

DarkFenX:
Works for me under linux.

Issue #1887, opened 2019-03-19
Options and UI: context menus, main menu, preferences overhaul


DarkFenX:
I am going to touch two issues with UI. Issue 1: some options are not visible. Currently, we're using two ways to set various options:

- stats panels context menus (incoming damage pattern, factor reload, target resists, price options)
- preferences (everything else)

When users explore functionality of pyfa, they see what's available in main menu, however it's not obvious at all that some stats panels have context menus available.

Issue 2: some main menu items feel out of place. `Window > Turn overrides on`, for example, or `Edit > Optimize Fit Price` which i added recently in `price_optimize branch`. In fact, two sections which look logical to me are `File` and `Help` (except for importing defaults part).

There might be many ways to solve them (feel free to suggest). I will post my proposal below,

blitzmann:
For the context menus on stats panel, I've had idea before of registering context menu to the panel itself, and to include a little menu icon in the header that, when clicked, opens the menu. This would be baked in functionality to our stats pane class. I dont think we have many which is spanlwn based on where you click in the panel, just panel overall, so I think thos would be a good solution.

Definitely open to ideas on main menu, I have no idea what to do with it

DarkFenX:
I would change following:

- move or copy some stats panel context menu items (to preferences or main menu, depending on how often they are needed)
- overhaul main menu structure

We have following options available in via stats context menu items:
- incoming damage pattern: duplicate to main menu (likely to be often used)
- factor reload: duplicate to main menu (likely to be often used)
- target resists: duplicate to main menu (likely to be often used)
- "total" price options (move to preferences)
- price cache reset (just remove, preferences already has such option)

Main menu structure change:


File
- new tab (ctrl-t)
- close tab (ctrl-w)
---
- backup all fittings
---
- quit (ctrl-q)

Fit
- undo (ctrl-z)
- redo (shift-ctrl-z)
---
- to clipboard (ctrl-c)
- from clipboard (ctrl-v)
---
- factor reload (ctrl-r)
- incoming damage pattern menu
- target resists
---
- optimize price (ctrl-?)
- graphs (ctrl-g)
---
- browse ESI fittings (ctrl-b)
- export to ESI (ctrl-e)

Character
- save character
- save character as
- revert character
---
- export skills needed
---
- manage ESI characters

Global
- disable fitting restrictions (ctrl-d, if it's global setting)
---
- attribute overrides
- turn overrides on/off
---
- preferences (ctrl-p)

Editors
[...]

PR #1913, opened 2019-04-03
Market meta category options

DarkFenX:
- [x] create separate preferences category for price & market options
- [x] make preference dialog bigger under GTK to get rid of horizontal scrolling
- [x] move existing market-panel related options to new preferences panel (item shortcuts, price source and system, market search delay)
- [x] move "total" price options to new preferences panel
- [x] add option which controls metagroup behavior during market jump
- [x] add option which controls metagroup behavior when no items are in group with current filter
- [x] add option which controls metagroup behavior during item search
- [x] add option which controls metagroup behavior when selecting market group in market browser
- [x] rework price panel context menu to remove cache drop item, and move drones / cargo / character directly to main context menu
- [x] check if it's possible to drop selection from market group when starting a search w/o changing scroll position in market tree panel
- [x] change icon of the new preference group


DarkFenX:
Preferences:

image

image

DarkFenX:
All the new logic is implemented besides search-specific mode. I have no yet settled on how it should behave. I have 3 possible variants in mind:

1) no changes
2) enable all meta groups
3) enable all meta groups (preserve metagroup state pre-search and restore it once user selects market group in browser or does market jump)

But i do not know which should be exposed to options, maybe all 3.

DarkFenX:
@blitzmann in metaButton.py, there's a workaround for some bug:
```
    def setMetaAvailable(self, isAvailable):
        self.Enable(isAvailable)
        # need to also SetValue(False) for windows because Enabled=False AND SetValue(True) looks enabled.
        if not isAvailable:
            self.SetValue(False)
```
Could you check if it's still needed? I think it would be beneficial to see button state even while it's disabled. Makes it easier to understand what will be selected when you click market group with all meta categories available.

If it's bugging out only on windows - i will change code to make it windows-specific.

DarkFenX:
I've implemented all the logic i wanted to implement i think. The only remaining issue i noticed is that when in search mode user clicks already selected item in market tree, it doesn't get back to normal mode (which has always been like that, but still a little annoying). Is there a way to let user select already selected market group during search (or, rather, let market browser know one way or another)?

burnsypet:
@DarkFenX I built your branch on Mac and ran through some scenarios and it looks great I could not find any issues, but I have two questions:

- Is it possible to have an option to enable all the available meta buttons when navigating between market groups? For example if I have all available Damage Control buttons enabled (Normal, Faction, Officer) and then change to Energized Adaptive Platings, the enabled buttons does not expand to include Complex which is available for this market group. It's possible to achieve this at the moment, but you have to manually enable all of the buttons by going to a group that has them all available first.

- How do we plan to communicate to users that the context menus are now located in a different place?  It felt a bit jarring not being able to follow my learned route for changing these options, and if I hadn't re-read your post I wouldn't have guessed there was a new Preferences tab especially since I don't use the Preferences menu very often.

DarkFenX:
@burnsypet well, it depends on what's the use-case. If i imagine situation right, i can see two options to resolve that:

- make disabled meta buttons togglable when ctrl is pressed
- introduce another set of options which kicks in when you click market group in market group tree (not via market jump). Obvious choices would be "no changes" and "enable all meta buttons"

What would you prefer and why? If you prefer second - do you see other option values which might be useful?

Regarding communicating to users - first of all, why do you ask, because of the price menu? How often did you use it and why? As for communicating - we will have to come up with something within scope of #1887 (which I also wanted to implement before next release), here I just did small part of it related to prices.

burnsypet:
I would prefer this option:

> introduce another set of options which kicks in when you click market group in market group tree (not via market jump). Obvious options would be "no changes" and "enable all meta buttons"

If it was possible to toggle the buttons when they were not available for the market group I think it would cause confusion.

My use case for the price context menu is:

If I want to check the cost of a fitting without drones/ammo because I know that I own those already then I right click the total cost and toggle off the ones I don't want to see. I frequently toggle these options when looking at fits I import from zKillboard, but when I create a new fit from nothing myself I prefer to have the total price contain everything. I usually toggle these options multiple times per session.

I can understand that for some users this might be a setting they set once and then forget about it, but it is a toggle that I use regularly and moving it to preferences increases the effort required.

DarkFenX:
Ok then my bad for removing them, i will put them back later.

DarkFenX:
I restored old context menus and added settings group for market group selection. However, i did not find quick way to implement it so will do that later.

Also, regarding pricing options, i think that context menu should show drones/cargo/character menu directly rather than subitem - for convenience (and clear prices menu item should be removed).

DarkFenX:
@burnsypet I've added full support for this option, now just some UI polish remaining.

DarkFenX:
I finished all the changes i wanted to implement, looking for feedback on following 2 things:

1) I've added market group unselection upon using search. Does it work under windows/mac without side-effects (like switching scrolling position)?
2) Meta button windows-specific workaround:
```
    def setMetaAvailable(self, isAvailable):
        self.Enable(isAvailable)
        # need to also SetValue(False) for windows because Enabled=False AND SetValue(True) looks enabled.
        if not isAvailable and "wxMSW" in wx.PlatformInfo:
            self.SetValue(False)
```
Is it still needed?

DarkFenX:
I'm going to merge it now, but still would like to get answers.

DarkFenX:
1) Works on windows
2) It's still needed

PR #1916, opened 2019-04-09
Stats menu indicators

blitzmann:
See #1887


DarkFenX:
Works for me. Merging.

PR #1973, opened 2019-05-14
Add link to Gentoo pyfa overlay

ZeroPointEnergy:
Do to the situation in Gentoo where there is still no wxpython-4 ebuild I had to look for another solution. Luckily there is an ebuild from a third-party available but that will probably never make it upstream as it is because it still needs a lot of work and none of the original package maintainers seem interested in doing it. Even the pull requests for the dependencies are moving very slowly. But it works for Pyfa and that is enough for me to be able to make a working Pyfa ebuild again.

So for now I created this overlay which should probably be easier to keep updated to deliver the most recent Pyfa versions and all required dependencies for Gentoo users.

I will also try to get this registered so it can be easily installed with the "layman" tool and keep the install instructions updated on the overlay README.

P.S. FreeBSD is UNIX not Linux :wink:

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,13 +21,14 @@ Apart from the official release, there is also a [Homebrew](http://brew.sh) opti
 $ brew install Caskroom/cask/pyfa
 ```
 
-### Linux Distro-specific Packages
+### Linux and BSD Distro-specific Packages
 The following is a list of pyfa packages available for certain distributions. Please note that these packages are maintained by third-parties and are not evaluated by the pyfa developers.
 
 * Debian/Ubuntu/derivatives: https://github.com/AdamMajer/Pyfa/releases
 * Arch: https://aur.archlinux.org/packages/pyfa/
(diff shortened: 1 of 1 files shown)


DarkFenX:
Wouldn't it make more sense to name it "Linux and BSD", as these terms seem to be from more similar categories? Rather than unix-like, which is an umbrella term for not just BSD but for Linux too - so if you mention UNIX it makes little sense to mention linux as well. If you agree - please update pull request.

Regarding wxpython - yeah it can be painful. I want to build new pyfa version, with Qt or web-based UI, as wx-related issues are quite problematic, but that's gonna take some time.

ZeroPointEnergy:
You are right, Linux and BSD makes probably more sense as Mac OS could also be considered a UNIX. I will update the PR.

QT would be very awesome, but sounds like a massive undertaking.

DarkFenX:
I checked these links and removed those which linked unmaintained packages. Removed BSD due to that as well.

Issue #1976, opened 2019-05-17
Feature Request: Include Implant Skill Requirements in Fitting Skill Requirements


jtaylormayfield:
## Bug Report
Ideally, the overall fit skill requirements should include implant skill requirements. Currently they do not.

### Expected behavior:
Fit skill prerequisites include implant skill requirements.

### Actual behavior:
Fit skill prerequisites ignore implant skill requirements.

### Detailed steps to reproduce:

1. Create a new Impairor fit.
2. Set character to "All 0"
3. Ensure basic skill prerequisites have been met (Change Ship Skills > Spaceship Command > Level 1)
4. From the market, fit an ORE Mining Director Mindlink.

### Fits involved in EFT format (Edit > To Clipboard > EFT):
[Impairor, test]

[Empty Low slot]
[Empty Low slot]

[Empty Med slot]
[Empty Med slot]

[Empty High slot]
[Empty High slot]

ORE Mining Director Mindlink

### Release or development git branch? Please note the release version or commit hash:
Release 2.9.3

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
Windows 10

DarkFenX:
Merged, thanks!

PR #1977, opened 2019-05-17
Applied implant and booster skill check

jtaylormayfield:
Include skill prerequisites for applied implants and boosters in fit skill check.


jtaylormayfield:
Related issue: https://github.com/pyfa-org/Pyfa/issues/1976

PR #1981, opened 2019-05-21
Fixed issue with character name extending into text box

zwparchman:
The character selection combo box in the top right of the main window was being cut off on my system. I have resolved that issue. Thanks for the wonderful tool!

Before
2019-05-20-202256_1920x1080_scrot

After
2019-05-20-202330_1920x1080_scrot

### DIFF
--- a/gui/characterSelection.py
+++ b/gui/characterSelection.py
@@ -46,7 +46,7 @@ class CharacterSelection(wx.Panel):
         self.charCache = None
 
         self.charChoice = wx.Choice(self)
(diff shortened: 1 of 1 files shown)


zwparchman:
And I just realized I lost the buttons to the side. I'll reopen this when I fix that issue.

DarkFenX:
It can be annoying. I've brought this issue up to @blitzmann and he tinkered with it a little, but hasn't come up with solution.

Core issue here is that GTK widget takes width of the longest character name. You have to cut character name somehow or do some UI magic to fix that.

PR #1987, opened 2019-05-26
improve sizing on esi fit import browser

AaronOpfer:
On my machine, using the ESI import feature and clicking "Fetch Fits" gives the appearance of nothing happening. Turns out, this was because by default the tree on the left-hand side of the window has zero width by default and the parent window needs to be resized to see the fits. Fix this by setting a minimum size for the fitting browser tree and increase the default window size to compensate.

### DIFF
--- a/gui/esiFittings.py
+++ b/gui/esiFittings.py
@@ -24,7 +24,7 @@ class EveFittings(wx.Frame):
 
     def __init__(self, parent):
         wx.Frame.__init__(self, parent, id=wx.ID_ANY, title="Browse EVE Fittings", pos=wx.DefaultPosition,
-                          size=wx.Size(550, 450), style=wx.DEFAULT_FRAME_STYLE | wx.TAB_TRAVERSAL)
+                          size=wx.Size(750, 450), style=wx.DEFAULT_FRAME_STYLE | wx.TAB_TRAVERSAL)
 
         self.SetBackgroundColour(wx.SystemSettings.GetColour(wx.SYS_COLOUR_BTNFACE))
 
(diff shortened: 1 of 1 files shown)


DarkFenX:
How does it look for you pre-fix?

AaronOpfer:
image

Debian Stable, pyfa is master 1d8c9d2c4043e0aef98e77d397c5fbcc5c31a554, UI scaling factor I think is 125% but I'm not sure. 2560x1440 resolution.

`pip freeze`
```
asn1crypto==0.24.0
beautifulsoup4==4.6.3
certifi==2018.10.15
cffi==1.11.5
chardet==3.0.4
cryptography==2.4.1
cycler==0.10.0
diskcache==3.1.0
filelock==3.0.10
idna==2.7
kiwisolver==1.0.1
Logbook==1.4.1
markdown2==2.3.6
matplotlib==3.0.2
numpy==1.15.4
packaging==18.0
pluggy==0.8.0
py==1.7.0
pycparser==2.19
pyparsing==2.3.0
Pypubsub==4.0.0
python-dateutil==2.7.5
PyYAML==3.13
requests==2.20.1
roman==3.1
six==1.11.0
snakeviz==2.0.0
SQLAlchemy==1.2.14
toml==0.10.0
tornado==6.0.2
tox==3.5.3
urllib3==1.24.1
virtualenv==16.1.0
wxPython==4.0.3
```

DarkFenX:
Which DE and display server do you use? Fractional scaling factors are experimental as far as i know. They might affect that. Also, wx 4.0.3 might too.

This is how it looks on my debian testing rig (wx 4.0.4, wayland, gnome, 100% scale):

image

DarkFenX:
That said, I do not see anything bad in making the window a little bigger, merging.

PR #2050, opened 2019-10-02
Support loading images from zip

ZeroPointEnergy:
This commit reenables the ability to load the images from a zip file instead of a directory structure. It is possible to set the location of this file via configforced.py

This possibility was partly present at one time but was commented out and not really ported to python3. I maintain this patch in the Gentoo overlay since I package the images into a zip. It would be great if we could upstream this functionality but no hard feelings if this isn't something you want :wink:  .

### DIFF
--- a/gui/bitmap_loader.py
+++ b/gui/bitmap_loader.py
@@ -19,6 +19,7 @@
 
 import io
 import os.path
+import zipfile
 from collections import OrderedDict
 
 # noinspection PyPackageRequirements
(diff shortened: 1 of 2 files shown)


DarkFenX:
Yeah, huge effect folder and huge image folder are both painful. We changed effects already, and I did not touch images for one reason - it might increase load times considerably, so I will have to check that. Effects were different story as majority of them are used in any realistic operation.

gui/bitmap_loader.py
```diff
@@ -31,16 +32,13 @@
...
-    # try:
-    #     archive = zipfile.ZipFile(os.path.join(config.pyfaPath, 'imgs.zip'), 'r')
-    #     logging.info("Using zipped image files.")
-    # except (IOError, TypeError):
-    #     logging.info("Using local image files.")
-    #     archive = None
-
-    pyfalog.info("Using local image files.")
-    archive = None
+class BitmapLoader(object):
```
DarkFenX: Remove (object), it's legacy py2 thing - all of objects in py3 are "new"

DarkFenX:
Just realized that it's about adding support, rather than actually moving images to an archive. I guess there can be no objections to accepting it then.

ZeroPointEnergy:
Yes it is completely optional. It will try to load the zip specified in `imgsZIP` or `imgs.zip` in the applications root directory. If it doesn't find it, it will simply load the images from the directory as before.

PR #2063, opened 2019-10-16
Re-written in much greater detail instructions how to set up the project

Gochim:
It wasn't very clear how to set up project when I decided to contribute myself, so decided as a first task to create clear and simple guide. 

- Written instructions how to set up the project in CLI.
- Added new guide to set project up using PyCharm/IntelliJ. 
- Since it took too much space - separated in into separate file *contributing.md*.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,11 +4,11 @@
 
 - Python 3.6
 - Git CLI installed
-- Python, pip and git are available as commands (add to path if needed)
-
-## Setting up the project manually
+- Python, pip and git are all available as command-line commands (add to path if needed)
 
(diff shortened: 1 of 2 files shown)


DarkFenX:
Thanks for writing it!

DarkFenX:
Actually I wonder if having it on wiki would be more reasonable, and just linking it from readme?

@Gochim, could you check [the wiki](https://github.com/pyfa-org/Pyfa/wiki) please and tell what you think?

Gochim:
Well it's definitely a bit old. It seems to me that usually people look at readme.md or contributing.md first, so they should stay where they are. And copy/pasting it into wiki is not a good way either. So, if you don't mind, I'll simply add a link to contributing.md into the wiki page - so we'll keep one source of information.

DarkFenX:
I do not. In fact, I've asked exactly because I have no idea where people look first. Did not know that contributing.md was a standard.

Gochim:
Well, my Udacity course at least said so :) 

And... done.

Issue #2084, opened 2019-11-14
Cant copy fits


Felony67:
Error message:

pyfa v2.14.2
EVE Data Version: 1604553 (2019-11-12 07:15:35)

OS version: Windows-7-6.1.7601-SP1
Python version: 3.6.8 (tags/v3.6.8:3c6b436a57, Dec 23 2018, 23:31:17) [MSC v.1916 32 bit (Intel)]
wxPython version: 4.0.6 (wxWidgets 3.0.5)
SQLAlchemy version: 1.3.11
Logbook version: 1.5.3
Requests version: 2.22.0
Dateutil version: 2.8.1

####################

Traceback (most recent call last):
  File "C:\projects\pyfa\gui\mainFrame.py", line 780, in exportToClipboard
  File "C:\projects\pyfa\gui\copySelectDialog.py", line 117, in __init__
  File "C:\projects\pyfa\gui\copySelectDialog.py", line 158, in toggleOptions
  File "C:\projects\pyfa\gui\copySelectDialog.py", line 161, in GetSelected
AttributeError: 'CopySelectDialog' object has no attribute 'copyFormat'

I have tried reinstalling

DarkFenX:
I believe this issue is caused by export format which is not available which is stored in settings. It is entirely possible as XML format has been disabled for export a while back. Can be replicated with this patch:

```
$ git diff
diff --git a/gui/copySelectDialog.py b/gui/copySelectDialog.py
index 9beb9698..22f2e651 100644
--- a/gui/copySelectDialog.py
+++ b/gui/copySelectDialog.py
@@ -91,7 +91,7 @@ class CopySelectDialog(wx.Dialog):
             else:
                 rdo = wx.RadioButton(self, wx.ID_ANY, formatName)
             rdo.Bind(wx.EVT_RADIOBUTTON, self.Selected)
-            if self.settings['format'] == formatId:
+            if self.settings['format'] == formatId and False:
                 rdo.SetValue(True)
                 self.copyFormat = formatId
             mainSizer.Add(rdo, 0, wx.EXPAND | wx.ALL, 5)
```

Which results in following traceback:
```
pyfa v2.14.2
EVE Data Version: 1604553 (2019-11-12 15:15:35)

OS version: Linux-5.2.0-3-amd64-x86_64-with-debian-bullseye-sid
Python version: 3.7.5 (default, Oct 27 2019, 15:43:29) 
[GCC 9.2.1 20191022]
[...]

PR #2085, opened 2019-11-14
Possible fix for #2084

Gochim:
Not sure that it will fix the problem since I don't see the reason for it in the code in the first place. But at least it should fix that no copyFormat variable is defines


DarkFenX:
This fix will work but please change it to `self.copyFormat = self.copyFormatEft`.

Gochim:
Sure. Fixed both in here and in default values.

DarkFenX:
Looks fine, merged.

PR #2094, opened 2019-11-22
Add keyboard shortcuts to character level editor

RussianCow:
This adds the ability to set the level of a skill by pressing a number key, e.g. pressing 1 will set the skill's level to 1, and so on. I didn't add a way to "unlearn" a skill via a shortcut, but I can easily make it respond to `Delete` or something else.

I've never used wxPython before, so sorry if using a custom event isn't the best way to achieve it.


DarkFenX:
I think it's alright, although I will change few things here and there after merging it.

Issue #2073, opened 2019-10-30
[Implants] Add ability to create an Implant Set Editor group from the fit view


Gochim:
Generally simplify work with implants groups. 

## Suggestion
Give an ability to create new set in the Implant Set Editor from the Implants tab in the main view of the fit.

## Expected behavior 
In the main fit window, switch to the "Implant" tab. In the context menu select "Create new implant set". On selection - ask for the name of a new group.

--- Edit. By "Create new implant set" I mean create a new group from implants that are currently in the "implants" list for the fit.

DarkFenX:
Accepted pull request with some extra renames. Will fix some bugs later, and maybe will add ability to disable it.

DarkFenX:
There's issue with it: when user presses "cancel", it opens implant set editor anyway.

DarkFenX:
Changed in 8c2788fd78a4b2576e0bb526f8d67d263e2a7e7b

PR #2103, opened 2019-12-12
Issue #2073: Added ability to add a list of implants from implant view to a new implant set

Gochim:
Did two things:
- Added new menu "Add As New Implant Set". After entering the name for the set (and checked that such name is not used), implant set editor is opened with focus on the new set.
- Renamed "Add implant set" to "Load Implant Set" for consistency.


DarkFenX:
Maybe "Apply Implant Set" would be better than "Load Implant Set"?

Also, i would make this context menu disablable via preferences. Idk about others, but for me having somewhat clean menu is a must, so I just disable things which i never use (like "Fill with module").

Gochim:
> Maybe "Apply Implant Set" would be better than "Load Implant Set"?
> 
> Also, i would make this context menu disablable via preferences. Idk about others, but for me having somewhat clean menu is a must, so I just disable things which i never use (like "Fill with module", and this context menu as well).

Sure,  "Apply Implant Set" makes more sense.

As for the disabling them via preferences - it is only visible in Implants section. So it won't clutter any other popup menu. And I'm not sure if implementing a checkbox for literally two popup menu items that are visible only in one tab is worth it... What do you think?

Issue #2128, opened 2020-01-28
Pyfa 2.16.2/3 Started flipping the UI to "RTL"


leonbur:
A picture is worth a thousand words:

![Capture](https://user-images.githubusercontent.com/8167543/73277063-4db7be80-41f2-11ea-8f9f-ef728611abee.PNG)

This started as I updated from 2.16.0 to 2.16.2 as well as 2.16.3.
I am located in the middle east but my Windows is localed to EN-US. 

Environment:
Windows 10 64bit, Python 3.6.

DarkFenX:
This is likely caused by update of wxpython 4.0.7. Could you make screenshots of all relevant regional settings in windows? I will try to reproduce it.

DarkFenX:
https://github.com/wxWidgets/Phoenix/issues/1008 this ticket contains more info, please try changing the setting the ticket mentions

leonbur:
My mistake, it appears that my region is set to Israel (even though the UI localization remains English). When I set it to United States the Pyfa UI goes back to normal. 
It's still a bug though, right?

Current Region:
![current region](https://user-images.githubusercontent.com/8167543/73356430-050c0e00-42a3-11ea-840e-5606fad1f068.PNG)

All Regions:
![all regions](https://user-images.githubusercontent.com/8167543/73356448-11906680-42a3-11ea-9501-ea6e10966a4c.PNG)

DarkFenX:
Ok as a next step, could you confirm again that 2.16.1 worked for you even if it's set to Israel?

leonbur:
Tested both 2.16.1 and 2.16.0:

2.16.1 is incorrect just like the newer versions
2.16.0 works correctly

DarkFenX:
I posted a comment to the issue i linked above. I do not think there's anything I can do, except for switching back to wxpython 4.0.6 (which I will probably do, as 4.0.7 doesn't seem to fix anything from our other upstream bugs).

DarkFenX:
@leonbur https://ci.appveyor.com/project/DarkFenX/pyfa/builds/30527549/artifacts try one of these builds.

To fix several issues, I had to update multiple libraries and make pyfa a 64-bit application. Among those library updates there was python update to 3.8, which should fix this issue for you (if info from wxpython maintainers is correct).

leonbur:
The new build doesn't solve the issue.

DarkFenX:
[...]

Issue #2136, opened 2020-02-13
2.17.1 Opening character editor throws an error


leonbur:
Title says it all. here's the error window output:

> pyfa v2.17.1
> EVE Data Version: 1663997 (2020-02-11 15:13:50)
> 
> OS version: Windows-10-10.0.18362-SP0
> Python version: 3.8.0 (tags/v3.8.0:fa919fd, Oct 14 2019, 19:37:50) [MSC v.1916 64 bit (AMD64)]
> wxPython version: 4.0.7.post2 (wxWidgets 3.0.5)
> SQLAlchemy version: 1.3.13
> Logbook version: 1.5.3
> Requests version: 2.22.0
> Dateutil version: 2.8.1
> 
> ####################
> 
> Traceback (most recent call last):
>   File "gui\characterSelection.py", line 183, in charChanged
>   File "gui\mainFrame.py", line 443, in OnShowCharacterEditor
>   File "gui\auxFrame.py", line 59, in openOne
>   File "gui\characterEditor.py", line 170, in __init__
>   File "gui\characterEditor.py", line 111, in __init__
>   File "gui\builtinViews\entityEditor.py", line 57, in __init__
> wx._core.wxAssertionError: C++ assertion "strcmp(setlocale(LC_ALL, NULL), "C") == 0" failed at ..\..\src\common\intl.cpp(1579) in wxLocale::GetInfo(): You probably called setlocale() directly instead of using wxLocale and now there is a mismatch between C/C++ and Windows locale.
> Things are going to break, please only change locale by creating wxLocale objects to avoid this!
>

DarkFenX:
Thanks. And meh, maybe I will have to revert to wxpython 4.0.6 after all, unless I will find a way to work around this issue too.

DarkFenX:
This is another manifestation of an upstream issue, which was originally filed for pyfa as #2134. Added comment to actual upstream issue: https://github.com/wxWidgets/Phoenix/issues/1515#issuecomment-585764207

Will fix by rolling back to 4.0.6

pezz:
Just FYI - updated Arch Linux to 2.17.2 and the character editor works fine, python-wxpython version on Arch is 4.0.7.2 (last updated Feb 04).

Cheers.

DarkFenX:
Yeah it is most likely windows-specific issue. Still there is a chance you can reproduce it under linux, but you need specific locale settings for that.

pezz:
[...]

Issue #2134, opened 2020-02-11
Pyfa 2.17 crash on startup


Thoreas:
### Expected behavior:
Running pyfa.exe brings up the GUI.

### Actual behavior:
Running pyfa.exe causes it to crash before any GUI is displayed.

### Release or development git branch? Please note the release version or commit hash:
Release 2.17.0

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
Same scenario on two separate PCs:
Windows 10 Build 17763.1012
Windows 10 Build 18362.356

### Other relevant information:
Log:

```[2020-02-11 16:23:25.467177] INFO: __main__: Starting Pyfa
[2020-02-11 16:23:25.467177] INFO: __main__: 
[2020-02-11 16:23:25.467177] INFO: __main__: Writing log file to: C:\Users\Swevener\.pyfa\pyfa.log
[2020-02-11 16:23:25.467177] INFO: __main__: Running in a frozen state.
[2020-02-11 16:23:25.512188] INFO: eos.db: Initializing database
[2020-02-11 16:23:25.512188] INFO: eos.db: Gamedata connection: sqlite:///C:\Games\Pyfa\eve.db?check_same_thread=False
[2020-02-11 16:23:25.512188] INFO: eos.db: Saveddata connection: sqlite:///C:\Users\Swevener\.pyfa\saveddata.db?check_same_thread=False
[2020-02-11 16:23:25.512188] DEBUG: eos.db: Initializing gamedata
[2020-02-11 16:23:25.516188] DEBUG: eos.db: Getting gamedata version
[2020-02-11 16:23:25.518189] DEBUG: eos.db: Initializing saveddata
[2020-02-11 16:23:25.519189] DEBUG: eos.db: Importing gamedata DB scheme
[2020-02-11 16:23:25.566679] DEBUG: eos.db: Importing saveddata DB scheme
[2020-02-11 16:23:25.606756] DEBUG: eos.db: Importing gamedata queries
[2020-02-11 16:23:25.606756] DEBUG: eos.db: Importing saveddata queries
[2020-02-11 16:23:25.607756] DEBUG: service.prefetch: Run database migration.
[2020-02-11 16:23:25.611757] DEBUG: service.prefetch: Starting database validation.
[2020-02-11 16:23:25.611757] DEBUG: eos.db.saveddata.databaseRepair: Running database cleanup for character skills.
[...]

PR #2143, opened 2020-03-09
Update the new requirements.txt

vivian-oier:
On WIndows 10 x64 with Python3.8.

For some reasons, we have no way to use old version.

Changed new one.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,14 @@
-wxPython == 4.0.6
+wxPython >= 4.0.7.post2
 logbook >= 1.0.0
 matplotlib >= 3.1.2
 python-dateutil
 requests >= 2.0.0
-sqlalchemy >= 1.3.0
-cryptography >= 2.3
-markdown2 >= 2.3.5
-packaging >= 16.8
-roman >= 2.0.0
-beautifulsoup4 >= 4.6.0
-pyyaml >= 5.1
+cryptography
+diskcache
+markdown2
+packaging
+roman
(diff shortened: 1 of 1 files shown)


DarkFenX:
Actually you can use older wxpython, you just need to build it (and to install windows SDK to do that). But in your case the easiest solution is to use python 3.7 i think.

While I have no idea about details on the situation you are in, please understand pyfa's:
- 4.0.6 works alright, but it had several issues - for example, #2128
- I decided to do some facelifting and updated a number of dependencies, including python to 3.8 and wxpython to 4.0.7.post2. Soon i started getting reports about crashes, e.g. #2134. They happened on systems which used some specific "format" setting in windows. I also fixed #2128 by resetting app's locale, as maintainers of wxpython suggested. This locale to which app is reset also led to crashes on all systems, regardless of "format" regional setting on Windows. I traced issue back to "resolution" parameter of PNGs which wx could not parse. To fix that, i stripped this parameter from all the PNGs which were bundled with pyfa. Despite those efforts, issue #2136 popped up. It revealed that images bundled with wx itself are affected (and pyfa uses those in many areas). I have no control over that so I just downgraded build system to python 3.7 and wxpython to 4.0.6.

You can visit linked issues and see links to upstream issues. If you have any idea how to fix that, you are welcome. But until that or until wxpython get their shit together, build system's python version stays at 3.7 and wxpython version stays at 4.0.6. There is no reason to upgrade, just set of unsolvable issues.

Issue #2045, opened 2019-09-18
Automatically add Pyfa character when adding ESI character


Zixradoom:
### Expected behavior:
ESI character available in the character drop down

### Actual behavior:
ESI characters not available in the character drop down

### Detailed steps to reproduce:
1) fresh install of pyfa on windows 10
2) Character Menu Option -> Manage ESI Characters -> Add Character
3) choose fit
4) click "Character:" drop down top right
5) character not available

### Fits involved in EFT format (Edit > To Clipboard > EFT):
[Vargur, Umbra II]

Tracking Enhancer II
Tracking Enhancer II
Gyrostabilizer II
Gyrostabilizer II
Gyrostabilizer II

Gist B-Type Large Shield Booster
Adaptive Invulnerability Field II
Adaptive Invulnerability Field II
Republic Fleet Large Cap Battery
Large Micro Jump Drive
Tracking Computer II

800mm Repeating Cannon II
800mm Repeating Cannon II
800mm Repeating Cannon II
800mm Repeating Cannon II
Small Tractor Beam II
Salvager II
Salvager II
Bastion Module I

Large Projectile Burst Aerator II
Large Capacitor Control Circuit I

Salvage Drone I x5
Warrior II x10

### Release or development git branch? Please note the release version or commit hash:
Version v2.11.0
EVE Data Version: 1564394 (2019-09-10 07:28:33)

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
Windows 10

DarkFenX:
Yeah, this is not how it works (but how it should).., you add ESI character, but even then you have to add Pyfa character, go to EVE SSO tab, pick ESI character (thus linking the two) and pull the data.

I will change this issue to be improvement, so that pyfa creates new character which has name of added ESI character with his skills, if such character hasn't been created yet.

DarkFenX:
Done in #2159, also adjusted the behavior a little in dcb058a718ada4e1539fb6b198654a0ce30cf7c0

PR #2159, opened 2020-04-10
Regarding #2045: automatically create a character for a new sso character 

soro:
and add button to retrieve all chars directly linked to existing sso characters.

I just ran into this issue today and was annoyed enough to just fix it. Let me know if you see any issues with this patch.

### DIFF
--- a/gui/characterEditor.py
+++ b/gui/characterEditor.py
@@ -833,7 +833,7 @@ class APIView(wx.Panel):
     def fetchSkills(self, evt):
         sChar = Character.getInstance()
         char = self.charEditor.entityEditor.getActiveEntity()
-        sChar.apiFetch(char.ID, self.__fetchCallback)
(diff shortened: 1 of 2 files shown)


DarkFenX:
I welcome this change. Thanks, will check it tomorrow before merging.

DarkFenX:
@soro i checked it and it looks fine. However, what do you think about scenario when there is a character with the same name as the character being added? Should it update existing character or add a new one?

Currently, it adds a duplicate. At first glance i think updating existing character would make more sense. But, maybe there are some scenarios where this is an undesirable behavior I couldn't think of.

If you have no time - let me know, I can merge and fix it myself.

soro:
Yeah I don't really have time to make the change for a variety of reasons, sorry.

DarkFenX:
It should happen if you remove SSO credentials in pyfa and it again later. But okay, I will handle that. Thanks for the PR.

Issue #1782, opened 2018-10-26
Feature Request --Night Mode


cptashen:
Hi, first time using this site.

I'm wondering if it would be possible to create a night mode for Pyfa 

Thanks

~Ash

blitzmann:
Hi @cptashen!

A few attempts have been made to create simple theming support. I actually have an older branch that kinda sort of does it, but there's some serious issues with it. One of the major issues is that that are some widgets that just don't seem to play nice with it. For example, the fitting window column headers - it's impossible to change these from white on Windows.

We do support system default colors out of the box for linux operating systems. I know that means nothing to windows users though. >_>

This is something that I revisit every few months to try to devise a solution that looks good, but it isn't high on the priority list, at least it hasn't been in the past. There are a few tricks we might be able to utilize now that pyfa 2.0 ships with wxPython 4, maybe I should take another look at this soon :)

DarkFenX:
Duplicate of #1206

Issue #1206, opened 2017-06-01
[Request] Dark/Nightmode


Seismix:
Even though I have no idea how feasible that is, is there any chance we could get a dark/nightmode pyfa? It's a lot easier on the eyes, even when f.lux is enabled.

blitzmann:
@Seismix It's not impossible, but it would require a sit down and review of basically all GUI panels in pyfa, and then add code to change the default system colors. It's been requested before, just haven't ever really got around to figuring it out.

Will keep this one around for a rainy day. (note, at least in Linux, if you use a dark theme, pyfa inherits it IIRC. So just switch to Linux! :P)

blitzmann:
@Seismix It's not impossible, but it would require a sit down and review of basically all GUI panels in pyfa, and then add code to change the default system colors. It's been requested before, just haven't ever really got around to figuring it out.

Will keep this one around for a rainy day. (note, at least in Linux, if you use a dark theme, pyfa inherits it IIRC. So just switch to Linux! :P)

Folles:
yes please, 1000 times this please.

Ebag333:
Still got a long way to go, but it's a start....

https://puu.sh/wd8yc/91a796fa0f.png

(shown with the teletype font  for maximum console effect 😁 )

Ebag333:
Annnnd it's done.

Personally I don't much care for the dark modes (especially since there are some GUI elements that are difficult/impossible to change color on), but who am I to judge? :)

You can find a release here, but keep in mind that this is NOT Pyfa, but a fork.
https://github.com/Pyfa-fit/Pyfa-fit/releases/tag/2017.06.08

@blitzmann feel free to merge this into Pyfa, it's mostly in a single PR.
https://github.com/Pyfa-fit/Pyfa-fit/pull/123

Folles:
does this mean that you'll never implement it into the main version of the program?

Ebag333:
>does this mean that you'll never implement it into the main version of the program?

It means that it's entirely up to Blitzmann to implement. The work is done.

I'm also tweaking it a bit further. Already tweaked it so that the background of the inner panes is subtly different than the main frame, as well as a bunch of improvements to scaling based on font size.
https://puu.sh/wkYf4/6439d5e5b2.png

I'm not one for white text on a dark background, but that particular scheme I rather like. :)

Folles:
so relaxing to the eyes... thank you so much

Folles:
[...]

PR #2175, opened 2020-04-19
Added initial dark mode settings

mccloudy:
The beginning of dark mode.  

What currently works:
- All main panels (fittings, ship browser, bottom tab panes, market)

What doesn't work:
- Right-click context menus
- Main menu bar (and submenus)
- Other windows that can be opened from the 
- Redrawn bitmap buttons post-click
- Highlighted lines in fitting view (selected line is blue/white)

Bugs:
- Issue with color-coded slot in fit overriding dark mode (looks nasty)


DarkFenX:
Obligatory #1206 link since this PR is supposed to fix it.

While I do not think it is possible to make it pretty (some elements are impossible to paint IIRC), I am inclined to accept it since some folks miss it and in this case you went ahead and implemented it. However, I think there should be some text next to the option which describes that this is a hack and people should not expect good looks from it.

Before merging, gonna ask a few questions:

- have you tested it on macOS?
- have you tested it on linux?
- do you plan to try adding "darkening" of elements from "doesn't work" list?
- are you willing to fix various issues, if I find any before merging?

Just to clarify: those are no-pressure questions. I have access to both linux and mac so I just might test yourself, but if you did already, I might just skip this part. Same goes to fixing major issues - I can handle them but it would be cool if you did.

mccloudy:
There are probably some elements that can't be changed, but I think the vast majority can be.  Others can be, but there are quirks (e.g. certain menu elements can have color set before being attached to the menu bar but not after).

As for your questions:

- I have not tested it on Linux or Mac OS.  It worked fine for me under Windows Subsystem for Linux with Windows 10, so I'm assuming it'll work in a native linux environment.
- I do plan on adding elements from the "not working" list, because I hate wxPython and need to learn more about it so that my hate is more justified :P 
- The color contrast may be too high, but then again, there are many EVE redditors who would complain about the air they breathe, so I take it with a grain of salt.  But yes, I'm happy to look into a more "discord-like" coolor scheme.

My original intent was to build a color picker for the color scheme, but given the nature of the code, felt this would be easier to do first.  If implemented, people could pick their custom color settings.

DarkFenX:
```
 mode change 100755 => 100644 pyfa.py
 mode change 100755 => 100644 scripts/effectUsedBy.py
 mode change 100755 => 100644 scripts/findNonMarket.py
 mode change 100755 => 100644 scripts/itemDiff.py
 mode change 100755 => 100644 service/port/efs.py
```
This has to be fixed (first 4, last one is fine to leave without execution permissions). As for code and functionality on linux / mac - gonna test it later today, if I have some time.

DarkFenX:
While testing on linux:
```
  File "/home/dfx/Workspace/eve/pyfa/gui/mainFrame.py", line 539, in OnShowPreferenceDialog
    with PreferenceDialog(self) as dlg:
  File "/home/dfx/Workspace/eve/pyfa/gui/preferenceDialog.py", line 66, in __init__
    prefView.populatePanel(page)
  File "/home/dfx/Workspace/eve/pyfa/gui/builtinPreferenceViews/pyfaGeneralPreferences.py", line 99, in populatePanel
    self.cbToggleDarkMode.SetValue(self.sFit.serviceFittingOptions["useDarkMode"])
TypeError: CheckBox.SetValue(): argument 1 has unexpected type 'NoneType'
```

DarkFenX:
On linux, colors are different. This is how it looks on current master:

skin_default

This is how light skin looks on your branch:

skin_light

Having search field background as white is fine, the rest needs to be fixed.

mccloudy:
Understood.  I'll fix these.

Curious about the background coloring mismatch, but I think I know what's going on there.

DarkFenX:
Background color might need some platform-specific code, mind you. I haven't checked what kinds of colors are used on each platform, though.

mccloudy:
The reason for the color differentiation is because of mainFrame's recursive call to all its children, indiscriminately updating all panels and widgets by only background and foreground colors.  For the "system" color mode, these are set by SYS_COLOUR_WINDOW and SYS_COLOUR_WINDOWTEXT, which is not what we want (dang; it almost worked).

The fix I have in mind is a delegating function that chooses color based on the type of widget/panel it's trying to draw to.  This involves creating colorscheme definitions similar to mainFrame's current modes, but for all widget/action types that SYS defines.  (e.g. SYS_COLOUR_HIGHLIGHT.)  I'm positive this won't work for some specific widgets, but I'll deal with those on a case-by-case.

Here is the color palette I want to map for Dark Mode: https://store.kde.org/p/1376146/.  This is close to what @blitzmann was trying to achieve in #1206 and is near enough to discord's colors.  Is this a colorscheme you want to use?

DarkFenX:
The color scheme looks good. Just FYI: i do not plan to use dark mode myself, so it has to be convenient for other users in 1st place, and I found concern in that comment pretty legit.

Tell me when you're done, I will check again, on all 3 platforms (not just on linux).

DarkFenX:
@mccloudy wxPython 4.1.0 is out. I think it is question of time until we switch to it. Upstream issue progress is going to be tracked here: #2179 just giving you heads up, because newer version might mean some older hacks do not work / some newer hacks are possible. So, might make more sense to test your code with it.

Also I am starting to do some work in `hidpi` branch. Given the nature of what you are doing and what I am doing, we're likely to run into conflicts during merges. If it turns out that I merge this branch to master earlier than we merge this PR, let me know if you run into any complex conflicts when updating your branch before the merge.

Metallicow:
Im a wrye bash dev, and they suffer from the same type of things, because they do not know wx/wxPy as well as they should. wxPy4.0.x will be a good standing point for a while. If not forever.... if having to use Py2. x64 py2 performance is so much better at the moment. I reallly look forward to gettting wxPy4.1 performance issues out of the way.

DarkFenX:
We're py3 already and not making this switch back. It's too painful to backport some py3 features there. As for wx - sure, 4.0.7post2 would be great version. But since 4.1.0 finally allows me to implement DPI-aware windows version of app more or less easily (thanks to `wx.Window.FromDIP`, and possibly ability to use hi-res assets and have hires matplotlib renders on GTK (if some issues are ironed out on wx side), I am going to switch to it without any doubts.

Badya:
Consider linking #1782 as duplicate

m-sasha:
Just mentioning that wxPython 4.1.0 uses the system's theme (so Dark Mode, if it's enabled), at least on Mac OS X. It looks good, too:
Screen Shot 2020-07-30 at 11 56 27
What I see that needs fixing is
- The slot colors aren't good for dark background.
- The custom tabs need a different color. 
- Some of the icons don't look good on a dark background.

DarkFenX:
@m-sasha this ticket is for "custom" dark mode, implemented on pyfa side. The thing you refer to is better to track here: https://github.com/pyfa-org/Pyfa/issues/2179

m-sasha:
Yeah, I understand that. Implementing a "custom" dark mode may not be necessary if the native one works well.

DarkFenX:
@m-sasha for mac and linux yes, for windows (which is our largest user base) there will be no native dark mode.

PR #2198, opened 2020-06-18
WIP: add translation for zh_CN

zhaoweny:
**This is a work in progress.**

So earlier this week I spent some time on researching GNU `gettext`, and then started to annotate and add zh_CN translation for current version of Pyfa.

Currently I just got it up and running, annotated some most frequently used panels. But the plan is that I can give it a start, then when I ( or anyone ) get interested in continuing the work, they can skip the setup phase.

### TODOs:

* [ ] annotate the whole GUI
* [ ] translate strings to zh_CN
* [ ] add option to switch between languages, ideally hot reload the translation
* [ ] *any ideas?*

### DIFF
--- a/gui/builtinStatsViews/resourcesViewFull.py
+++ b/gui/builtinStatsViews/resourcesViewFull.py
@@ -79,7 +79,7 @@ class ResourcesViewFull(StatsView):
         self.headerPanel.Layout()
 
     def getHeaderText(self, fit):
-        return "Resources"
+        return _("Resources")
 
     def getTextExtentW(self, text):
         width, height = self.parent.GetTextExtent(text)
@@ -105,8 +105,13 @@ class ResourcesViewFull(StatsView):
         base = sizerResources
         sizer.AddStretchSpacer()
         # Turrets & launcher hardslots display
-        tooltipText = {"turret": "Turret hardpoints", "launcher": "Launcher hardpoints", "drones": "Drones active",
-                       "fighter": "Fighter squadrons active", "calibration": "Calibration"}
+        tooltipText = {
+            "turret"     : _("Turret hardpoints"),
+            "launcher"   : _("Launcher hardpoints"),
+            "drones"     : _("Drones active"),
+            "fighter"    : _("Fighter squadrons active"),
+            "calibration": _("Calibration")
+        }
         for type_ in ("turret", "launcher", "drones", "fighter", "calibration"):
             box = wx.BoxSizer(wx.HORIZONTAL)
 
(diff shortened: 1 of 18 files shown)


DarkFenX:
Fyi `pyfa-org:hidpi` is wrong branch to merge to

zhaoweny:
Should I merge into `develop` or `master`? Which branch is appropriate for a PR?

Since I'm just getting started, a `git rebase` is trivial.

DarkFenX:
yeah, use `master`

gui/builtinStatsViews/resourcesViewFull.py
```diff
@@ -105,8 +105,13 @@ def populatePanel(self, contentPanel, headerPanel):
         base = sizerResources
         sizer.AddStretchSpacer()
         # Turrets & launcher hardslots display
-        tooltipText = {"turret": "Turret hardpoints", "launcher": "Launcher hardpoints", "drones": "Drones active",
-                       "fighter": "Fighter squadrons active", "calibration": "Calibration"}
+        tooltipText = {
+            "turret"     : _("Turret hardpoints"),
```
DarkFenX, line 109: No padding for vertical dicts please, here and below

DarkFenX:
I checked the code and it looks alright (besides one style remark, which i commented in the diff). Will check functionality a bit later. I suppose with this code it loads chinese by default? Before merging, have to expose it as an option.

> ideally hot reload the translation
I think it would be disproportionally complex, since pyfa calculates sizes of panels based on text size, and adding special event and support for it in all the panels to recalculate size would be quite painful.

zhaoweny:
> hot reload translation is hard

I guess a restart is also reasonable then.

> need a option to trigger loading Chinese translation

I'll look for options to implement this. My current understanding is that Pyfa stores it's options in binary format that can't be edited by text editor - which means we'll have to build a panel for language switching.

gui/builtinStatsViews/resourcesViewFull.py
zhaoweny: I added a commit for reformat the code.

DarkFenX:
Just use existing infrastructure. Check how various settings are implemented in `service/settings.py`

blitzmann:
Just a note: I started down the path of getting UI elements wrapped in `_()`. I have a branch out there for it, I'll find it and link it here, but you could use that to merge into your branch to provide extra annotations.

blitzmann:
JFC, apparently I was pushing to a private sandbox repo of mine.

Here it is:
https://github.com/pyfa-org/Pyfa/tree/i18n

I haven't worked on it in the past 2 weeks due to other obligations, but it has the basics in place. I don't remember which files still need to be annotated, but can check tonight when I get off works.

A brief look at the differences between this branch and mine:
* I'm using a parameter when starting pyfa to set the language, while you're using the users settings. I think either would be acceptable - the only thing I would note is I'm not sure if the settings are available at the time of database initialization (which we'll eventually need so that we can pull translated text properly from the database). Eventually I would like to have a popup on first load ask the user which language they want to use out of the supported languages, but that's something we can work on at the end.
* Instead of `builtins._ = wx.GetTranslation` I'm adding `_ = wx.GetTranslation` to the top of all the files that use it. This is mostly to avoid PyCharm from yelling at me. It looks like you solved this by introducing PyCharm settings to ignore? I haven't loaded it yet to see, but if that's the case either way would work methinks as long as that doesn't ignore other legit things. 

Also, can confirm that hot reload is really hard, if not impossible. Restart of the application would be fine.

zhaoweny:
@blitzmann 

> an `i18n` branch

that's great! a `i18n` branch is very much needed.

> using _ = wx.GetTranslation

it's a trade off. Using `builtin._` can avoid putting `_ = wx.GetTranslation` to every file that use it. But there might be some legit use of `_` as identifier and get errors at runtime.

> settings might not be avaliable

There are 2 type of translation system, python `gettext` and wxpython `wx.GetTranslation`, both supports GNU gettext translation system. `gettext` is builtin so that pyfa don't need to be worried about wxpython is not available. But the latter have more streamlined API to work with.

But my honest opinion is that pyfa should package the dependency or specify it via dependencies, and do not check "is wxpython installed" at runtime.

I think it's time to close this PR, then rebase it onto `i18n` branch or `master` branch and open a new draft PR.

PR #2200, opened 2020-06-19
WIP: add translation for zh_CN (continue)

zhaoweny:
**This is also work in progress.**

After examine, I choose to do a quick and dirty merge of `i18n`, then select `i18n` as PR target branch. Because it will allow me to merge this PR to `pyfa-org/Pyfa` repo quicker than me developing on my own branch. The goal is to setup basic things, laying down the ground work, and merge it into `i18n` and continue from there.

TODOs: ( see also #2198 )
* [ ] clean up `lang.po` and redaudent strings in `pyfa.po`
* [ ] clean up `_ = wx.GetTranslation` in favor of `builtin._ = wx.GetTranslation`
* [ ] *might there are something else for merge cleanup?*
* [ ] work out a way for accessing locale settings as soon as pyfa startup.
* [ ] GUI stuff to choose between translations? - personally I don't have much experience on this
* [ ] some script for automatically generate `*.mo` and packaging translations - I think pyfa should rely on shell script for this, because even Git for Windows has required `msgfmt`, and python's implementation `msgfmt.py` is hard to detect (e.g. a virtualenv might not have this)


blitzmann:
A continuation of the discussion form the previous PR:

I think there was some confusion about when I said "settings might not be avaliable"

>There are 2 type of translation system, python gettext and wxpython wx.GetTranslation, both supports GNU gettext translation system. gettext is builtin so that pyfa don't need to be worried about wxpython is not available. But the latter have more streamlined API to work with.

I get that there's `gettext` and `wx.GetTranslation`. For pyfa's purpose, we absolutely should use `wx.GetTranslation`  as that already provides some translations for some standard UI things (OK/Cancel buttons, etc), and the fact that it simply integrates better with wx 😊

What I meant by "settings might not be available" is that our user settings service (what you're using to store the language) might not be accessible at such an early time in the runtime where we would need it (before the database is loaded). I don't actually know without looking, and I have a hunch that it should be fine, but I know a lot of our services have some quirks and assumptions about what's available when they come online. I don't think it'll be a problem to figure out regardless

I'm looking currently into cleaning the branch up and reviewing whats here. I very much agree that the `i18n` branch should be the target for these additions until we're ready to merge it in

Thanks!

blitzmann:
Hrm, it also seems like something went wrong during the merge lol 

image

I'm going to go through your commits and cherry pick the commits that are applicable
 and apply to our branch, as I feel this is partly my fault for being dumb and not pushing my changes to the right repo as I was doing them >.< I'll do that and put up a PR for you to review as well

I'm also going to enforce `_ = wx.GetTranslation` declaration on each file that uses it. After thinking about it a bit today, I feel like this is more explicit, and doesn't rely on updating pycharm-specific settings that other folks might not use.

blitzmann:
@zhaoweny  please see #2201 for my attempt of getting this branch to a place where we can merge it in :) if you're good with those changes, we can merge that PR, close this one, and from there continue with the work that needs to be done :)

zhaoweny:
closing in favor of #2201 , which is sane to review and merge comparing to this PR.

PR #2202, opened 2020-06-20
i18n Localization

blitzmann:
This PR is to show and discuss current efforts to support basic localization in pyfa.

We will only support languages that the official EVE client supports (it would be weird to have Klingon in the UI but everything in the EVE database is English 😀)

Feel free to contribute to this branch after brushing up on the latest comments to avoid duplication of effort.

Couple of notes: 
* The translations themselves aren't a high priority due to the chance of source strings changing during development. This PR is focused on getting core support for translations in pyfa, and the translations can come later. That being said, we will of course accept PRs with translations, just know that they may need to be re-done as this codebase progresses.
* At this time we're not looking to get full localization support (for example, `1,234` vs `1.234`, date formats, etc). We're focused on GUI text translations. But again, if you know how to do proper localization for digits, etc, PRs are welcome!

TODOs:
* [x] Complete GUI annotations (`_t("")`) (~~also need to switch current `_()` to `_t()`, see https://github.com/pyfa-org/Pyfa/pull/2202#issuecomment-646968963 for details~~ DONE) 
* [x] Work out a way for accessing locale settings as soon as pyfa startup (needed for `eos` initialization)
* [x] Set `eos` language based on user's saved preference
* [x] Database translations. Need input from @DarkFenX on best way to do this since I've been in a hole for the past year and things have changed 😁
    * Phobos supports dumping translations with the data. We'll then need to load `eos` with the needed translation string and probably alias the translated properties in SQLAlchemy so that we automatically pull the correct column.
    * To expose the language to `eos`, we'll probably just write `eos.config.lang = <lang>` in `pyfa.py` once we get that info.
* [x] GUI stuff to choose between translations
* [ ] (maybe) First time start up to have the user choose language (and default to user's current language if supported)
* [x] Some script for automatically generate `*.mo` and packaging translations. Unsure what we can do here that's cross-platform and not environment-bound. Open to suggestions (@zhaoweny suggests "I think pyfa should rely on shell script for this, because even Git for Windows has required `msgfmt`, and python's implementation `msgfmt.py` is hard to detect (e.g. a virtualenv might not have this)")
* [x] Windows / Mac builds: run translations automatically for each build. I assume we would only want to maintain the uncompiled `.po`, and then have the build scripts deal with the compiled `.mo`. This means that folks running from source would have to compile them themselves if they want localization...
* [ ] Ensure helper scripts still work 

Bugs:
* [x] https://github.com/pyfa-org/Pyfa/pull/2202#issuecomment-647032997
* [x] https://github.com/pyfa-org/Pyfa/pull/2202#issuecomment-663879122


zhaoweny:
I got a exception when I try to set character skill level in character editor:
```
pyfa v2.22.1
EVE Data Version: 1751741 (2020-06-18 18:35:35)

OS version: Windows-10-10.0.19041-SP0
Python version: 3.8.0 (tags/v3.8.0:fa919fd, Oct 14 2019, 19:37:50) [MSC v.1916 64 bit (AMD64)]
wxPython version: 4.1.0 (wxWidgets 3.1.4)
SQLAlchemy version: 1.3.17
Logbook version: 1.5.3
Requests version: 2.23.0
Dateutil version: 2.8.1

####################

  File "D:\src\pyfa-org\Pyfa\gui\characterEditor.py", line 665, in changeLevel
    _setTreeSkillLevel(grand, skillID)
  File "D:\src\pyfa-org\Pyfa\gui\characterEditor.py", line 649, in _setTreeSkillLevel
    _("Level {}").format(int(lvl)) if not isinstance(lvl, str) else lvl)
TypeError: 'str' object is not callable
```
It seems `_ = wx.GetTranslation` was overrided. Maybe pyfa should use something else for i18n. How about `_t = wx.GetTranslation`?

During the translation process, There are situation that pyfa should mark a string literal for translation, and do the actual translation later (citing gui/patternEditor.py:184). Maybe pyfa can use `_r` for GetRaw?

blitzmann:
I'll have to check to see what you mean about _ getting overridden and using some GetRaw when I get to a computer. Its not immediately clear at the moment

blitzmann:
Ahh, I see the string literal error now. it's due to this:

image

We set the first return value to `_` since we don't need it (it's a throw-away). But, since `_` is defined globally for the file, this overwrites it for the entire file instead of just the scope.

Hrm... part of me wants to continue to use `_` for translations since that's the standard way, but another part of me dreads hunting down all the throw-away `_` variables and changing them to something like `__`.

I think your proposal of using `_t()` would be a good middle ground. :) We would want to change everything over to that before we continue with annotations, just so that we can get it out of the way :)

Still unsure about your proposal on `_r()`, researching...

zhaoweny:
@blitzmann for example, say we have these variable for i18n: `name`, `direction`

```py
btn.SetToolTip(_("%s patterns %s clipboard") % (name, direction))
btn.Bind(wx.EVT_BUTTON, getattr(self, "{}Patterns".format(name.lower())))
```

so line 1 both variable should be localized, but `name` can't be localized on line 2 or it will cause runtime error.

blitzmann:
Aha, I see what you're saying... Hrm... does wxPython have a `wx.GetRawString` function for translated text?

I'm wondering for those situations if it's better to just add a identifier for these. The way that it's done currently is simply used as a shortcut, but we can do something like

```
 importExport = ((_("Import"), wx.ART_FILE_OPEN, _("from"), "import"),
                        (_("Export"), wx.ART_FILE_SAVE_AS, _("to"), "export"))

        for name, art, direction, attr_name in importExport:
            ...
            btn.SetToolTip(_("%s patterns %s clipboard") % (name, direction))
            btn.Bind(wx.EVT_BUTTON, getattr(self, "{}Patterns".format(attr_name)))
```

I think this would be a cleaner solution than to try to manage getting a raw value (unless you think otherwise?)

blitzmann:
Or hell, we could just get rid of some of the cruft in the middle:
```
 importExport = ((_("Import patterns from clipboard"), wx.ART_FILE_OPEN, "import"),
                        (_("Export patterns to clipboard"), wx.ART_FILE_SAVE_AS, "export"))

        for text, art, attr_name in importExport:
            ...
            btn.SetToolTip(text)
            btn.Bind(wx.EVT_BUTTON, getattr(self, "{}Patterns".format(attr_name)))
```

zhaoweny:
The second one, with full tool tip as a whole string is better for translation. Translated strings would be more natural to read comparing using place holder everywhere :).

And no need for `wx.GetRawString` because it would be:
```py
def _r(x):
    return x
```

I agree it would be better if we can have all literal strings translated before passing around, though it means that source code will require a little more modification.

blitzmann:
@zhaoweny agreed! :)

blitzmann:
Noticed a bug:

image

`Not Learned` shouldn't be prefixed with `Level`. Will add it to OP as a list of bugs associated with this PR

blitzmann:
@zhaoweny I went ahead and converted all `_()` to `_t()` as discussed, lets move forward with that from here on out 😁

blitzmann:
Added quick and dirty ui for changing language at the bottom of the General settings:

image

zhaoweny:
Great! A basic UI is good for most players. I've annotated (almost) all strings in preferences, translations (including this one) are coming :).

zhaoweny:
> This means that folks running from source would have to compile them themselves if they want localization...

The main concern was that not all python runtime have `msgfmt.py`. There are 2 ways to compile `lang.po`:
* `msgfmt`: part of GNU gettext toolkit, available from Git for Windows, Msys2, Cygwin, Linux, macOS
* `msgfmt.py`: part of cpython source, available from [some](https://docs.python.org/3/library/gettext.html#internationalizing-your-programs-and-modules) python distros (**not** available when running from virtualenv on Windows)

If we want to make sure that localization is ready to use even running from source, I think we should put `pygettext.py` `msgfmt.py` (from [`Tools/i18n`](https://github.com/python/cpython/tree/master/Tools/i18n)) into `scripts`. Then we can call these scripts on startup.

blitzmann:
Yep, that sounds good to me. I'll add those scripts - gonna look into getting this included into the build system so that will be handy to have

blitzmann:
Got language compile working with windows builds 

image

Unsure about mac tho. I don't have one to test with, and the artifact is failing to upload. Will have to test this out a bit more later

> upload to transfer.sh failed

blitzmann:
@zhaoweny does the eve client even support Chinese language?

image

I'm not seeing it on my side. But then, I'm not sure if there's an EVE Chinese client that has it (considering the rules surrounding products released in China), and the client apparently has translations for it:
image

zhaoweny:
> does the eve client even support Chinese language?

Serenity has Chinese support. 

According someone more familiar with the backstory: Tranquility had Chinese support long long time ago; then it was disabled intentionally when Serenity was ready for Chinese players. Chinese players who choose to stay on Tranquility, (or migrated from Serenity) would patch the client to have options for Chinese translation.

If you need a EvE client with Chinese translation, you can download a runtime patch from [this](https://zhpatch2.evemodx.com/) site.

blitzmann:
Interesting, thanks for the info! I don't need to run the client with it, was just curious since I didn't see it there. 

I've  started work on the work that is needed to support EVE database translations. It runs but browsing fits results in exception. Basically got it to open the GUI and accepted that as a stopping point for tonight 😋

image

This is on a separate branch as I expect this to take a bit longer, and wanted the normal translation work to be easily merged into the base `i18n` branch in the meantime. It's been a bit messy, complicated by the fact that we don't have any standards regarding object properties dedicated to display vs lookup, and the fact that sql alchemy mappings are kind of a mess to begin with, etc. Also, of course, we've used English names whenever hardcoding things, so we're forced to scour the codebase to find where these things live and direct them to the raw English version instead of the translated synonym.

I'll provide more details and the "why's" when I create a draft PR, but want to get this a bit more stable before that. But it's getting there!

blitzmann:
So... here's a problem.

image

That's after loading all available translations for the types. That's super big, and not something we can swing for a normal, official release (or maybe it is, I dunno).

A few possible solutions
* Package two separate versions: one with translations, and one without. The non-translated one will simply have the English variant and will more or less function the same as pyfa does today. This will probably be the easiest to get off the ground. This would provide a smaller download for the vast majority of our English-speaking users, but also allow access to translated clients.
* Have one installation, but provide multiple translated databases available for download. This would also be pretty easy I think, but it would break the update flow (unless we have some sort of auto-download system, which could be neat). Each database could be suffixed the same way the translated strings are (eg `eve_zh.db`) and we can, on load, just point the connection string to that.
* Re-evaluate the way we're even doing translations. The reason why we need translations in the database is so that lookups works - searching for a Rifter in Chinese (`裂谷级`) would bring it back. Same for market items. However, we only technically need searchable strings in the database. We could theoretically generate .po/.mo files for all the other text (descriptions, display names, etc) and use the same `_t()` functions on them. However, this would require more research and thought into how we prep and build these things. I think the .mo files are much more compressible, and thus might allows us to ship with the entire language set without too much headache in file size (obviously haven't test this - that would be the first thing we would have to do is to test to see what benefits this would bring us)

Anyway, these are things that we can discuss and decide on at a later time. I'm going to continue with the work of at least getting things language-agnostic as it relates to the database. Good news is that I have fits coming back and calculating:


image

blitzmann:
Never mind on that database size stuff... at least for now. I ran the latest db_update and it's down to 71 MB, and the build files actually comes out only a few megs larger, so shouldn't have any issues with that :D

blitzmann:
@zhaoweny merged in the database branch into the main i18n branch. 👍 

I think we can complete any outstanding GUI annotations and deal with any bugs that pop up here.

zhaoweny:
Here is another bug. Open a Jackdaw fitting (e.g. [this](https://wiki.eveuniversity.org/Jackdaw/Fittings) one from EvE Uni), then right click the fitting:

```
  File "D:\src\pyfa-org\Pyfa\gui\builtinViews\fittingView.py", line 663, in spawnMenu
    menu = ContextMenu.getMenu(self, mainMod, selection, *contexts)
  File "D:\src\pyfa-org\Pyfa\gui\contextMenu.py", line 116, in getMenu
    sub = m._baseGetSubMenu(callingWindow, srcContext, mainItem, selection, rootMenu, it, rootItem)
  File "D:\src\pyfa-org\Pyfa\gui\contextMenu.py", line 265, in _baseGetSubMenu
    return self.getSubMenu(callingWindow, context, rootMenu, i, pitem)
  File "D:\src\pyfa-org\Pyfa\gui\builtinContextMenus\shipModeChange.py", line 50, in getSubMenu
    menuItem = self.addMode(rootMenu if msw else sub, mode)
  File "D:\src\pyfa-org\Pyfa\gui\builtinContextMenus\shipModeChange.py", line 35, in addMode
    label = mode.item.name.rsplit()[-2]
IndexError: list index out of range
```

`mode.item.name` is localized name of 'Jackdaw Defence Mode' (with zh_CN localization '寒鸦级防御模式'). So call `rsplit` on this won't work as intended.

I think this issue can be resolved by having a dictionary to map various mode to it's localized name, then match `typeName` for appropriate menu label.

```python
d = {
    'Defense' : _t('Defense'),
    'Propulsion': _t('Propulsion'),
    'Sharpshooter': _t('Sharpshooter')
}
```

blitzmann:
@zhaoweny nice catch! Yes, I think that would be the best way to fix that

I just pushed something that will probably affect you: I simplified the codes that are used to determine the locale setting, so instead of `zh_CN`, it's just `zh`. Tou may have to override it using the `-i` flag and then reset the settings.

I also updated the README - my goal is to have a one-stop shop of all localization information stored there, so that someone who wants to contribute can do so easily without knowing much about programming or how the program is architecture. Feedback is welcome ;)

zhaoweny:
I think the statsView have some dps calculation issue. This [fitting](https://zkillboard.com/kill/85291197/) with my totally not skilled in-game character has 492 dps.

Same fitting on pyfa v2.22.1 has 866 dps:
Screen Shot 2020-07-01 at 11 10 46 PM

Same fitting on latest pyfa i18n (English, macOS) has 456 dps:
Screen Shot 2020-07-01 at 11 12 24 PM

Same fitting on latest snapshot [artifact](https://ci.appveyor.com/project/pyfa-org/pyfa/builds/33847745/artifacts) i18n (Chinese, Windows) only has 321 dps:
Screen Shot 2020-07-01 at 11 12 48 PM

zhaoweny:
Screen Shot 2020-07-01 at 11 32 32 PM

(Continuing previous comment) I see that EvE data version isn't the same - but it should not be *that* far off, right?

zhaoweny:
(Continuing previous comment) Recently When I open a fit in pyfa, there are lots of error in the logs:
```
[2020-07-01 15:04:45.841673] ERROR: eos.gamedata: AttributeError generating handler: module 'eos.effects' has no attribute 'Effect12'
[2020-07-01 15:04:45.842706] ERROR: eos.gamedata: AttributeError generating handler: module 'eos.effects' has no attribute 'Effect9'
[2020-07-01 15:04:45.842871] ERROR: eos.gamedata: AttributeError generating handler: module 'eos.effects' has no attribute 'Effect40'
[2020-07-01 15:04:45.843934] ERROR: eos.gamedata: AttributeError generating handler: module 'eos.effects' has no attribute 'Effect1173'
[2020-07-01 15:04:45.844144] ERROR: eos.gamedata: AttributeError generating handler: module 'eos.effects' has no attribute 'Effect42'
[2020-07-01 15:04:45.844371] ERROR: eos.gamedata: AttributeError generating handler: module 'eos.effects' has no attribute 'Effect2663'
```
Is that related to issues in screenshot? (latest log attached as [pyfa-logs.zip](https://github.com/pyfa-org/Pyfa/files/4858861/pyfa-logs.zip))

blitzmann:
The eve data version is different and may account for some of these issue. Since I needed the translations, this branch tedchnically has a fresher datadump. There may have been patches to TQ that the new dump included that aren't in the latest official release, though there definitely does seem to be other things going on (at the very least, the english / chinese comparison should show the same stats).

It's almost certainly the fact that certain hull skills aren't being applied correctly for certain effects, since we key off the skill name in those effect expressions. We got past most calculation issues with category and group names due to being able to change the underlying schema, but the types would need a bit of manual intervention I think. I will take a look when I get home.

Thanks for the testing! this is definitely the stuff that we need to find before releasing it :D

blitzmann:
kk, fixed that issue, it wasn't just hull skills, but all skills that are filtered in the effects.

The `module 'eos.effects' has no attribute 'Effect9'` errors are very normal. This usually means that these effects aren't implemented in pyfa (there's a variety of effects in the game that we do not implement because they are either implemented differently or simply aren't needed for fit calculations)

zhaoweny:
Great :). By a quick comparison, I think English and Chinese localization have same result.

blitzmann:
@zhaoweny We had a sensitive info leak, so I had to rewrite / rebase a bit of this branch. Before any new development work, it's probably best if you just hard reset to what's in pyfa's `i18n` branch :) Let  me know if you need help with this, sorry for the troubles!

zhaoweny:
Got more info for stats calculation issue. This [fit](https://www.eveworkbench.com/fitting/gila/44d54842-a4fd-4311-f3f2-08d7519c4df4) from eveworkbench uses 474/475 tf CPU on eveworkbench and pyfa (master), but uses 523/475 tf CPU on pyfa (i18n).


fitting-stats-miscalculation

blitzmann:
thanks for the info! Good stuff there, will take a look tomorrow

zhaoweny:
I think I've tracked down and solved this issue. At [db_update.py:111](https://github.com/pyfa-org/Pyfa/blob/i18n/db_update.py#L111), it seems you made a typo and returned partial data. Changing it to `return compiled_data` seems solves resources miscalculation issue (and weapon dps miscalculation issue, item stats empty affected by issue, item stats empty requirements issue).

blitzmann:
Noticed an issue with the missile ammo picker, the icons for the different damage types aren't displaying

image

Will take a look at this today, and planning on getting a pre-release out tonightish

blitzmann:
@zhaoweny take a look at 95af966 - we were originally import MainFrame (which, for better or worse, imports basically everything else in the project) __before__ initializing the application with the locale. I've moved it after, and things like default param translations (`def func(thing=_t("Translation"))`) now work. I expect that a bunch of the things that we moved from out of classed to inside them to take advantage of the translations happening on class init instead of on file import would work now. Something to keep in mind - we may be able to revert some of those work arounds now if we desire.

zhaoweny:
At [pyfa_gauge.py:346](https://github.com/pyfa-org/Pyfa/blob/c46a59e3b1e110cbf9c58a22f023d926aa2a7f8d/gui/pyfa_gauge.py#L346), the string literal `"{0:.{0}f} over"` and `"{0:.{0}f} left"` is what I want to translate. Translation in Chinese requires reorder the string (e.g. `"over 1.23"` and `"left 1.23"`). So I think I have to make some workaround for dual-layered format string ...

zhaoweny:
> moved `import MainFrame` after initialize PyfaApp

That's great! But consider this line of code at [targetProfile.py:33](https://github.com/pyfa-org/Pyfa/blob/9ddfcc894fc0ad524d2de5383825a126d4a0a058/eos/saveddata/targetProfile.py#L33) and [damagePattern.py:29](https://github.com/pyfa-org/Pyfa/blob/9ddfcc894fc0ad524d2de5383825a126d4a0a058/eos/saveddata/damagePattern.py#L29). These builtin profiles seem like come in effect with `import eos.db` at [pyfa.py:129](https://github.com/pyfa-org/Pyfa/blob/76674435c9810adf3fa67c4ad1bca74250ca8bb4/pyfa.py#L129). I'm wondering could we initialize `PyfaApp` even earlier, right after reading configuration for i18n.

zhaoweny:
I'm started to work on the Graph panel. First thing I noticed is the graph from `matplotlib` can't display Chinese characters. A bit of research reveals it seems a `matplotlibrc` file with a bit of customization can solve this ([source](https://matplotlib.org/3.2.1/tutorials/introductory/customizing.html)). But what font should we use?

zhaoweny:
It seems [Poedit](https://github.com/vslavik/poedit) itself is using GNU gettext for translation. I think It's nice to have translations managed by Crowdin. A quick search for pyfa on Crowdin results that some people have already done some part of translation, but abandoned it long time ago :D.

Maybe not this time, but would pyfa consider to move to Crowdin for official translation?

blitzmann:
re: `matplotlib`. I have no opinion on that unfortunately. I haven't touched the graph code in literally years, and it's been vastly reworked by @DarkFenX recently. I remember us having an issue with font cache for matplotlib when we... updated it at some point? I don't recall, i feel like we had to do something special with the builds (ran fine locally, but the windows / mac builds didn't work). Anything that deals with matplotlib needs to be pretty carefully evaluated and we need to make sure it's included in the build process and works when updating the binaries, etc. But, if you can come up with a solution that works for you, I can take a look and give some feedback and try to validate it for builds. Just don't have the bandwidth at the moment to help you look into it :(

re: Crowdin. I'm not familiar with this service... does it just host .pot files? I'm not sure what advantage that gets us? (I've done all of 5 minutes reading on it, but maybe you have a better understanding of what it is and how we can utilize it)

zhaoweny:
> matplotlibrc

It's not complicated for a single language. For example, I can set `font.sans-serif : Microsoft YaHei, san-serif` to tell matplotlib to use `Microsoft YaHei` instead of fonts that do not support Chinese characters. It's difficult to go i18n and come up a combination of fonts that display all the characters correctly. *Can we get the font that wxPython is using?*


> crowdin

Just got myself signed up on crowdin.com. Comparing to manage translations directly at Github, crowdin has some advantage:

* Lots of translators have experience with crowdin, but not familiar with committing to github. ([Poedit](https://crowdin.com/project/poedit) and [Factorio](https://crowdin.com/project/factorio) are using crowdin)
* crowdin seems to have some kind of peer-review mechanic, to achieve higher quallity of translation. Offloading review pressure from devs to other translators.
* Crowdin seems to have open-source support. At the end of [pricing page](https://crowdin.com/pricing), there is a link to [Open Source Request](https://crowdin.com/page/open-source-project-setup-request). "Each granted license will include an unlimited number of projects, strings, and members."

blitzmann:
@zhaoweny I've created a Crowdin account and submitted a request for an open source license. I'll dig around it the next couple days to get a feel for it and how to best integrate it. 👍 

They have a Github integration, tho I'm not sure if this stuff needs to be in master first before anything useful happens with it. We shall see!

blitzmann:
As for the font, not sure. I think wxWidgets uses the systems font (I've seen linux screenshots where you can custom the font of the gui and pyfa respects thosse for the most part). I'm not sure if it would be easy to direct matplotlib to use the same font...?

Just a precaution, I would be wary of using `Microsoft`-anything simply because those might not be bundled with matplotlib for mac and linux :)

blitzmann:
`wx.SystemSettings.GetFont(wx.SYS_SYSTEM_FONT)` I think will get you the current system font used by wxPython... but glancing at the properties it looks more informational than anything, so not sure how to get that loaded into matplotlib

image

This may be helpful: https://stackoverflow.com/questions/16574898/how-to-load-ttf-file-in-matplotlib-using-mpl-rcparams

Unfortunately all of this is outside the scope of my understanding :(

blitzmann:
@zhaoweny I've finished integration with Crowdin (at least I think?). I see that you've joined the project there already - I modified your role to be a proofreader for zh, so you should be able to contribute as well as approve changes / contributions. :)

I'm still not 100% sure what the workflow is. I'm going to create a test account there and "translate" some phrases in some language using Google, and see what exactly happens. Want to get a strong idea down before we unleash this to the rest of the community to tackle 😁

zhaoweny:
Soooo, what do I have to do if I want to update the translation now? Do I need to annotate the strings, then (re)generate the POT file locally, or just annotate strings and wait for crowdin to generate POT file? Seems like locale/README.md needs some rewrite for crowdin-based workflow :D.

blitzmann:
Yep, gonna have to rewrite the readme, thats on my list of things to do this weekend.

From what I understand, under normal circumstances, we would generate new annotations and update the POT file, which will then become available for translating in Crowdin by translators. The designated proofreaders in crowdin would the approve the translations, and crowdin makes a PR for those new translations once approved to be included back into the branch.

Buuuut, for the situation of the translator doing the annotations as well, I assume you could annotate, generate POT, and submit .po translated files directly to the branch, and Crowdin should sync them when it gets merged in.

All that is slapped with a big ¯\_(ツ)_/¯ tho, this is ofc uncharted territory. Will take a look at your PR tomorrow and see what it does for crowdin :D

blitzmann:
So this is interesting. https://github.com/pyfa-org/Pyfa/pull/2228/commits/fa0720f963c073963f4710575b5f3b0c077777fc

I made this translation on an account with Translator role, and it still produced a PR without it being approved. So it seems that approval isn't necessary for it to generate a commit. I don't know whether this is good or bad TBH - allows anyone to generate commits which is good for getting it off the ground... but also allows anyone to generate commits lol

blitzmann:
@zhaoweny Also, seems that crowdin didn't pick up on the translations you did:

https://github.com/pyfa-org/Pyfa/pull/2235/commits/af76103957000e2b36e923b66d662fc88d288270

This is what I think happened:
1) I merged in the pot and po files from #2235
2) I synced the crowdin project with pyfa, which updated the pot file. I -did not- upload the new translations tho (this is a different option)
3) It updated the pot file on crowdin, which produced updated .po files for all languages, but they were blank since it didn't get the updated translations.

I'm going to try top upload the translations now and see what happens, will report back what I find :)

blitzmann:
Yep, uploading the translations fixed it both on crowdin and on the PR after syncing again... okay. So if translatins are introduced through git, we'll just have to be sure to upload them on the crowdin projects. Doesn't seems too bad.

But I think, going forward, lets just update the POT file in git and do everything else through Crowdin to avoid confusion. But this is good, I understand this stuff a bit better now :D

zhaoweny:
Clear. Glad we (you) got Crowdin workflow sorted out. 😃

blitzmann:
Prepping a dev release right now :)

blitzmann:
I'm going to make a tweak to how we deal with setting eos language. I found out today that Italian isn't actually supported by the game client (whoops), but we already have UI translations for it. So what I'm going to do is decouple pyfa translations and eos translations, which will give us the ability to select the independently. As part of this, if we have a language that doesn't have an associated eos language, by default it'll set the English version.

zhaoweny:
@blitzmann sounds good. 👍

blitzmann:
And there we go 
image

blitzmann:
I'm kind of wanting to devise a way to auto-generate the supported languages based on what's found in `locale`. That would allow us to simply turn on other languages in Crowdin and them magically appear in pyfa when merged

blitzmann:
Done with dynamic translation listings, and also made it show a readable name for it.

image

Think we're pretty close to just merging this in and getting it ready for an actual release

blitzmann:
Well, found an issue, super strange too.

If I select Turkish (new language from someone who requested it in Crowdin) and restart, I see russian things for things that aren't translated yet. example, firepower panel:

image

Russian:
```
msgid "Firepower"
msgstr "Огневая мощь"
```

Turkish:
```
msgid "Firepower"
msgstr ""
```

But things that have been converted to Turkish are applying correctly. For example, "Browse ESI fittings": 
image

Russian:
```
msgid "&Browse ESI Fittings"
msgstr "&Просмотр ESI Фитов"
```

Turkish:
```
msgid "&Browse ESI Fittings"
msgstr "&ESI Kurulumlarını Gezin"
```

It's almost as if it's defaulting to Russian for some reason instead of English for translations that aren't available. Haven't noticed this with any other language yet.

zhaoweny:
> Dynamic translation listing

That's awesome! It would be nice to have some kind of complete-ness indication (like crowdin's translated strings %), to help people choose between different translations. (This idea comes from Factorio; awesome and additive game, BTW.)

> it's defaulting to Russian for some reason instead of English for translations that aren't available

I researched a bit and It seems we don't really have control over what language it's fallback to. About fallback to English thing, the only English translation `en_US` currently only have a little of translation. If you turn on translation tracing `wx.Log.AddTraceMask('i18n')` you can see that there are lots of logs about strings are missing translation :smile:.

I've pushed some experiments to `i18n-staging` branch. There is a form of `gettext` and `wx.GetTranslation`, that accepts a `context` parameter to disambiguate strings with different context. We can abandon the not-so-complete English translation if we adopt this way of specifying string context. Plural string format also needs some tinkering when calling `xgettext` to generate `lang.pot`.

blitzmann:
> It would be nice to have some kind of complete-ness indication

Yep, already looked at the APIs, we could theoretically hit the API and write this data to a file at build time that the preference panel can ingest. Planning on looking at that tonight.

About the English strings, the fallback should be the source string. That's the way I've always understood it. So if the english translation doesn't have anything, it should be fine. Also, removing the translation outright will probably screw with the new dynamic listing, something to keep in mind.

I thought I looked into the context stuff, but it was determined that the wx version we have doesn't support it?  I might be misremembering, but will take a look at your branch to see what you found out. I'll also try to reproduce the Turkish / Russian issue in a small app and ask about it on wxPython forums, would be nice to know what's going on there

blitzmann:
Have a question out tot the wxPython forum:

https://discuss.wxpython.org/t/wx-locale-missing-turkish-translation-is-using-the-russian-translation-instead-of-english-as-default/34800

If no response there I'll log a bug on their github

blitzmann:
Got a quick and dirty progress indication

[image](https://user-images.githubusercontent.com/3904767/88466741-7d900800-ce9d-11ea-8414-7f5d9f705f0f.png)

AppVeyor isn't building correctly tho, still need to fiddle with it. Also need to be aware that these numbers can be inflated if the translations there haven't been merged in yet

blitzmann:
@zhaoweny would you be able to explain how you used `wx.Log.AddTraceMask('i18n')` to get log messages? I've never used the wx logging classes, and I'm having trouble getting this stood up to show log messages

zhaoweny:
I don't have much experience either. All I did was that appended `wx.Log.AddTraceMask('i18n')` in `PyfaApp.OnInit()`.

I was using macOS for these translation trace messages. I can't reproduce it on windows. I think you might have to use a debug build of wxwidget to produce the log AND a debugger or 3rd-party tool to read these messages according to [this](https://wxpython.org/Phoenix/docs/html/log_classes_overview.html#logging-overview) doc.

blitzmann:
Yep, I must have skipped over that part. Doing this following does indeed produce output:

```
wx.Log.AddTraceMask('i18n')
wx.Log.SetVerbose(True)
```

But I cannot get the trace-level details. I guess I'll be looking into building wx myself to get access to these. At least with the verbose logging, I'm able to see that en_US is laoded form our own message catalog, but any other language is loaded from an internal wx one. I assume this is wx loading the catalog for built-in translations (fine), but I don't see it loading `lang` when set to russian, chinese, etc

```
[2020-07-27 04:52:25.231965] DEBUG: gui.app: Setting language to: en_US
12:52:25 AM: using catalog 'lang' from 'C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\locale\en_US\LC_MESSAGES\lang.mo'.
```

```
[2020-07-27 04:55:44.255720] DEBUG: gui.app: Setting language to: ru_RU
0:55:44: using catalog 'wxstd' from 'C:\Users\holme\Documents\Sync\Git\blitzmann\Pyfa\venv\lib\site-packages\wx\locale\ru\LC_MESSAGES\wxstd.mo'.
```

blitzmann:
I'm going to have to open a bug report on wxPython's repo, the Discuss site isn't generating any replies. I'm not sure if we can launch this feature with random languages popping up. This can't be the only project that has run into this issue either. :/

zhaoweny:
I think it's fine. English version of pyfa is still there, the translation is opt-in, there is a indicator of how complete the translation is. They will switch back to English if translation not satisfy their need. 😄

zhaoweny:
Should we initialize all existing translation with English text? There is a tool called `msgen` for that.

blitzmann:
> If the string is not found in any of the loaded message catalogs (see Internationalization ), the original string is returned.

I noticed that as well in my search, but I find it hard to believe that's what most programs would tolerate. If you set language to Russian, and it doesn't find it, I would think the vast majority of applications would want to see the source string, and not some other random language. So that leads me to think that there may be some technique that others have done to avoid that issue (like loading a very specific catalog and then ignoring the rest, but then I don't know how we would be able to dynamically generate the list of available).

As for pre-loading them with English, that's absolutely something that we can probably do. I'm not exactly sure how Crowdin would work with that tho. I guess if we use it on the `.pot` file, that would default all the `msgstr` to the English version (assuming). Crowdin might get confused those and think those are translations, thus the % complete wouldn't be correct. That's something we could try tho, and if that's the case then we can switch to showing percentage approved instead and consider "approved" strings to be translated

As for releasing with this issue, I'm just generally iffy with the idea. I get that it's opt-in, but I feel like if folks opt in and the translation isn't there, they would expect to see English and that should be fine with them- they understand it's incomplete. But if they start seeing their language along with 3 or 4 other random languages, they may think it's completely broken and not bother with it in the future. It's a perception thing.

I saw we try the `msgen` route and go from there to see if the issue persists and how Crowdin deals with it

zhaoweny:
How about make it a step in the build? We don't have to commit these not-translated-at-all translations into pyfa's code base. 🤔

blitzmann:
Well that would make too much sense! 😋

Good idea though, that would solve everything I think. I'll do some research tonight and try to get something worked out for the builds. 👍

blitzmann:
I have it working with appveyor builds, works pretty well 😁 Now I've got to get it to work with the travis builds.

Not really sure what to do about linux. This is using gettext tools directly, not something that's been converted to python. So linux folks will have to make sure to download gettext and run msgen against the po files, and then run the mo conversion

blitzmann:
kk, everything should be fairly situated now. Last thing I have to do is ensure all our helper scripts still function correctly and poke DarkFenX for a quick rundown on the changes. But I think we're in a good spot now that we've got that empty translation thing figured out,

zhaoweny:
@blitzmann @DarkFenX is there anything I can help to get this PR merged?

blitzmann:
I poked @DarkFenX last week to take a look as I think it's ready. As this PR changes how static files are generated and he's the one that's been doing updates, I would like for him to sign off on it. I actually haven't checked slack since then (and for some reason it just simply refuses to auto start when I login to the desktop 😣), and I've been busy with other projects recently.

I'll check it tonight and try to keep it moving. I'm hoping we can get it merged in this weekend, and maybe a new pre-release out as well :)

DarkFenX:
FYI i got past huge workload I had at work. Will check this PR and how localization works in general after today's pyfa releases.

zhaoweny:
@DarkFenX great 😄 . Happy to answer any question you might have :).

blitzmann:
@DarkFenX don't remember if I ever gave you this, but to dump the data I've been running:

`dump_data.py --eve E:\EVE --json ..\staticdata -p C:\path\to\Phobos`

Should help in checking it out. It imports phobos files and uses a custom class to split the filess, avoiding the filesize limitations on Github.

Other than that, most of the codebase changes are simply GUI annotations. I added some logic that switches which columns are used in the database based on the language selected that you may be more interested in.

DarkFenX:
FYI yesterday I've looked into changes to phobos and adjusted them, so that:

- by default, no grouping is used
- maps are sorted before splitting them into groups to reduce diffs between staticdata updates

I grabbed EVE client data dump today using --group=10000, all files are sub-50M. So, that's all for the phobos part, now I need to go through pyfa part - see what the changes are in db generation and UI are, then we should be able to make a release.

zhaoweny:
Ping 😄 .

I understand there might be something else that need to be taken care of other than Pyfa. If there is anything I can do to help this PR go through, please let me know.

DarkFenX:
I am looking into it now. Want to combine sisi release with localization as another beta release

DarkFenX:
FYI i am in process of merging i18n and singularity branches. There were quite a few conflicts, I am yet to resolve `gui/builtinContextMenus/envEffectAdd.py`. Also, some message IDs changed (the text in code), some were added during the merge, I am yet to figure out what else has to be changed to accomodate to that. Hopefully will make a release tomorrow, if everything works alright.

DarkFenX:
I merged it manually. It works at 1st glance, but i will thoroughly test it today and tomorrow, and add latest sisi updates as well. Noticed a few quirks which look strange.

PR #2203, opened 2020-06-20
I18n: more zh_CN translations and a better locale/README.md

zhaoweny:
Though currently add more translation is not the priority, it's a good way to find out which part of GUI still need annotate.

Currently zh_CN translation is sitting at 292/314 (92%), with these components still need more work:

* Preferences window - completely not translated
* Graph window - completely not translated
* Fit table high/med/low/rig slot names
* Various column names

Edit: of course there is a grammar error.

### DIFF
--- a/gui/characterEditor.py
+++ b/gui/characterEditor.py
@@ -340,7 +340,7 @@ class SkillTreeView(wx.Panel):
         self.skillBookDirtyImageId = self.imageList.Add(wx.Icon(BitmapLoader.getBitmap("skill_small_red", "gui")))
 
         tree.AppendColumn(_("Skill"))
-        tree.AppendColumn(_("Level"), align=wx.ALIGN_CENTER)
+        tree.AppendColumn(_("Level"))
         # tree.SetMainColumn(0)
 
         self.root = tree.GetRootItem()
--- a/gui/chrome_tabs.py
+++ b/gui/chrome_tabs.py
@@ -24,7 +24,7 @@ from gui.bitmap_loader import BitmapLoader
 from gui.utils import color as color_utils, draw, fonts
 from service.fit import Fit
 
-
+_ = wx.GetTranslation
 _PageChanging, EVT_NOTEBOOK_PAGE_CHANGING = wx.lib.newevent.NewEvent()
 _PageChanged, EVT_NOTEBOOK_PAGE_CHANGED = wx.lib.newevent.NewEvent()
 _PageAdding, EVT_NOTEBOOK_PAGE_ADDING = wx.lib.newevent.NewEvent()
@@ -227,7 +227,7 @@ class ChromeNotebook(wx.Panel):
         self.page_container.Layout()
 
         self._pages.append(win)
-        self.tabs_container.AddTab(title, image, closeable)
+        self.tabs_container.AddTab(_(title), image, closeable)
 
         self._active_page = win
         self.ShowActive(True)
--- a/gui/chrome_tabs.py
+++ b/gui/chrome_tabs.py
@@ -215,7 +215,8 @@ class ChromeNotebook(wx.Panel):
 
         wx.PostEvent(self, PageChanged(current_page, new_page))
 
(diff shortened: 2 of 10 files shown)


blitzmann:
checking over this now, good point on the translations making it easy to spot 😁

gui/characterEditor.py
```diff
@@ -340,7 +340,7 @@ def __init__(self, parent):
         self.skillBookDirtyImageId = self.imageList.Add(wx.Icon(BitmapLoader.getBitmap("skill_small_red", "gui")))
 
         tree.AppendColumn(_("Skill"))
-        tree.AppendColumn(_("Level"), align=wx.ALIGN_CENTER)
+        tree.AppendColumn(_("Level"))
         # tree.SetMainColumn(0)
```
blitzmann, line 344: curious why we got rid of the align?

gui/chrome_tabs.py
```diff
@@ -227,7 +227,7 @@ def AddPage(self, win=None, title="Empty Tab", image: wx.Image=None, closeable=T
         self.page_container.Layout()
 
         self._pages.append(win)
-        self.tabs_container.AddTab(title, image, closeable)
```
blitzmann, line 230: I don't think we need to translate variables like this. I feel like we should pre-translate anything thats passed into `AddPage()` and simply display it here. I'm not sure if it's possible to even do this since gettext won't know what this variable would be. Thoughts?

gui/multiSwitch.py
```diff
@@ -37,7 +37,7 @@ def handleDrag(self, type, info):
             if h:
                 h(type, info)
 
-    def AddPage(self, tabWnd=None, tabTitle=_("Empty Tab"), tabImage=None):
```
blitzmann, line 40: Same as above, I think we need to translate the literal strings like this which are then passed into the other function as-is.

I'm wondering if you ran into a problem that required you to change it? I have done very little testing with the stuff that I've commited, so it's possible I'm not seeing the reasoning :)

gui/characterEditor.py
zhaoweny: I think before this change "Level" is not aligned like "Skill":

before:
Screen Shot 2020-06-21 at 12 37 58 AM

after:
Screen Shot 2020-06-21 at 12 37 29 AM

gui/multiSwitch.py
zhaoweny: Yeah, It's a problem regarding default tab "Empty Tab". I have that literal string translated but the UI still showing the raw string.

Now I think about this, when does `_(...)` called? I think it's when we call `import gui.multiSwitch`, and at that time, we might not have locale correctly set up. and when the `gui.mainFrame` calls this with default args (see also `gui/mainFrame.py:270`), I assume it would bind to the string literal "Empty Tab" before translation could happen.

If you get rid of `_ = wx.GetTranslation` and use this snippet:
```py
def _(*args, **kwargs):
    return wx.GetTranslation(*args, **kwargs)
```
Then add a break point on the `return` line, you will see that it's a series of `import` calls, all the way back to `pyfa.py:140`.

blitzmann: I just tested on a small test app, and you're right - when used as a default argument, the translation isn't done at runtime.

For now I think a decent workaround would be to set the default to `None` and then default it in the function itself:

```
def AddPage(self, tabWnd=None, tabTitle=None, tabImage=None):
    tabTitle = tabTitle or _("Empty Tab")
```

I'm going to make some inquiries on the wx forum to see if there's a better / standard way of dealing with this

blitzmann: https://discuss.wxpython.org/t/translations-dont-happen-when-set-as-default-argument/34719 
https://github.com/wxWidgets/Phoenix/issues/1693

I have these two feelers out, it occurred to me ass I was making the issue that I've never attempted a function call as a default argument in python before, so it might just not be a thing.

But I'm good with using things like `tabTitle = tabTitle or _("Empty Tab")` as a workaround

blitzmann:
Made a couple tweaks per the discussions above, merged. 👍

PR #2204, opened 2020-06-22
i18n: annotate Preferences, update zh_CN translations

zhaoweny:
This time the major changes are about Preference. The zh_CN translation now have ~~454/455 (99%)~~ 556/565 (98%) strings translated.

Components that still not translated:

- graph window
- context menus
- some stat view tool tips
- column names
- Fit table high/med/low/rig slot names

Edit: update stat with context menus (mostly) translated.

### DIFF
--- a/gui/builtinItemStatsViews/itemAttributes.py
+++ b/gui/builtinItemStatsViews/itemAttributes.py
@@ -115,7 +115,7 @@ class ItemParams(wx.Panel):
 
         with wx.FileDialog(
                 self, _t("Save CSV file"), "", exportFileName,
-                _t("CSV files (*.csv)|*.csv"), wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT
+                _t("CSV files") + " (*.csv)|*.csv", wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT
         ) as dlg:
 
             if dlg.ShowModal() == wx.ID_CANCEL:
--- a/gui/builtinItemStatsViews/itemAttributes.py
+++ b/gui/builtinItemStatsViews/itemAttributes.py
@@ -11,6 +11,8 @@ from enum import IntEnum
 from gui.builtinItemStatsViews.attributeGrouping import *
 from service.const import GuiAttrGroup
 
+_t = wx.GetTranslation
+
 
 class AttributeView(IntEnum):
     NORMAL = 1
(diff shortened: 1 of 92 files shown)


gui/builtinItemStatsViews/itemAttributes.py
```diff
@@ -111,8 +114,8 @@ def ExportItemStats(self, event):
         exportFileName = self.item.name + " (" + str(self.item.ID) + ").csv"
 
         with wx.FileDialog(
-            self, "Save CSV file", "", exportFileName,
-            "CSV files (*.csv)|*.csv", wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT
+                self, _t("Save CSV file"), "", exportFileName,
+                _t("CSV files (*.csv)|*.csv"), wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT
```
blitzmann, line 118: For these guys, I adopted a different convention in other places, simply because the filetype shouldn't change with the translations: 

` _t("CSV files") + " (*.csv)|*.csv")`

If there's no objections with that, let switch to it to keep consistent 😊

gui/builtinItemStatsViews/itemProperties.py
```diff
@@ -95,5 +95,5 @@ def PopulateList(self):
 
         self.paramList.SortItems(lambda id1, id2: (idNameMap[id1] > idNameMap[id2]) - (idNameMap[id1] < idNameMap[id2]))
         self.paramList.RefreshRows()
-        self.totalAttrsLabel.SetLabel("%d attributes. " % idCount)
+        self.totalAttrsLabel.SetLabel(_t("%d attribute.", "%d attributes.", idCount) % idCount)
```
blitzmann, line 98: Oh wow, I did not realize wx.GetTranslation had another argument for plural strings. Nice find! I'm sure there's other places in the codebase where using this would be helpful

gui/builtinPreferenceViews/pyfaHTMLExportPreferences.py
```diff
@@ -39,24 +41,24 @@ def populatePanel(self, panel):
...
-                                             wx.DefaultPosition, wx.DefaultSize,
-                                             URL='file:///{}'.format(self.HTMLExportSettings.getPath()),)
+                                                               wx.DefaultPosition, wx.DefaultSize,
+                                                               URL='file:///{}'.format(self.HTMLExportSettings.getPath()), )
         mainSizer.Add(self.PathLinkCtrl, 0, wx.ALL | wx.EXPAND, 5)
 
-        self.fileSelectDialog = wx.FileDialog(None, "Save Fitting As...",
-                                              wildcard="EVE IGB HTML fitting file (*.html)|*.html", style=wx.FD_SAVE)
+        self.fileSelectDialog = wx.FileDialog(None, _t("Save Fitting As..."),
+                                              wildcard=_t("EVE IGB HTML fitting file (*.html)|*.html"), style=wx.FD_SAVE)
```
blitzmann, line 49: Same thing as the first comment. Will probably just want to find all `wildcard=` and update them 👍

gui/builtinItemStatsViews/itemAttributes.py
zhaoweny: I'm on it. I've come up a regex for that: `['"].+\|\*\.\w+['"]`. I think `|*.xxx` part is required part for this wildcard.

zhaoweny:
Now with more translation, this time mainly focused on context menus.

zhaoweny:
Some additional note: the PO file is edited with Poedit, which reorders entry based on settings. If someone edit the PO file sort-by-file, and another person edit the PO file sort-by-strings. It will cause huge diffs and merge conflicts.

Currently the solution is use `msgcat`, which is part of GNU gettext tookkit. Suppose we want final PO file sort-by-file, we can use `msgcat -o lang.po --sort-by-file [INPUTFILE]` to sort and merge existing PO files (specified as `[INPUTFILE]`).

@blitzmann should I run `msgcat -o lang.po --sort-by-file lang.po` before committing the PO translation file? I think it is better to have relatively fixed order of `msgid`s.

blitzmann:
Great work!

I didn't get a chance to take a closer look last night, but will try to do it tonight. Lets hold off on any more commits to this branch until I can complete the review, it's getting big! (feel free to continue with the progress, but it would be easier to manage on my end if you make a secondary branch and then PR that one once this one goes through 😁)

zhaoweny:
Will do. A staging branch is a good idea :).

PR #2208, opened 2020-06-26
i18n Localization - Database

blitzmann:
This is a sister PR to #2202 - whereas #2202 focuses on GUI translations and annotating, this PR is focused on EVE database translations. It will eventually be merged into that PR tho

EVE data already includes these translations and it's possible to generate them via Phobos which is wonderful. However, there's plenty of gotchas surrounding translated texts from EVE. I'll attempt to go over the various changes here and what still needs to be done / decided upon.

Screenshot:
image

### Changes:
* ~~Stores files in `staticdata` in LFS. This is something new for this repo, and isn't something I've worked with in the past. I'm not sure if other contributors need to install the LFS extension for this or not. This was required due to the size of some of the files (we could technically split them if we added support in Phobos, but meh)~~ 
* Stores staticdata files as split. This is to avoid the GitHub limit of 100MB per file, which happens with `evetypes.json` when you do translations.
* EVE database generator and schema changes
    * So, lots of changes here of course. I'm going to try to describe the flow as best I can, but keep in mind that the current schema is old as fuck and has been through several iterations over the years, that there wasn't a one-size-fits-all solution.
    * I tried my best to have the English text target the original column if possible
    * I took advantage of the `synonym()` definitions that we already had in place to dynamically determine which translated text to target
    * There's some columns that I had to rename altogether (tbh, this was before I found the field map functionality in db_update, so maybe these weren't needed /shrug)
* There's many places in the source where we reference a hardcoded value with the assumption that it's English. I've gone through and changed these to target the literal raw English property. Likewise, there's plenty of places that weren't using the synonym to display but instead using the base column - I've changed these as well.

### TODO:
* [x] Go through codebase and try to identify any more instances of hard coding to the incorrect property that need to be address
* [x] Fix the broken preference saving
* [ ] Test things like item / ship searching (make need a flag for the query to switch on `typeName` (always english) vs `name` ("display name") to handle the hardcoded type searches)
* [x] erm... export / import. No idea how this stuff should work 

### BUGS:
* [x] `db_update` fails when pyfa startup runs it and the locale is not English. This is due to the fact that `db_udpate` needs to be run  within an English context


blitzmann:
Travis build is erroring for some reason on the db_update... I wonder if they don't have LFS support.

Appveyor seems to be building, haven't checked the build yet tho

zhaoweny:
Here is the [doc](https://docs.travis-ci.com/user/customizing-the-build#git-lfs) for using git-lfs on macOS. It seems we have to do a `brew install git-lfs` before using it.

zhaoweny:
I got the following error when running `git lfs pull` on macOS.
```
batch response: This repository is over its data quota. Account responsible for LFS bandwidth should purchase more data packs to restore access.                
error: failed to fetch some objects from 'https://github.com/pyfa-org/Pyfa.git/info/lfs'
```
After checking out [this](https://help.github.com/en/github/managing-large-files/about-storage-and-bandwidth-usage) doc, it seems that Git-LFS quota is rather small considering files been put into LFS.

blitzmann:
Hrm, yeah looks like there's some cost associated with LFS. Funny, the page it directed me to when it rejected my original commit didn't mention that.

image

We'll have to come up with a better system them, couple options:
* Update Phobos to dump data in split files (`evetypes.01.json`, `evetypes.02.json`, etc) and then we can concat them in db_update.
* Update Phobos to allow a callback to determine which things to dump (more granular than it's current miners). Phobos has changed a decent amount since I last looked at it, so I'm unsure how feasible either of these two options are. 
* Perhaps it might be possible to dump the translated text into a separate file? Tho, I feel like that's just offloading the problem from evetypes to some other file. We could also have one file per translated language. So `evetypes_en-us.json`, `evetypes_zh.json`, etc. This would unfortunately mean that a lot of the data between the files is redundant, but would result in similar-sized files throughout.
* Alternatively, we can zip them / gz / bzip to reduce the file size, but... I think the main reason for including the static data in the repo was to move away from saving the `eve.db` binary in favor of text-based solutions, so that git commits were smaller in size (git works well diffing text, but you can't diff a binary so it has to store the entire new version)

@DarkFenX would really like your input on this. I feel like the file-per-language would be the nicest output, and then we can fix up the db_update to basically recreate the original files using the various other translations, but I have no idea on the effort of that from Phobo's side.

blitzmann:
I've got Phobos splitting the JSON files - will make a PR to that and let @DarkFenX advise. Whether or not we go this route long term, it'll at least get us moving on this PR in the meantime

image

blitzmann:
@zhaoweny I've updated this branch to remove LFS and update the files tot he split versions. Can you try to pull now and run?

zhaoweny:
I got a error when switched from `i18n` to `i18n_db`, then start pyfa with old `eve.db` or no (deleted) `eve.db`. I ran `python db_update.py` after deleting `eve.db` and it seems working.

( log attached as [hang.log](https://github.com/pyfa-org/Pyfa/files/4841598/hang.log) )

blitzmann:
> I got a error when switched from i18n to i18n_db, then start pyfa with old eve.db or no (deleted) eve.db. I ran python db_update.py after deleting eve.db and it seems working.

Yeah, if you have zh_CN as your saved language and the db_update happens within pyfas startup, it's going to break. db_update assumes it's being run within the context of English due to the way the SQLAlchemy synonyms are set up. It's something I've got to fix before merging, hopefully it's easy enough

blitzmann:
>Edit: here is another error exception-key_error.log when click to open this specific fit Nyx - Nyx fit.xml.txt. This fit was using a Strong Blue Pill Booster, and it will load with en_US locale, and throw exception with zh_CN locale.

I'm getting something slightly different. I can set that booster on the en, but when loading it in zh there is no booster listed. If I try to add another booster we get database integrity issues. It seems that the booster itself is removed from the database, but the side effects for it are kept around. Bizarre, will look into it

blitzmann:
Pushed a fix for the booster thing, and made #2212 when we get around to it (extreme edge case and shouldn't ever really happen outside of us breaking the code / someone screwing with their database)

blitzmann:
Actually, the `group.name` thing is more pervasive... I'll take a look at it tomorrow - I may go the route that I went with category and keep `name` the English version, and create like a `displayName` to show it. Group really is only used for filtering specific types of modules, it isn't used in the UI much

zhaoweny:
If you set locale to zh_CN, open a fit and right click at 'Projected' (环境效果), you can get this exception:

```
  File "D:\src\pyfa-org\Pyfa\gui\builtinAdditionPanes\projectedView.py", line 332, in spawnMenu
    menu = ContextMenu.getMenu(self, mainItem, selection, *contexts)
  File "D:\src\pyfa-org\Pyfa\gui\contextMenu.py", line 116, in getMenu
    sub = m._baseGetSubMenu(callingWindow, srcContext, mainItem, selection, rootMenu, it, rootItem)
  File "D:\src\pyfa-org\Pyfa\gui\contextMenu.py", line 265, in _baseGetSubMenu
    return self.getSubMenu(callingWindow, context, rootMenu, i, pitem)
  File "D:\src\pyfa-org\Pyfa\gui\builtinContextMenus\envEffectAdd.py", line 73, in getSubMenu
    data = self.getData()
  File "D:\src\pyfa-org\Pyfa\gui\builtinContextMenus\envEffectAdd.py", line 111, in getData
    data.groups['Abyssal Weather'] = self.getAbyssalWeather()
  File "D:\src\pyfa-org\Pyfa\gui\builtinContextMenus\envEffectAdd.py", line 193, in getAbyssalWeather
    key = name_parts[1].strip()
IndexError: list index out of range
```

It's because Chinese does not use space to separate words, and you can't really tell which word start at where. So this line of code is broken with zh_CN locale ([envEffectAdd.py:191](https://github.com/pyfa-org/Pyfa/blob/i18n_db/gui/builtinContextMenus/envEffectAdd.py#L191)) :
```
                # Localized effects, currently, have a name like "(size) (type) Cloud"
                # Until this inevitably changes, do a simple split
                name_parts = beacon.name.split(" ")
```

blitzmann:
Thanks! Fixed in the latest commit, however the WH menu is kind of a mess and half broken for the most part, mostly because we rely so much on manipulating the names and trying to organize it in a way that it's not organize in the EVE data.

I'll be working on that a bit today to hopefully get it in a better spot

blitzmann:
@zhaoweny does this tree look correct?

image

I swear I'm going to be multilingual after all this is done :D

These are interesting issues because the environmental effect tree is based solely off of the name of the item. So we have to get translations for part of the word as well as the "junk" strings ("system effects" and "effects")

Haven't looked at the other context menus for environmental effects, but they are broken too.

blitzmann:
I've pushed up my updates for the environmental effects context menus, @zhaoweny if you can give it a look and see if it looks proper :) I've done the "translations" for the item name substrings already to get the menu working properly, but there's still custom strings that would need to be translated for that context menu to be fully done :)

blitzmann:
A note on eft / other format exports...

I think we'll keep these english-only. It looks like EVE exports to clipboard and English-only eft format, and we would also want to continue compatibility on all other applications and services that use these exports

zhaoweny:
(PR is coming.)

There is a translation that's slightly not correct：（小）“型生物荧光”, it can be literally translated back as (Sm)"all Bioluminescence". :D

There is a issue about 'Effect', because this `msgid` has already used in [itemStats.py:192](https://github.com/pyfa-org/Pyfa/blob/i18n_db/gui/itemStats.py#L192) and we can't really have same `msgid` for two translations. 

* 'Effect' that's describing item effect, translated literally as 'Effect' （效果）
* 'Effect' that's describing WH effect, translated literally as 'Effect Light Beam' （效果光束）

So  I think the best way to deal with this is have a English translation, then we can have full control of what `msgid` to use, avoiding collisions. The second best way is that compromise a bit and leave it translated as 效果, but this method has consequences that context menu is not as clean as the best method, and maybe other languages would need other kind of work around that this method can't support.

Screenshot for leave 'Effect' translated as 效果, with part of it literally translated back to English:
context-menu

blitzmann:
@zhaoweny so basically, the translation string is dependent on the context. "Effects" doesn't translate tot he same text in all circumstances....

So you propose having the code annotation like `_t("Effect_environmental_context_menu")` and then have explicit English translations for it ass well as other languages to match the string properly? This would allow us to have different translations for the same English text depending on the context.

This make total sense to me if I'm thinking of it correctly, and allows us more flexibility. I can make that change when I get home tonight

zhaoweny:
Your understanding is totally correct. I would like to describe this issue more precisely.

The original issue is caused by string manipulation. And manipulated string is super hard for translate. In this case, The official Chinese translation effectively added more infomation compared to original English text. 

According to extracted offical translation, `Effect Beacon` is translated as `效果光束`. Original English text before substitute is `Class 1 Effects`,  official Chinese translation is `一类效果光束`, which can be translated back to English as `Class 1 Effect Beacon` (*according to offical translation*). When substitute `Effect` (效果) from the translation, the result in Chinese is `一类光束`, which can be translated back to English as `Class 1 Beacon`.

English | Chinese | Chinese translated back to English
----- | ----- | -----
`Effect` | `效果` | - 
`Effect Beacon` | `效果光束`(official) | -
`Class 1 Effects` | `一类效果光束`(official) | `Class 1 Effect Beacon`
`Class 1` (substitute `Effect`) | `一类光束` | `Class 1 Beacon`



So the conclusion:

* For this specific issue, an approach could be get rid of string manipulation, and name every level of context menu explicitly.
* Use a English translation is generalized way to handle translation mismatch while keep using string manipulation, but might not work when translated string should be rewrited instead for more natural meaning.
* Current Chinese after merge #2213 is ok in my mind.

blitzmann:
Thanks for that write up :) Very helpful in understanding this stuff.

I think the string manipulation will have to stay, at least for now. Even if the manipulated strings aren't exactly natural, I think they are good enough or at least convey the intention. We could look into other tricks that will keep the auto-generation or the menu, but allow the translation to substitute the manipulated string with something more natural.

Also, Apparently GetTranslation has the ability for a context parameter:
https://wxpython.org/Phoenix/docs/html/wx.functions.html#wx.GetTranslation

I haven't looked into it, will play with it tonight, but something to keep in mind.

blitzmann:
https://github.com/wxWidgets/Phoenix/issues/1704

Apparently `context=` isn't a thing at the moment unless I'm doing it wrong. Oh well. I've prefixed those strings for manipulations, so you should be able to have different translations for those vs other "Effect" strings.

I also updated the zh po file using the merge command with `--update` - this showed a bunch of changes, looks like different ordering. Are you using the same command? 

Anyway, I'm going to focus tonight on making sure the db_update is language agnostic, work on a dump script to get the data from phobos, and get export/import tested/working. Will probably merge this in tonight or tomorrow

PR #2213, opened 2020-06-29
i18n_db: sort and (slightly) update zh_CN translation

zhaoweny:
I think it's best practice to follow official translation, so sorting `msgid`s won't really affect these translations.

On a side note: i18n_db is getting big, please consider to merge it early and solve issues in separate commit / PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ matrix:
     osx_image: xcode11.3
     language: generic
 before_install:
-  - git lfs pull
   - bash scripts/setup-osx.sh
 install:
   - python3 scripts/compile_lang.py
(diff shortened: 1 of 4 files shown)


zhaoweny:
I found that ship browser is using `name` but not `displayName`, so I included a small fix for that :).

blitzmann:
Thanks!

On the sorting, I was attempting to keep them together since they aren't really translations, but rather what strings we need to match on for proper functionality of the context menu. This affects how the code operates, and improper values there can actually cause things to break.

However, with your suggestion of IDing these specific things in a different way, we can call them `_environmental_context_Effects` or something with the same prefix so they automatically group up. As I mentioned in the other PR, I'll get this done today, unless you want to get a PR out for it :D

As for merging the db branch, it's definitely getting more stable. I want to test / fix a few more things first (export / import being the big one), but i'll probably merge it in the next couple days :)

PR #2214, opened 2020-06-29
add market source ceve-market

copyliu:


blitzmann:
Thanks! Will take a look tonight, but the code seems reasonable.

We've noticed an influx of Chinese interest in pyfa in the last few months (to the point where we're currently developing localizations). Did something happen with Serenity that has generated more playtime?

copyliu:
because Serenity was finally re-opened and synced up to TQ's version

blitzmann:
Oh, I had no idea that serenity shut down for a time. Thanks for the info, was just curious :)

DarkFenX:
FYI i split market sources in 2 groups and added code to process those groups, to avoid cross-fetches, in 63632e09b333e05047f0682858282f537cb9a5cb

PR #2217, opened 2020-06-30
i18n: sort zh_CN translation, more zh_CN translation

zhaoweny:
It seems that the default output from GNU gettext is sorted by file, or not sorted at all. Currently I'm using `-s` option for various GNU gettext tools to sort lang.po by msgid.

### DIFF
--- a/gui/builtinAdditionPanes/commandView.py
+++ b/gui/builtinAdditionPanes/commandView.py
@@ -160,7 +160,7 @@ class CommandView(d.Display):
             self.fits.sort(key=self.fitSort)
             stuff.extend(self.fits)
         if not stuff:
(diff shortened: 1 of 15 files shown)


blitzmann:
The only thing I'm iffy with is moving the options from their respective files to the copySelectDialog, although I know why it was done...

Perhaps we should keep the options in the different export files, but remove their descriptions, and have a dictionary set up on the copy select window that maps them to their descriptions. That way we remove the GUI aspect from the service... but then if there was ever an options added we would need to remember to add them in both places... :/

For now I'll merge this in - it's just an odd quirk of how we're doing things now, but might revisit this at a later date

PR #2219, opened 2020-07-01
i18n: more translation for zh_CN

zhaoweny:
In current stage, I think Chinese translation is somewhat stable. Instead having everyone running from source, could we have alpha release of localized pyfa, and let people try it out?

### DIFF
--- a/gui/builtinContextMenus/shipModeChange.py
+++ b/gui/builtinContextMenus/shipModeChange.py
@@ -37,11 +37,8 @@ class ChangeShipTacticalMode(ContextMenuUnconditional):
         return _t("Tactical Mode")
 
     def addMode(self, menu, mode):
-        label = None
-        for raw, locaized in self.modeMap.items():
-            if raw in mode.item.typeName:
-                label = locaized
-                break
+        key = mode.item.typeName.rsplit()[-2]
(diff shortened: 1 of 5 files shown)


blitzmann:
> could we have alpha release of localized pyfa, and let people try it out?

Yep! I'm planning on a dev release this weekend, that was one of the main points of getting more info into the README for people to reference and help out with the other languages.

gui/builtinContextMenus/shipModeChange.py
```diff
@@ -32,7 +37,12 @@ def getText(self, callingWindow, itmContext):
         return _t("Tactical Mode")
 
     def addMode(self, menu, mode):
-        label = mode.item.name.rsplit()[-2]
+        label = None
+        for raw, locaized in self.modeMap.items():
+            if raw in mode.item.typeName:
+                label = locaized
+                break
+
```
blitzmann, line 45: I think this can be simplified using the original code and accessing the dictionary using it:

```
key = mode.item.typeName.rsplit()[-2]
label = self.modeMap[key]
```

This would avoid a double loop :)

zhaoweny: Did not think that way. Let me change it real quick.

PR #2220, opened 2020-07-02
i18n: return compiled_data in _readData

zhaoweny:
As described in [this comment](https://github.com/pyfa-org/Pyfa/pull/2202#issuecomment-652857272), this PR fixes lots of issues caused by returning partial data.


blitzmann:
oh wow, nice catch!

PR #2222, opened 2020-07-08
i18n: more annotation, more translation

zhaoweny:


pyfa.py
```diff
@@ -125,6 +125,10 @@ def _process_args(self, largs, rargs, values):
         if db_needs_update() is True:
             update_db()
 
+        from gui.app import PyfaApp
+
+        pyfa = PyfaApp(False)
```
blitzmann, line 130: @zhaoweny curious on if there was a reason for moving this? In general, the order in which things are imported and initialized in this file are pretty important, so I just want to know so that I have an idea why

blitzmann: Disregard, just saw your other comment on the other PR.

So... I didn't realize those were in eos. We have a hard requirement not to include wx in the eos module itself for various reasons, so we shouldn't be doing translations in them... Not really sure where we would put them then, let me give it some thought this weekend and I'll let you know :)

zhaoweny:
@blitzmann I'm totally understand your concerns. Here are some random thoughts:

* I included some control part in string literal ( e.g. `_t('[NPC]')`). Would you like me to change it to `'['+_t('NPC')+']'`?
* does `from wx import GetTranslation as _t` satisfies the requirement of not importing wx in eos?

blitzmann:
1) I mean technically `'['+_t('NPC')+']'` would be more proper because the square brackets are identifiers for menu creation code. But that being said... that would get super ugly super fast. I'm okay with leaving them as-is - the translators will just need to be aware that pyfa's code is looking for `[]` to create submenus
2) Unfortunately not. There are a few projects out there that utilize the `eos` module from pyfa's repo for calculations, without actually using other bits of pyfa (including ui). So, for that reason, we can't use anything wx-related in the `eos` module, otherwise those projects will have to get the wx package, and have to set up an app, which might not work right if it's on a headless server, etc. So to avoid any hassle there, I'd rather just move the translations out of that file. 

I have an idea for dealing with those in `eos`, testing now. Will update PR for your consideration when done :)

zhaoweny:
How about migrating `[]` for sub-menus to just using `|`? That way we can get rid of all `[]` in string literals.

blitzmann:
Yeah, I was thinking about that as well, but to do that we would have to: 

1) Create migration script to update users database with the change (and there may be users that use `|` in their own, which would cause it to break)
2) Not absolutely required, but would require a conversion thing to convert `[]` to `|` when importing. There's a bunch of resources online that have various saved patterns that people then import - if they aren't converted to a new format and we don't do it, it'll fail.

#2 isn't a huge deal, just a potential annoyance to users. #1 would be an absolute must. I feel like it's kind of outside of the scope of this effort for right now, but could be something we tackle in the future.

BTW, I can't seem to push to you branch. Do you have "Allow edits from maintainers" enabled for this PR? If not, if you can enable it, that would make it easier, otherwise I can make a PR to your repo's branch for my proposal
https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork

image

blitzmann:
I sent you a PR with my suggestions: https://github.com/zhaoweny/Pyfa/pull/1

PR code itself fixes the translations being in eos and instead moves them out, but keeping the definition of the built in profiles in eos

PR description also goes into a possible solution to the whole `[]` situation. Let me know what you think :)

zhaoweny:
I do have "allow edits by maintainers" checked - not sure what happened there.

screenshot-allow-edits

zhaoweny:
with a bit of regexp, now damage pattern and target profile are using deferred translation 😄 .

zhaoweny:
I pushed a small fix that I want include into this PR, which fixes that the Poedit thinks '%' is part of escape sequence (see https://github.com/vslavik/poedit/issues/645). I'll push no more improvements into this branch, but only small fixes on request ;).

blitzmann:
Excellent, taking a look at changes now

PR #2232, opened 2020-07-17
i18n: string annotation for graph panels

zhaoweny:
The last major piece of GUI translation is Graph window  (I think). The translation is ready, but the graph panel produced by `matplotlib` still has font issue (can't display Chinese characters). Which is annoying, but the issue does not block string annotation and translation.

I'm not sure about the workflow with Crowdin, so I kept `lang.pot` and Chinese translation in a staging branch `i18n-staging`  😄 .

### DIFF
--- a/graphs/data/fitCapacitor/graph.py
+++ b/graphs/data/fitCapacitor/graph.py
@@ -18,31 +18,34 @@
 # =============================================================================
 
 
-from graphs.data.base import FitGraph, XDef, YDef, Input, InputCheckbox
+import wx
+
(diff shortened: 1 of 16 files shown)


blitzmann:
@zhaoweny lets go ahead and get the pot files and translations into this PR, then we can merge :) We'll have top figure out the font stuff sometime, could be an iterative bug fix after this initial set goes live

zhaoweny:
Then I'll push these changes into i18n branch :)

PR #2240, opened 2020-07-25
I18n: more string annotations (mostly tooltips)

zhaoweny:
This time I went for tool tip strings :). These changes also include a named placeholder for Russian translation.

### DIFF
--- a/eos/saveddata/fit.py
+++ b/eos/saveddata/fit.py
@@ -42,8 +42,11 @@ from eos.saveddata.targetProfile import TargetProfile
 from eos.utils.float import floatUnerr
 from eos.utils.stats import DmgTypes, RRTypes
 
-
 pyfalog = Logger(__name__)
+
+
+def _t(x):
+    return x
 
 
 class FitLite:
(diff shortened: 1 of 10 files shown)


zhaoweny:
I'm currently using this command for `xgettext`:
```
find * -name \*.py | xgettext --from-code=UTF-8 -o locale/lang.pot -d lang -k_t -f - -s
```
It's different from what we have in `locale/README.md`. Since we now have annotation in `graph/` and `eos/`, we can't really keep using the old `find gui/ *.py -name \*.py` command.

blitzmann:
I think I moved over to explicitly stating the directories because it kept getting things in my venv (which is in my pyfa directory). But I'll update the readme with whatever it needs to be :)

.travis.yml
```diff
@@ -12,7 +12,7 @@ matrix:
 before_install:
   - bash scripts/setup-osx.sh
 install:
-  - python3 scripts/compile_lang.py
```
blitzmann, line 15: Was there a reason for this change? I'm good with it, but curious what's the purpose?

zhaoweny: I rebased my commits on top of `scripts/compile_lang.py` `import config`. I think I should re-rebase my commits then :).

blitzmann: Hrm... that might actually be what I needed to get the config to work in the script.will test it out tonight

zhaoweny:
I'm updating `locale/README.md`, to include plural spec and updated `find` command. It will also get us ready for translation with context when we update wxpython to 4.1.0+.

zhaoweny:
I'm also using a in-tree python virtualenv. It's located at `.env` and when I use bash to run `find * -name "*.py"`, it's automatically excluded :).

Append `-not -path "./path/to/venv/*"` to `find` command should exclude the folder `./path/to/venv` recursively.

zhaoweny:
@blitzmann Hey, is there a way to re-run checks? I think It's failed because crowdin translation progress script :(.

blitzmann:
To rerun check, you would have to either commit to the branch again, or I can rerun them manually. I'll be taking a look at this PR today and wrapping it up for merging, sorry for the delay!

blitzmann:
Merging, going to figure out the build issues on the main `i18n` branch 👍

PR #2248, opened 2020-07-30
Do not sleep for half a second in OpenFitsThread.run()

m-sasha:
From the comment on that line, it looks like it was a workaround for some bug.
It seems to work fine without it on Mac OS X. Would appreciate testing on Linux/Windows.


blitzmann:
That was added way back in 2014, before the 2.0 release. It's very possible that 2.0 (and thus the updated wxPython / python) made that obsolete. 👍

PR #2256, opened 2020-08-16
Re-render tab when its disabled state changes.

m-sasha:


m-sasha:
This fixes the tabs at the bottom not redrawing their icon as disabled/enabled when switching e.g. between an empty-fit and an actual fit.
Before my performance fix which prevented excessive work in `_TabRenderer.SetSize`, `SetSize` would trigger the redraw. Now that `SetSize` ignores the call if the size is unchanged, we need to trigger a redraw explicitly when the disabled state changes.

PR #2260, opened 2020-08-26
Fix HiDPI scaling issues in VectorPicker

YaarPodshipnik:
VectorPicker was DPI unware, so when asking GetClientSize() it would get the actual size multiplied by the scaling factor of GetContentScaleFactor(). This made the widget seem to be cut off and display only the upper left
quarter of the circle.
Similarly, when choosing the font size for the percentages, it wouldn't scale the maximum size, resulting in very large text.


YaarPodshipnik:
@DarkFenX could you review?

This is how it looked without patch on my OpenSuse Tumbleweed with KDE on a Dell XPS13 with 4k display and 200% scaling
image

This is how it looks with patch applied:
image

DarkFenX:
Some elements seem to be too bold now. I have 200% scaling on my laptop, and this is how it looks on current master w/o your fix, on wxpython 4.0.7:

image

No fix, wxpython 4.1.0 (which indeed confirms the bug you were talking about):

image

With fix, 4.0.7:

image

With fix, 4.1.0:

image

DarkFenX:
Well, i think your result is somewhat better, so I am going to merge it regardless, and try fixing lines later

YaarPodshipnik:
Thanks! I guess there's some code that chooses line weight that depends on size, and it would need a similar fix. Will take a look again at this bit later.

DarkFenX:
Changed point size in 5181e8e95ae42f067a2a122e71ea517c857315e7

There is no way to fix the rest, though, since brush defines fill of elements and pen defines outline. Wide border is with pen width 1. If it goes a little below (even 0.99), it interprets border width as 0 and just disappears. I suggest to leave it as-is.

PR #2264, opened 2020-09-05
Delete all of a character's fits via ESI

jrdn:


jrdn:
Not sure if this is something anyone else would want but I had a personal usecase I hacked it together for and figured I'd at least put it up here just in case.

This adds a button to the browse ESI fittings window to remove all of a character's fits via ESI. I mostly use this since I like to keep all my personal fits on one character so I can export them to a file and import them on other characters so they're in sync, but then when I want to re-import them to update they'd all get duplicated.

DarkFenX:
I think it has more use-cases than just this i think. Looks cool, merged.

DarkFenX:
Hmmm, I ran into some issues with it:
```
pyfa v2.27.0
EVE Data Version: 1800064 (2020-09-09 08:42:42)

OS version: Linux-5.7.0-3-amd64-x86_64-with-glibc2.29
Python version: 3.8.5 (default, Aug  2 2020, 15:09:07) 
[GCC 10.2.0]
wxPython version: 4.0.6 (wxWidgets 3.0.5)
SQLAlchemy version: 1.3.16
Logbook version: 1.5.3
Requests version: 2.23.0
Dateutil version: 2.8.1

####################

  File "/home/dfx/Workspace/eve/pyfa/gui/esiFittings.py", line 188, in deleteAllFittings
    sEsi.delFitting(activeChar, fit['fitting_id'])
  File "/home/dfx/Workspace/eve/pyfa/service/esi.py", line 100, in delFitting
    super().delFitting(char, fittingID)
  File "/home/dfx/Workspace/eve/pyfa/service/esiAccess.py", line 114, in delFitting
    return self.delete(char, EsiEndpoints.CHAR_DEL_FIT.value, character_id=char.characterID, fitting_id=fittingID)
  File "/home/dfx/Workspace/eve/pyfa/service/esiAccess.py", line 272, in delete
    return self._after_request(self._session.delete("{}{}".format(self.esi_url, endpoint)))
  File "/home/dfx/Workspace/eve/pyfa/service/esiAccess.py", line 251, in _after_request
    raise APIException(
service.esiAccess.APIException: HTTP Error 520: FittingsStopSpamming, details: {"remainingTime": 325950595}
```

DarkFenX:
Fixed it in 3673e7f39e1fc417456190e96d87498924692571, now it deletes as much as it can

blitzmann:
> HTTP Error 520: FittingsStopSpamming, details: {"remainingTime": 325950595}

Interesting, I wonder what rate of HTTP calls CCP considers as spam? And what the hell is that remaining time? Surely not milliseconds.

Might have to ping someone at CCP to find out what an acceptable rate is. My guess is that 1 / sec would be fine. That would definitely increase the time it takes tho and while it works the UI will be blocked unless we thread this out, but at the same time aren't in game fittings limited to something like 200 fittings? So assuming 1/sec is acceptable to the CCP gods, 200ish seconds to complete isn't too bad especially since this is somewhat of a niche feature.

blitzmann:
Oh actually:

https://test-eveonline-third-party-documentation.readthedocs.io/en/latest/crest/character/char_fittings.html

Gee, would be nice if they put this on the legit ESI swagger docs

So we could simply catch a FittingsStopSpamming error (unsure if CCP uses status code 520 for all their custom errors or if it's specific to this), and pause for a while before continuing.

DarkFenX:
I already handled it, see the commit. It just stops deleting and prints error message into status bar.

blitzmann:
Huh, I must have just overlooked that comment. Indeed, that way also works :D

jrdn:
Cool stuff, thanks a lot all!

Issue #2270, opened 2020-09-21
Tech 1 weapons show incorrect attributes when switching from tech 2 weapon and charge using variations menu


Candyguitar:
## Bug Report
If a tech 2 weapon is loaded with tech 2 charges, switching to a tech 1 variation causes it to show the base attributes instead of modified. This is very noticeable on attack battlecruisers with their role bonuses for large weapons.

### Expected behavior:
Using the variations menu, switching to a tech 1 weapon from a tech 2 version should always show the attributes modified by character skills and ship bonuses.

### Actual behavior:
If one of the tech 2 weapons had tech 2 charges, the last tech 1 weapon will use base attributes instead of modified. This results in incorrect calculations for CPU, power grid, range, etc. The rest (if applicable) are calculated correctly. Closing and reopening the fit won't fix it, but changing it in any way will even after it is undone; for example, fitting another module and then pressing CTRL-Z.

### Detailed steps to reproduce:

1. Start with any ship fitted with at least one tech 2/faction/storyline/officer weapon and tech 2 charge.
2. Right click on the weapon or list of weapons and select a tech 1 variation.

### Fits involved in EFT format (Edit > To Clipboard > EFT):

```
[Oracle, Weapon attribute bug test]

[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]

[Empty Med slot]
[Empty Med slot]
[Empty Med slot]

Polarized Mega Pulse Laser, Conflagration L
Polarized Mega Pulse Laser, Conflagration L
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]

[Empty Rig slot]
[Empty Rig slot]
[Empty Rig slot]

```
```
[Tornado, Weapon attribute bug test]

[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]

[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]

1200mm Artillery Cannon II, Quake L
1200mm Artillery Cannon II, Quake L
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]

[Empty Rig slot]
[Empty Rig slot]
[Empty Rig slot]
```
```
[Kestrel, Weapon attribute bug test]

[...]

PR #2271, opened 2020-09-21
Fixed issue #2270 where t1 weapons would use base attrs after switching from t2

Candyguitar:
Fixes [issue 2270](https://github.com/pyfa-org/Pyfa/issues/2270), where tech 1 weapons would use base attributes after switching from tech 2 weapons with tech 2 charges using the variations menu.


DarkFenX:
See comment to the related issue

Issue #2071, opened 2019-10-30
Add price column to market list view


DarkFenX:
As requested in https://forums.eveonline.com/t/pyfa-the-python-fitting-assistant/27156/255

![image](https://user-images.githubusercontent.com/275209/67863519-1a9f5980-fb35-11e9-89a0-0e685a84b294.png)

PR #2277, opened 2020-10-02
add prices to market items

esunder:
#2071 Adds a price column to items in the market view

### DIFF
--- a/gui/builtinMarketBrowser/itemView.py
+++ b/gui/builtinMarketBrowser/itemView.py
@@ -20,7 +20,8 @@ class ItemView(Display):
     DEFAULT_COLS = ["Base Icon",
                     "Base Name",
                     "attr:power,,,True",
-                    "attr:cpu,,,True"]
+                    "attr:cpu,,,True",
+                    "Price"]
 
     def __init__(self, parent, marketBrowser):
         Display.__init__(self, parent, style=wx.LC_SINGLE_SEL)
--- a/gui/builtinViewColumns/price.py
+++ b/gui/builtinViewColumns/price.py
@@ -24,6 +24,7 @@ from eos.saveddata.cargo import Cargo
 from eos.saveddata.drone import Drone
 from eos.saveddata.fighter import Fighter
 from eos.saveddata.module import Module
+from eos.gamedata import Item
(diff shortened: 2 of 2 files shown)


esunder:
Note: I am trying to be a good steward of the code and add some tests. However, I am having some trouble with imports. There are no tests for other columns, like price, so it is tough to mimic an existing example.

I think it is a pathing issue, but it is able to import price.py just fine. Any ideas?

ImportError while importing test module '/home/sunder/Pyfa/tests/test_modules/test_gui/test_builtinViewColumnPrice.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.6/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_modules/test_gui/test_builtinViewColumnPrice.py:8: in <module>
    from gui.builtinViewColumns.price import Price
gui/builtinViewColumns/price.py:23: in <module>
    from eos.saveddata.cargo import Cargo
eos/saveddata/cargo.py:25: in <module>
    import eos.db
eos/db/__init__.py:120: in <module>
    from eos.db.saveddata import booster, cargo, character, damagePattern, databaseRepair, drone, fighter, fit, implant, implantSet, \
eos/db/saveddata/cargo.py:25: in <module>
    from eos.saveddata.cargo import Cargo
E   ImportError: cannot import name 'Cargo'

DarkFenX:
Hey, regarding your PR - sorry for not replying earlier. Adding price column is easy, and I considered it in the past. The reason why it wasn't added is not to overload price sources with updates. Any kind of search or active browsing will generate requests for thousands of items per minute. While caching policy on client side is decent, I think there has to be more stuff in place to add the price column to default distribution.

1) Batch request of multiple prices for all shown items (except for cached entries). This is not trivial thing to do, but it would benefit fitting panel as well.
2) Some kind of extra delay after showing items before price fetching requests are sent, like 1s at least - to avoid requesting stuff during searches, but request them only for final results

And those complications are the reasons why it is not in pyfa yet. It is cool end-user feature, but i for sure would not like to have pyfa user-agent banned from price services.

Issue #1225, opened 2017-06-09
Blurry look when upscaled by Windows


bethouxa:
## Bug Report
Pyfa looks blurry when scaled up by the OS (on Windows). Especially visible on high DPI displays.

### Actual behavior:
Pyfa looks like this when upscaled : [image](https://puu.sh/weR3E/6c951fbdf1.png)

### Detailed steps to reproduce:
In the Windows control panel, go to "Display", and increase the value of the "Change the size of text, apps..." slider. 

### Release or development git branch? Please note the release version or commit hash:
Release branch : commit  #461f521

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
Windows 10

blitzmann:
Yeah... unfortunately pyfa is one of those apps that don't play nicely with scaling. This is a limitation of the underlying toolkit, I believe. There's quite a few apps that still look funky when changing the scaling for the desktop (PyCharm, for example, which is the IDE pyfa is developed with, has blurriness), and pyfa is one of them. There may be a few things that can be done, but a quick Google search returns limited information. Further research into this is required. :)

Ebag333:
Unfortunately wxPython was never designed with DPI scaling in mind.  DPI scaling just increases the ratio for the app, so if your app is setup so that it has the best resolution at a 1:1 ratio, then it'll look terrible at a 1:2 ratio.

Modern apps will use extremely high resolution fonts and images even at low resolutions, so you can scale it without losing any quality.

Not much to be done there, since we don't control the font or images.

If your issue is the same as Bart in #1183 then might be able to work around the DPI issue......

VioletGiraffe:
> Modern apps will use extremely high resolution fonts and images even at low resolutions
I don't think that's an accurate nor true description. The fonts, for example, are rendered to the required size. A GUI framework that is DPI-aware will just know that the required size is x1.75 for 175% scaling, it will render a properly sized bitmap that can be mapped to the screen pixel-perfect. A framework that's non-DPI-aware will render at x1 and then the OS will scale up this bitmap, producing blurriness.

blitzmann:
wxWidgets (supposedly) has better support for hidpi in windows in 3.1. Unfortunaltey, wxPython is built with 3.0. I've been pushing Robin here and there in getting a 3.1 version built so that we can try it out. HiDPI is already available in OS x and linux builds, but linux support iirc is still iffy.

[...]

PR #2292, opened 2020-11-25
Add High DPI Support for Windows

rcmurphy:
This makes Pyfa much more readable on 4K displays, and shouldn't cause any issues on non-windows machines.

### DIFF
--- a/gui/app.py
+++ b/gui/app.py
@@ -36,6 +36,22 @@ class PyfaApp(wx.App):
         # Set language stuff and update to last used language.
         self.UpdateLanguage(config.language)
 
+        try:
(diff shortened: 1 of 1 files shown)


DarkFenX:
Last time i checked:
https://github.com/pyfa-org/Pyfa/issues/1225#issuecomment-620053190
It took much more work than just setting DPI awareness flag. It just disables scaling of window render by GPU, which leads to window contents being too small for hi-dpi screens, and various disproportions, since some elements are scaled using one type of point, others scaled another way.

Either way, tomorrow I will take a look at how it looks on my 3840x2160 display on windows

PR #2334, opened 2021-04-22
Update lang.po

Barmaley7:


DarkFenX:
These updates are supposed to come from https://crowdin.com/project/pyfa/ rather than being done here in the repo. I will transfer both of those to crowdin, though.

PR #2375, opened 2021-11-11
Enforce a strict 4.0.6 version for wxPython

blitzmann:
Have had a couple issues pop up lately with people installing the wrong version of wxPython when running from source. This should prevent those issues


DarkFenX:
`dep_facelift` branch which will hit master soon makes this obsolete

PR #2403, opened 2022-02-08
Probes x8 instead of x1000

shyadow:
This PR changes multi-add option for items in the "Scan Probes" market group to be x8 instead of x1000, since you usually only want 8 or 16 when adding probes to cargo.
image

Let me know what you think and if anything should be changed :)


DarkFenX:
thanks

Issue #2402, opened 2022-02-03
ERROR: The JWT token has expired: Signature has expired. FIX


videxerion:
### Expected behavior:
Login to ESI.

### Actual behavior:
JWT token error, The JWT token has expired: Signature has expired

### Detailed steps to reproduce:
1) Open 'PYFA'
2) Open 'Character'
3) Open 'Manage ESI Characters'
4) Click 'Add Character'
5) Enter ESI and get base64 string (Manual method)
6) Get local server startup error

### Release or development git branch? Please note the release version or commit hash:
The version was obtained via 'git clone'.

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
5.16.2-arch1-1

### Other relevant information:
When checking the token, the program considers it 'obsolete', after googling a little, I found out that the 'exp' parameter in the token itself is responsible for this. The check takes place in the 'esiAccess.py' file, 'services' folder.
![изображение](https://user-images.githubusercontent.com/63922155/152305515-6ec5cd56-24fc-41be-8e9a-25d727e36746.png)

After a little searching, I realized that you use the 'python-jose' library to work with jwt tokens, by opening its code and finding the decode method, I found the '_validate_exp' method in which the token is checked for 'validity'.

![изображение](https://user-images.githubusercontent.com/63922155/152305888-6b8265b4-3f65-4bfd-900c-4fbc5b1c13af.png)

As it turned out, the library developers use the current time zone of my computer (+3:00) and the time specified in the JWT token is in the zone (+0:00), which is why the token is immediately considered 'not valid'.
![изображение](https://user-images.githubusercontent.com/63922155/152306300-5b00c803-3d7d-4d78-a611-5e4be82380e3.png)
 
A quick fix error on the user's side can be corrected by changing the clock, in my case it worked. On the part of the developer, you can refuse to check the token or rewrite everything to a new library for working with tokens, which allows you to specify the time zone when checking the 'exp' parameter. 'Python-jose' and 'PyJWT' don't provide this feature as far as I've been able to find.

blitzmann:
Thanks for going one step further and diving into things, it make it much easier on us.

This is interesting, I would have assumed we would have heard about this long before now, as if it's a timezone issue there's got to be more than 1 person experiencing the issue. I'll try to reproduce this tonight. However, I have a few concerns. python-jose uses `datetime.utcnow()` which should give you a time that is in UTC (+0) regardless of your local time zone.

[...]

PR #2424, opened 2022-03-30
Add option in settings to disregard JWT exp issues

blitzmann:
See #2402

I don't really like including this option, but without being able to reproduce the error, and knowing that it's something we can work around, figured we can add this until we can figure out what is actually happening.


nmalaguti (approved):
The change to the JWT token verification looks good to me. The rest of it also looks good, but I have no experience with wx to do a proper review.

DarkFenX:
Looks good to me from wx side, going to merge it now for the release

Issue #2381, opened 2021-11-27
Feature Request:  Temp Equip Bucket (window)


Napoleon-BlownApart:
Great workbench for assessing ship fittings.
One thing I would like to propose is a new window/frame that allows the user to drag modules into it from where ever, and then allows the user to drag modules from it to where ever (without automatically removing them).

I find myself bouncing back and forth between `market` and `fitting` way too much, when the module I'm looking for is already in another fit.  I'm unable to use copy/paste because ctrl-c/v are used for the fitting clipboard.

It would be nice if I could select some items and drag them into this window, either from the market or from another ship, and then drag them to the ship fit I'm working on.  This window could be persistent, say as an option, and allow the user to remove items from it, say right clicking.  There is no need to store quantities, except perhaps for charges, cargo items, etc.

As an additional suggestion, it would be useful to be able to drag/fit items from the `compare` window.

PR #2439, opened 2022-05-07
Kg sd issue#2381

KefronG:
issue #2381 Cart system to store modules ... etc. GUI portion is set up and shares the same functions as cargo. remaining issues; Cut calc code from cart to avoid interfering with cargo calculations and then create a static fit that runs in the background simulating an empty cargo then link this fit ID to cartView.py. Getting no attribute error when loading in a fit due to no link from Fit.py(the fit import from API) which is another reason a static fit is needed.


blitzmann:
I always envisioned something like this as a Favorites section in the market, much like the "recently used modules", where you an right click modules and add to favorites. This also has the advantage of having all the market code associated with it for adding it to the fit (with modifiers, etc)

I'm not sure how I feel about this being one of the tabs at the bottom - those are designed to be specific to the fit, not global.

blitzmann:
@KefronG Now that I'm able to sit down and absorb this PR, first off, welcome to the repo! Thanks for opening a pull request and trying to tackle a feature, I always like to encourage this stuff :D

That being said, lets have a discussion about this before we go further. Ideally we some discussion would have taken place to flesh out the details of how we might implement such a feature. I've thought about this a bit more today, and I have some concerns with the current direction. 

> then create a static fit that runs in the background simulating an empty cargo then link this fit ID to cartView.py

Can you elaborate here? Are you envisioning a dummy fit that is initialized that is used as the carrier of the cart modules? If that's the case, it seems like a pretty heavy handed approach.

As I mentioned in my previous post, I would think this would be the perfect opportunity to implement a "Favorites" section, under "Recently Used Modules" 

image

We could right click and set any module / item in the market to "Favorite", and it would show up there. This is something that has been requested by multiple folks  over the years, and we've just never got around to it. I think this would implement much better than a tab at the bottom of the fit (which is supposed to be fit-specific)

Thoughts?

KefronG:
I'm definitely down for re working this issue completely, this issue was originally a class assignment so all my current methods are just the fastest and direct way I could think of implementation.        So the static fit was meant to work independent from the default fit you add from fitting the window. This is to avoid calculating errors with having the same ship. This second static fit would also work at an unrestricted container by changing the restrictions for this fit.
I think adding a favorites options in the drop down is a lot cleaner and does not seem as invasive as adding a completely new tab. Also we would not need a second fit to run parallel. I need to but some more thought into it I definitely like it and would like to implement this. p.s finishing up finals so sorry if I respond slowly.
Sent from Yahoo Mail on Android 
 
  On Sun, May 8, 2022 at 10:28 PM, Ryan ***@***.***> wrote:   


@KefronG Now that I'm able to sit down and absorb this PR, first off, welcome to the repo! Thanks for opening a pull request and trying to tackle a feature, I always like to encourage this stuff :D

That being said, lets have a discussion about this before we go further. Ideally we some discussion would have taken place to flesh out the details of how we might implement such a feature. I've thought about this a bit more today, and I have some concerns with the current direction.


then create a static fit that runs in the background simulating an empty cargo then link this fit ID to cartView.py


Can you elaborate here? Are you envisioning a dummy fit that is initialized that is used as the carrier of the cart modules? If that's the case, it seems like a pretty heavy handed approach.

As I mentioned in my previous post, I would think this would be the perfect opportunity to implement a "Favorites" section, under "Recently Used Modules"



We could right click and set any module / item in the market to "Favorite", and it would show up there. This is something that has been requested by multiple folks over the years, and we've just never got around to it. I think this would implement much better than a tab at the bottom of the fit (which is supposed to be fit-specific)

Thoughts?

PR #2466, opened 2022-09-04
Feature/compare highlight

yeaido:
Double clicking an item in the Compare window will now highlight and bold that row, can be returned to normal by double clicking again
compare

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,5 +125,5 @@ gitversion
 *.fsdbinary
 /locale/progress.json
 
-launch.json
-eve.db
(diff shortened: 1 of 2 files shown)


DarkFenX:
Looks good, merging. Although I might change color later, bright yellow might be a bit too bright

DarkFenX:
...well, just noticed it has eve.db which shouldn't be there. I am removing that

DarkFenX:
I changed implementation so that highlight persists through column sorting, and preserves no-highlight background color

PR #2485, opened 2022-11-28
fix proxy proto string strip

kols:
It's possible an `http://` url string set for the `https` proxy, and the code can't strip it.


DarkFenX:
Unsure who does that, but, i guess, okay

PR #2496, opened 2023-02-18
Added option to sync all fittings to ESI

josephdouce:
Added option to Sync all fittings to ESI, also forced fetch before deleteAll as pressing delete all before fetch throws an error.


josephdouce:
image

josephdouce:
Made some changes, the tool now replaces fits based on name, if names match in ESI and Pyfa, fits won't be added/removed, if there is no match the new fit will be added to ESI, ESI fits that are not matched with Pyfa will be removed.

josephdouce:
Give me a shout if you want this feature and I'll make a fresh pull form my up-to-date fork.

PR #2561, opened 2023-12-20
Add mutation info to XML export/import.

m-sasha:
Currently, when exporting in XML format, the mutation information is lost, and when reimporting the file, Pyfa will simply ignore the mutated item, leaving an empty slot.

This PR implements writing/using mutation info when exporting/importing in XML format.

For mutated items (modules and drones), write and read these `hardware` tag attributes:
- `base_type`, specifying the name of the base type.
- `mutaplasmid`, specifying the name of the mutaplasmid.
- `mutated_attrs`, specifying the mutated attributes and their values, in the format of EFT export, e.g. `capacitorNeed 108.0, cpu 34.4`.

### DIFF
--- a/service/port/muta.py
+++ b/service/port/muta.py
@@ -28,17 +28,20 @@ from service.esiAccess import EsiAccess
 
 def renderMutant(mutant, firstPrefix='', prefix=''):
     exportLines = []
+    exportLines.append('{}{}'.format(firstPrefix, mutant.baseItem.name))
+    exportLines.append('{}{}'.format(prefix, mutant.mutaplasmid.item.name))
+    exportLines.append('{}{}'.format(prefix, renderMutantAttrs(mutant)))
+    return '\n'.join(exportLines)
+
+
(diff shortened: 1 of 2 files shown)


DarkFenX:
Does EVE use XML for anything fit-related?

Primary reason I didn't add it to XML right away was that EVE was using XML for import/export, and I didn't want to break it (or implement, then test, then fix or find workarounds for stuff which was broken).

DarkFenX:
Ok, seems like it still uses XML for batch export/import. Can you test how mutated modules and drones are interpreted by EVE during import?

m-sasha:
It ignores the extra attributes, and reacts to the mutated item as before:

image

m-sasha:
Just to clarify - as EVE ignores the extended attributes, this PR changes nothing as far as it's concerned.

A fit with mutated modules exported from EVE and then imported back into EVE will have the same effect as the same fit exported from Pyfa (which is exactly how it behaved before).

PR #2568, opened 2024-01-04
Add drone orbit modeling to `getDroneMult`

dambrisco:
Drone damage application is especially important in contexts like ratting where tracking frequently comes into play in fitting decisions. As-is, the Pyfa engine always assumes drones hits, but this, of course, isn't accurate.

This proposed change uses approximations of drone behavior to provide a _more_ realistic (but still admittedly lacking) calculation based on the drone's orbit speed, the target's speed, and (very) broad generalizations about dynamic orbit planes. The result is more accurate tracking-based damage numbers relative to the in-game experience.

The biggest downside to this change is that the calculation is unintuitive for small, fast-moving ships. In-game, these ships are generally in some sort of orbit around you, but the calculation used here will result in an extremely low angular velocity to the drones which doesn't take into account orbit ranges or curvature of the flight path and which, in turn, results in better application for slow-moving drones than you'd see in an in-game scenario. Given that the existing chance-to-hit value is simply `1`, I feel this is still a net improvement.

##### Current
- All velocities
  - 64.31 DPS
##### Proposed
- 0% velocity
  - 3.759 DPS
- 50% velocity
  - 22.59 DPS
- 100% velocity
  - 64.31 DPS
##### Current
- All velocities
  - 482.4 DPS
##### Proposed
- 0% velocity
  - 476.2 DPS
- 50% velocity
  - 476.9 DPS
- 100% velocity
  - 478.8 DPS

These numbers might _look_ weird, but they're a result of the way orbits work and the included approximation of distortions to the orbit plane caused by a moving target - drones orbit stationary targets at a perpendicular, but spend more time approaching targets that are moving at velocities closer to their own cruise speeds.

### DIFF
--- a/graphs/data/fitDamageStats/calc/application.py
+++ b/graphs/data/fitDamageStats/calc/application.py
@@ -257,19 +257,49 @@ def getDroneMult(drone, src, tgt, atkSpeed, atkAngle, distance, tgtSpeed, tgtAng
     ):
         return 0
     droneSpeed = drone.getModifiedItemAttr('maxVelocity')
-    # Hard to simulate drone behavior, so assume chance to hit is 1 for mobile drones
-    # which catch up with target
+    # A drone chasing a target will always try to get within `entityFlyRange` before
+    # stopping its MWD, and will always reactive its MWD when outside of its optimal
+    # range. In cases where the MWD is almost always active, we assume the drone is
+    # constantly approaching the target from behind.
+    # If the drone orbits at a speed faster than or as fast as the target is moving,
+    # we can treat it as though it never uses its MWD and we can average the angle
+    # based on the fact that the drone will always try to move perpendicular to the
+    # target, but the closer the target's speed it to the drone's speed, the lower
+    # the average effective angle.
+    # In both cases, distance doesn't matter because the drone is within its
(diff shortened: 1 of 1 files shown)


DarkFenX:
This is rather complex subject. The solution given seems strange to me and doesn't include the fact that drones lag behind fast target before resuming MWD cycle. I think I had a dude on reddit who gave me an approximate solution to this, but I never got to testing / implementing it. Here is a [reference to oNordak's comment](https://www.reddit.com/r/Eve/comments/ctvpp4/pyfa_v2100_released/exr0cgd/) with formulas.

dambrisco:
@DarkFenX Appreciate the consideration. If you don't feel this PR is a sufficient improvement over the existing implementation, I'll suggest that we go ahead and close this. I stand by it as a net improvement (in general but especially for PvE), but I'm content to keep this patched to my personal fork pending an improved upstream impl.

I'll note that oNordak's suggestions don't account for orbital angular velocity, so if you do decide to pursue that you may want to account for those using either the approach I have here (using relative non-MWD velocities and orbit range to calculate average angular velocity parallel to the target's trajectory) or a more robust approach (which would need to account for variable orbital plane offsets from the target's trajectory).

PR #2601, opened 2024-03-29
Prepend Current Fit's Ship and Name to the Window's Title

Nucanot:
When you have multiple fits open, the names of the fits in the tabs get truncated and it's hard to tell which fit is which (especially when looking at different variations of the same fit).

This PR solves the issue by adding extra information to the window's title—the ship and the fit name.

Format is `<ship name> "<fit name>" — <default title>`
`<default title>` is the title, either "pyfa" or "pyfa v1.2.3 - Python Fitting Assistant"

The title is update when the active fit changes (via the tab switcher) or the current fit is renamed.

### DIFF
--- a/gui/mainFrame.py
+++ b/gui/mainFrame.py
@@ -234,6 +234,9 @@ class MainFrame(wx.Frame):
 
         self.Bind(GE.EVT_SSO_LOGIN, self.onSSOLogin)
 
+        self.Bind(GE.FIT_CHANGED, self.onFitChanged)
+        self.Bind(GE.FIT_RENAMED, self.onFitRenamed)
+
     @property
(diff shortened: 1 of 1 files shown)


DarkFenX:
Thanks. I will try to include it for next release (don't have time for the today's one, sorry)

Nucanot:
> Thanks. I will try to include it for next release (don't have time for the today's one, sorry).
> 
> Although there might be side effects, I know on some OS' there is a notification when window title changes (think it was default gnome desktop? might want to test it there to see if it's causing any issues or not)

I've only tested on macOS. I will try to spin up a Linux VM when I have time and test there.

I would be surprised if the GNOME window manager fires notifications on title change though (haven't seen this when I was using Linux before).

Issue #2261, opened 2020-08-29
Error!!  (Pyfa told me to report the error and post the log here)


Davak-Kateelo:
## Bug Report
```
pyfa v2.27.0
EVE Data Version: 1792574 (2020-08-25 04:10:32)

OS version: Windows-10-10.0.18362-SP0
Python version: 3.7.8 (tags/v3.7.8:4b47a5b6ba, Jun 28 2020, 08:53:46) [MSC v.1916 64 bit (AMD64)]
wxPython version: 4.0.6 (wxWidgets 3.0.5)
SQLAlchemy version: 1.3.19
Logbook version: 1.5.3
Requests version: 2.24.0
Dateutil version: 2.8.1

####################

Traceback (most recent call last):
  File "gui\builtinShipBrowser\sfBrowserItem.py", line 285, in OnPaint
wx._core.wxAssertionError: C++ assertion "!bitmap.GetSelectedInto() || (bitmap.GetSelectedInto() == GetOwner())" failed at ..\..\src\msw\dcmemory.cpp(116) in wxMemoryDCImpl::DoSelect(): Bitmap is selected in another wxMemoryDC, delete the first wxMemoryDC or use SelectObject(NULL)

####################

Traceback (most recent call last):
  File "gui\builtinShipBrowser\sfBrowserItem.py", line 292, in OnPaint
  File "gui\builtinShipBrowser\sfBrowserItem.py", line 236, in Render
wx._core.wxAssertionError: C++ assertion "bmp.IsOk()" failed at ..\..\src\msw\dc.cpp(1313) in wxMSWDCImpl::DoDrawBitmap(): invalid bitmap in wxMSWDCImpl::DrawBitmap

####################

Traceback (most recent call last):
  File "gui\builtinShipBrowser\sfBrowserItem.py", line 292, in OnPaint
  File "gui\builtinShipBrowser\sfBrowserItem.py", line 236, in Render
wx._core.wxAssertionError: C++ assertion "bmp.IsOk()" failed at ..\..\src\msw\dc.cpp(1313) in wxMSWDCImpl::DoDrawBitmap(): invalid bitmap in wxMSWDCImpl::DrawBitmap

####################

Traceback (most recent call last):
  File "gui\builtinShipBrowser\sfBrowserItem.py", line 292, in OnPaint
  File "gui\builtinShipBrowser\sfBrowserItem.py", line 236, in Render
[...]

Issue #2225, opened 2020-07-09
Crash on search. 


kanyda:
Crashed when I search for a stork

https://i.imgur.com/akCNjDC.png

Unfortunately I can't copy the error box, I did screencap the first and last part of it for whatever value that is: 

https://i.imgur.com/kFS1PVj.png

https://i.imgur.com/RUc9h6g.png

I search for what I thought was the error log and I think this was it: 

[2020-07-09 08:31:54.990517] INFO: service.fit: ==========recalc time: 0.04455113410949707==========
[2020-07-09 08:31:54.990517] DEBUG: service.fit: Check states for fit ID: pvp bifrost booster (Bifrost)
[2020-07-09 08:32:12.603327] DEBUG: service.fit: Searching for fit: sto
[2020-07-09 08:32:12.619342] DEBUG: service.fit: Fetching all fits for ship ID: 1944
[2020-07-09 08:32:12.626348] DEBUG: service.fit: Fetching all fits for ship ID: 33472
[2020-07-09 08:32:12.637358] DEBUG: service.fit: Fetching all fits for ship ID: 37482
[2020-07-09 08:32:12.645375] DEBUG: service.fit: Fetching all fits for ship ID: 12747
[2020-07-09 08:32:12.650571] DEBUG: service.fit: Fetching all fits for ship ID: 54732
[2020-07-09 08:32:12.736447] DEBUG: service.fit: Searching for fit: stoi
[2020-07-09 08:32:12.859286] DEBUG: service.fit: Searching for fit: sto
[2020-07-09 08:32:12.865289] DEBUG: service.fit: Fetching all fits for ship ID: 1944
[2020-07-09 08:32:12.870295] DEBUG: service.fit: Fetching all fits for ship ID: 33472
[2020-07-09 08:32:12.876301] DEBUG: service.fit: Fetching all fits for ship ID: 37482
[2020-07-09 08:32:12.881296] DEBUG: service.fit: Fetching all fits for ship ID: 12747
[2020-07-09 08:32:12.886310] DEBUG: service.fit: Fetching all fits for ship ID: 54732
[2020-07-09 08:32:13.116510] DEBUG: service.fit: Searching for fit: sti
[2020-07-09 08:32:13.146546] DEBUG: service.fit: Fetching all fits for ship ID: 633
[2020-07-09 08:32:13.155545] DEBUG: service.fit: Fetching all fits for ship ID: 2836
[...]

PR #2629, opened 2024-08-15
Fix wxWidget crash caused by GDI object limit

cryonox:
This closes 

https://github.com/pyfa-org/Pyfa/issues/2121
https://github.com/pyfa-org/Pyfa/issues/2261
https://github.com/pyfa-org/Pyfa/issues/2225

### DIFF
--- a/gui/builtinShipBrowser/pfListPane.py
+++ b/gui/builtinShipBrowser/pfListPane.py
@@ -163,6 +163,10 @@ class PFListPane(wx.ScrolledWindow):
     def RemoveAllChildren(self):
         for widget in self._wList:
             widget.Destroy()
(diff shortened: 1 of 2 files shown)


DarkFenX:
Sorry for missing the pull request. Was busy with AT and kids. Will look into it later this week

cryonox:
Ya it was used to check "memory leak" from gc and to know the current gdi object count so that I know if it's about to crash or not. I thought it might be useful in case this kind of thing happen again since I am pretty sure shipbrowser is not the only place this is happening. it's just the worst case of it.

DarkFenX:
Thx, i wish i had enough experience to debug similar crashes on linux/mac

cryonox:
Did you forgot to merge? @DarkFenX

DarkFenX:
Oh wtf

DarkFenX:
Sorry about that

PR #2653, opened 2024-11-30
Importing the localized fitting file(xml) exported from EVE

jeffy-g:
Hello!

I haven't played EVE for a while, but I've started playing again recently.

Inevitably, I ended up using `Pyfa` again,
but I feel that I need to make some changes to the import of localized fitting files, so I am submitting a PR.

For simple testing, I created **localized fitting files** for each language and placed them in **`tests/localized-fitting-files`**.
You can see how it works by importing all the xml files at once with `Pyfa`.

It also seems that the character limit for fitting notes has been increased from **400** to **500**,
so I've included those changes in this PR.

If the character count exceeds 500, the text will turn red.

thanks for reading o7


jeffy-g:
The German and Russian localized fitting files export xml files with the exact same contents regardless of important names.

> hint attr is **&lt;official name>**, text content is **&lt;official name>***

DarkFenX:
I will take a look at it (and possibly merge) later. Will make today release w/o this, sorry

jeffy-g:
Okay! np :-)

Please check it out, if there's anything you don't like or any problems, please fix it.

PR #2700, opened 2025-11-22
Fixed bug with Clipboard on Wayland

skyride:
There's currently a bug with the Linux release under Wayland where the copy/paste functionality doesn't work. It doesn't reliable access the clipboard as expected, so it'll silently fail to copy _from_ clipboard, and most applications are unable to access things paste _to_ the clipboard.

This PR fixes that by forcing it to use the actual Clipboard instead of Primary. Primary is a neat feature, but the behaviour is inconsistent if your application isn't built with its special usage in mind.

## Testing

I've tested this on Wayland Linux, but I'd suggest double checking it works as expected on Windows and Mac before merging. I'm trying to see if I can get some help with that.

### DIFF
--- a/gui/utils/clipboard.py
+++ b/gui/utils/clipboard.py
@@ -1,22 +1,69 @@
 # noinspection PyPackageRequirements
 import wx
+from logbook import Logger
+
+logger = Logger(__name__)
 
 
 def toClipboard(text):
-    clip = wx.TheClipboard
-    clip.Open()
-    data = wx.TextDataObject(text)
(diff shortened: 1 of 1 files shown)


DarkFenX:
I will merge and test windows version via wine. Unfortunately, no mac device to test it on, but I will just be ready to make a quick fix if it breaks anything.

DarkFenX:
Nevermind, it broke clipboard even under wayland for me. Import works fine, export does not

PR #2706, opened 2025-12-17
fix: Fixed clipboard issues on Wayland, using pyclip instead of wx

freeone3000:
Clipboard import and export was continually flaky under Bazzite. Possibly under other wayland? pyclip seems to be the most reliable on my local setup. 

Please review and let me know if this is acceptable. This working for me may still mean it does not work for others.

### DIFF
--- a/gui/copySelectDialog.py
+++ b/gui/copySelectDialog.py
@@ -151,7 +151,6 @@ class CopySelectDialog(wx.Dialog):
         def cb(text):
             if self.waitDialog:
                 del self.waitDialog
-            print("cb called. Text: {}", text) # TODO REMOVE DEBUG
             toClipboard(text)
             self.EndModal(wx.ID_OK)
 


DarkFenX:
https://github.com/spyoungtech/pyclip/blob/main/pyclip/wayland_clip.py seems like it's using `wl-copy` utility to do it, which means it's not just extra python package, but also extra binary dependency which has to be installed. Idk how pip/uv handle it. This is possibly a risk since e.g. linux builds need to have both dependencies installed (`xclip` for X11 and `wl-clip` for wayland for copying, similar utilities for pasting). Same for macos by the looks of it.

DarkFenX:
Well yes it needs extra dependencies to work. Pyclip package itself does not pull them. I guess I will have to see what it uses and ensure every image has appropriate extra packages installed

PR #2726, opened 2026-02-16
Ammo breakdown

MrWildGear:
Ammo Breakdown is a custom tool that scans a ship and reports exactly which ammo types it carries, then builds a table with key stats for each: ammo name, damage type, optimal range, falloff, alpha, and DPS. If you have ever tried to build a doctrine cheat sheet, you know it usually means loading every ammo type into each fit and manually copying the stats into a spreadsheet or app; this tool reduces that to a process that takes only a few seconds, with the main time cost being importing your doctrine ships into Pyfa. This code was originally generated with AI; if it does not fit the direction of the main project, that is completely fine, but I wanted to share it so anyone can use or build on it.

Example of the Ammo Breakdown.
image

### DIFF
--- a/_development/helpers.py
+++ b/_development/helpers.py
@@ -92,7 +92,7 @@ def DBInMemory_test():
     return helper
 
 # noinspection PyUnresolvedReferences,PyUnusedLocal
-@pytest.fixture
+@pytest.fixture(name='DB')
 def DBInMemory():
     print("Creating database in memory")
 
--- a/gui/ammoBreakdown/__init__.py
+++ b/gui/ammoBreakdown/__init__.py
@@ -1,0 +1,20 @@
+# =============================================================================
+# Copyright (C) 2010 Diego Duclos
+#
+# This file is part of pyfa.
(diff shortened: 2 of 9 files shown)


DarkFenX:
Since some point, there are graphs which serve similar, but slightly different purpose: they show volley/DPS breakdown per ammo. The biggest difference is that they do not limit ammo to things in ship's cargo.

Do you think having graphs is sufficient or not? If not, what specifically do you miss in those?

MrWildGear:
I apologizer it taken so long to get back to you.

I made the Ammo Breakdown tool to be a quick system to make a cheat sheet.
with the graphs in order to do the same thing you would need to copy to ship for each ammo type to see them all together in the graphs. wall that dose work it takes longer than it needs to. 

now if there is a way that you can see on the graphs page with all the ammo types all at once and can be done in one click or automatically based on what ammo is on the ship that would work. also, having an export button that break the graphs down into text would be helpful for doctrine cheat sheets.

PR #2741, opened 2026-04-03
Filter graphs on selection

Skybladev2:
# Problem

Showing multiple fits in Graph window becomes a mess quickly even for trivial cases when you want to compare how 2 fits perform against each other. This produces N*M graphs which are hard to read even with with color keys.

image

# Solution

Take selections in attacker and target lists into account so only graphs for selected fits are displayed.

If no fit is selected the behavior is unchanged.

Examples:

## Selecting attackers

image

image

## Selecting targets

image

image

## Selecting attackers and targets

image

### Selecting attackers and targets with reverse matchups

image

### DIFF
--- a/eos/saveddata/fit.py
+++ b/eos/saveddata/fit.py
@@ -19,6 +19,7 @@
 
 import datetime
 import time
+from contextlib import contextmanager
 from copy import deepcopy
 from itertools import chain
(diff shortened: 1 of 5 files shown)


DarkFenX:
I will take a look at this for next release. No time to check it for the sisi release I am doing right now, sorry

Issue #1193, opened 2017-05-27
Extraorbitant stacking with circular projection / command bursting


lunedis:
### Expected behavior:
Normal bonus onto Crucifier

### Actual behavior:
TD shows range of 1061km+354km with 581% strength (lol)

### Detailed steps to reproduce:
Create both fittings and open them.
Project Crucifier onto Nighthawk.
Add Nighthawk as Command Booster for Crucifier.

### Fits involved in EFT format (Edit > To Clipboard > EFT):
```
[Crucifier, TEST]

[Empty Low slot]
[Empty Low slot]
[Empty Low slot]

Tracking Disruptor II, Optimal Range Disruption Script
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]

[Empty High slot]
[Empty High slot]

[Empty Rig slot]
[Empty Rig slot]
[Empty Rig slot]
```
```
[Nighthawk, TEST]

[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]
[Empty Low slot]

[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]
[Empty Med slot]

Information Command Burst II, Electronic Superiority Charge
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]
[Empty High slot]

[Empty Rig slot]
[Empty Rig slot]


Information Command Mindlink
```
### Release or development git branch? Please note the release version or commit hash:
Pyfa 1.29.2

### Operating system and version (eg: Windows 10, OS X 10.9, OS X 10.11, Ubuntu 16.10):
Windows 7

### Other relevant information:

Yeah it's me with the Info links again... sorry

blitzmann:
Yep, this was reported to me privately along with data needed to verify (which was AT sensitive, so can't post the exact details). I never got around to making a general bug report about it. I know what the problem is and have a potential fix for it, but am wanting to test it out a bit more thoroughly (also, I didn't know this affected projections as well - I only verified and fixed the command issue. Testing recursive projections was on the list of things to test it with)

Will update issue with more info as required, or when a fix is in place. :)

PR #2742, opened 2026-04-05
Fix tab switch when projected is mutually applied

Skybladev2:
# Problem

When 2 fits are mutually applied in Projected tabs to each other, the first fit loses the projected effect.

For example, we have fit1:
image

and fit2:
image

Each one has a Stasis Webifier.

When the first fit applies projected effect from the second one everything is fine:
image

When the second fit applies projected effect from the first one everything is fine:
image

When we switch back to the first fit tab, the projected effect is lost:
image

It can be restored by re-enabling the projected effect, but then the effect is lost on the second fit tab:
image

If we re-enable it here, the first fit loses the projected effect and so on.

# Solution

Don't skip recalculation of modified attributes for projected calculation type (similar to command, see #1193).

### DIFF
--- a/eos/saveddata/fit.py
+++ b/eos/saveddata/fit.py
@@ -1108,10 +1108,7 @@ class Fit:
         if type == CalcType.COMMAND and targetFit in self.commandFits:
             pyfalog.debug("{} is in the command listing for COMMAND ({}), do not mark self as calculated (recursive)".format(repr(targetFit), repr(self)))
         elif type == CalcType.PROJECTED:
-            # A projected-source pass only applies this fit's remote effects to the target; it is not a full local
-            # calculation of this fit. Marking calculated would make switchFit skip recalc, so bidirectional
-            # projections show wrong stats when switching tabs until the projection is toggled.
-            pass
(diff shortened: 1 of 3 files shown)


Skybladev2:
AppVeyor build failed because [macos job](https://ci.appveyor.com/project/pyfa-org/pyfa/builds/53866701/job/tqxq8q33to920dfh) couldn't pull some dependencies:
```
ERROR: Could not install packages due to an OSError: HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Max retries exceeded with url: /packages/6b/3d/ac556c1208c701c4f6ccde77db0b2eeb84bf052d63260c2305fb5ace707c/pyinstaller-6.2.0-py3-none-macosx_10_13_universal2.whl.metadata (Caused by NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x1089a7f50>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))
```

Please rerun the job.

DarkFenX:
Ngl touching this code is scary. To avoid those issues, just duplicate one of fits to avoid looping.

All of those would be automatically fixed with the calc engine switch, so I'd prefer not to touch this for now.

Skybladev2:
The duplication is suboptimal to user experience, but I get the point and wait for the engine update.
Thanks for the info!

PR #2747, opened 2026-04-14
Thermodynamics Calculations

AnsbigetHildElarik:
### Summary

Adds thermodynamics calculations to estimate overload duration and show as column in fitting view. 

### Notes

- I wrote this code 12 years ago for my own fitting tool so I've left in the ported js code for reviewers but it shoud be removed prior to merge.
- It would be nice to i18n the one message but I'm not sure how to make that happen, there is a comment in there about it.
- There could be better ways to accompish the pythonic/eos/fit things, happy to see that improved if possible. 
- This could use some testing. I did some manual tests to verifiy the burnout times, they lined up fine and the math should be correct.

### Screenshots

pyfa-thermo

### DIFF
--- a/gui/builtinViewColumns/heat.py
+++ b/gui/builtinViewColumns/heat.py
@@ -52,7 +52,7 @@ class Thermodynamics():
         harm = [0,0,0,0] # 0 is a dummy slot, align with mod.slot constants, 1=low, 2=med, 3=hi, 4=rig, ...
 
         for mod in self.fit.modules:
-            if(mod.state == 2 and mod.slot <= 3):
+            if(mod.state == FittingModuleState.OVERHEATED):
                 harm[mod.slot] += mod.getModifiedItemAttr("heatAbsorbtionRateModifier")
 
         return harm
--- a/gui/builtinViewColumns/heat.py
+++ b/gui/builtinViewColumns/heat.py
@@ -34,7 +34,6 @@ class Thermodynamics():
     def __init__(self, fit):
         self.fit = fit
         self.hgm = fit.ship.getModifiedItemAttr("heatGenerationMultiplier")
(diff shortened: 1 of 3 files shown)


DarkFenX:
This is very useful, thanks. I couldn't find acceptable solution for engine rewrite here: https://github.com/pyfa-org/eos/issues/12

So, before merging I will check if solution exposed by sim makes sense. Also, I really wanted to give mean time & time range for some confidence interval (like 90%?) in tooltip, since heat due to its random nature can provide very random results (esp in edge cases like retribution with just MWD on it - due to fairly low chance to damage it, it can break in 3 cycles, can break in 500 cycles).

AnsbigetHildElarik:
Great to see this [info](https://github.com/pyfa-org/eos/issues/12) all in one place. 

Probability is definitely a tricky thing to expose to users, but keeping it simple might be better. I think it is nearly deterministic, like stating that a coin flip is 50/50. The [simTime](https://github.com/pyfa-org/Pyfa/pull/2747/changes#diff-7621b02399b8fd20ad87a461a5b5111104ac3ad9555ee344e017afea84da8047R39) should be increased, most mods probably converge long before the simTime limit and we may get slightly more accuracy from expanding [this](https://github.com/pyfa-org/Pyfa/pull/2747/changes#diff-7621b02399b8fd20ad87a461a5b5111104ac3ad9555ee344e017afea84da8047R191). At the time I was very worried about js performance.

I went down this rabbit hole a long time ago and I consulted a [wizard](https://github.com/pyfa-org/Pyfa/pull/2747/changes#diff-7621b02399b8fd20ad87a461a5b5111104ac3ad9555ee344e017afea84da8047R233) at  the end. Adding intervals could be easy but since the underlying formula is pretty complex I would ask a wizard (mathematician on stack exchange) to get it right. Back then my understanding was worthy and allowed me to speak their tongue, not today. It has been a long time but I think it is just a special case expansion of the negative binomial to allow for changing P with time.

DarkFenX:
@AnsbigetHildElarik https://github.com/pyfa-org/Pyfa/issues/2754 if you can, please look into it

PR #2748, opened 2026-04-14
Fix/calibration points overflow

Skybladev2:
# Problem

Current calibration points highlight logic is broken. It displays text in red when calibration points are not exceeded, but keeps normal text style when they are not exceed:

image

image

image

# Solution

Force label refresh, because the style was invalidated with a delay, causing visual discrepancy.


DarkFenX:
Thanks. It is a windows-specific issue I couldn't reproduce under wine, so couldn't really figure what kind of update in wx borked it.

PR #2760, opened 2026-05-11
Add damage projection graph

NicolasKion:
New "Damage Projection" graph plots the per-distance envelope of best effective DPS across each weapon's compatible ammo, summed across weapons. Charge swaps are batched per (item ID, state) group so the fit only recalcs ceil(N_ammo / N_group_size) times instead of once per ammo, and dominated ammo curves are pruned upfront so the per-distance loop only iterates candidates that can plausibly win. This makes it very straightforward to compare different hulls in terms of their projected dps capabilities.

AI has been used to produce the code in question

Screenshot 2026-05-11 at 12 43 15

### DIFF
--- a/graphs/data/fitDamageEnvelope/__init__.py
+++ b/graphs/data/fitDamageEnvelope/__init__.py
@@ -18,14 +18,6 @@
 # =============================================================================
 
 
-from . import fitDamageStats
-from . import fitDamageEnvelope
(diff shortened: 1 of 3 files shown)


DarkFenX:
Thanks (I also won't spend too much time checking it, busy with other stuff rn)

Issue #1605, opened 2018-05-22
Start up lag


blitzmann:
I'm not sure if this is still an issue, but there have been a few reports in the 2.x series of pyfa taking a long time to start up. I'm creating this issue to link the other issues.

### If you're experiencing this issue

First, please subscribe to the thread to receive new updates and posts! The subscribe button it to the right >>>>

Attaching log files can help determine where the slow down is taking place. Your logs are found in `%userprofile%/.pyfa` and end in `.log`. Find the one that shows longish start up times (each log is time-stamped so it should be easy-ish to find gaps when it starts...)

As I investigate, I may also post in this thread new builds that have additional logging to try to narrow down where the issues lie. it would be helpful to keep an eye on notification for any updates to this thread and provide logs and feedback as I work through this progress.

Thanks!

Calavoow:
My pyfa has been starting noticeably slower since 2.0. Here are my logs 
[pyfa-2018-05-31.log](https://github.com/pyfa-org/Pyfa/files/2056179/pyfa-2018-05-31.log)

blitzmann:
Thanks @Calavoow! That log shows about 3 seconds of start up, which isn't too bad and kind of normal. Dump another log if you notice it taking much longer than that :) (unless of course this one as a long startup, and the log just didn't catch it?)

Calavoow:
Another one, after a fresh start [pyfa-2018-06-01.log](https://github.com/pyfa-org/Pyfa/files/2060698/pyfa-2018-06-01.log)
Seems like 13 seconds this time

blitzmann:
Thanks!

I need to put a lot more debugging statements when various thing initialize to try to track down where this is taking a while. I'll try to get you a test build that you can run that should show a lot more granularity

Nimos:
[pyfa_debug-2018-06-01.log](https://github.com/pyfa-org/Pyfa/files/2063473/pyfa_debug-2018-06-01.log)

here's one of mine, I've been struggling with long startup times for a while

blitzmann:
I'm attempting something that is either completely clever, or completely crazy

With bb978c3dffd772d48304af123a33d4827184da06 you should now be able to specify a path to save profiling information. While this introduces additional overhead, as well as possible increased start up times, it will track the function calls within pyfa and allow me to determine what is eating up time. This will save to file in whatever directory you specify, and then you can upload it here for me to analyse.

If you don't specify a profile location, then it works just as it does now, so no overhead

[...]

PR #2764, opened 2026-05-16
Use onedir PyInstaller bundles for faster macOS launch

Collinw24:
## Summary

This changes the macOS PyInstaller bundle from onefile to onedir.

The current onefile app has a large amount of work before pyfa itself starts. On macOS that makes the packaged app look like it is hanging for a long time before any window appears. With onedir, the bundled files live inside the app bundle instead of being unpacked into a temp directory on every launch.

This keeps the output as `pyfa.app` and does not change pyfa startup/runtime code.

The packaging script now uses `zip -y` so macOS bundle symlinks are stored as symlinks in the release archive instead of being followed and duplicated.

Refs #1605.

## Measurements

Measured on:

- Apple M3 Max
- macOS 26.3.1
- Python 3.11.9
- PyInstaller 6.2.0, matching `scripts/osx-setup.sh`
- local package builds from `scripts/osx-package.sh`

| Build | Package format | First window |
| --- | --- | --- |
| `upstream/master` at `a46ad2b81` | onefile app bundle | no window within the 60s timeout in 3/3 measured runs |
| this branch | onedir app bundle | 1.54s / 1.59s / 1.68s, median 1.59s |

I also used a temporary instrumented build to split out the pre-Python time. That instrumentation is not included in this PR. With the onedir bundle, the median timing was:

- first pyfa bytecode: 0.158s
- `MainFrame.Show()`: 1.48s
- first idle: 1.66s

## Tradeoff

The installed app bundle is larger and contains more files, but preserving symlinks in the zip keeps the release archive close to the current onefile size:

| Build | App bundle | Zip | Bundle contents |
| --- | ---: | ---: | ---: |
| onefile baseline | ~80M | ~81M | 4 files |
| onedir | ~251M | ~96M | 5,062 files + 94 symlinks |

## QA

- `python scripts/compile_lang.py`
- `python db_update.py`
- `PYFA_VERSION=qa-pyi62-onedir-symlinkzip bash scripts/osx-package.sh` using PyInstaller 6.2.0
- extracted the script-built zip with `/usr/bin/unzip` and verified it preserved 94 symlinks
- `codesign --verify --deep --strict --verbose=2` on the extracted app
- extracted-app launch check with a 5s settle between relaunches: 1.64s / 1.65s / 1.65s / 1.71s / 1.99s, median 1.65s
- `git diff --check`

Local pytest note: I did not treat pytest as the validation path for this packaging-only change. In this checkout, current local collection fails before tests run due existing test setup/import issues (`saveddata.db` path on normal run, then `IPortUser` import under `TRAVIS=true`).

### DIFF
--- a/dist_assets/mac/pyfa.spec
+++ b/dist_assets/mac/pyfa.spec
@@ -65,20 +65,28 @@ pyz = PYZ(a.pure, a.zipped_data,
 
 exe = EXE(pyz,
           a.scripts,
-          a.binaries,
-          a.zipfiles,
-          a.datas,
+          exclude_binaries=True,
           name='pyfa',
           debug=False,
           strip=False,
           upx=True,
-          runtime_tmpdir=None,
           console=False ,
           icon=icon,
+          contents_directory='app',
           )
 
-app = BUNDLE(
+coll = COLLECT(
     exe,
+    a.binaries,
+    a.zipfiles,
+    a.datas,
+    strip=False,
+    upx=True,
+    name='pyfa',
+)
+
+app = BUNDLE(
+    coll,
     name='pyfa.app',
     version=os.getenv('PYFA_VERSION'),
     icon=icon,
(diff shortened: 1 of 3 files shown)


Collinw24:
Closing this draft in favor of a fresh PR with the macOS archive-size tradeoff clarified after preserving bundle symlinks in the release zip.

PR #2767, opened 2026-05-25
Fix damage projection graph and add volley output

NicolasKion:
Reimplements the damage projection graph from #2760, utilizing work done by @GET-R3AL. The previous version had several correctness issues around charge handling and projected fits; this rewrite fixes them and splits the logic into focused modules per weapon type so it's easier to reason about.

Also adds a Volley output alongside DPS, so you can switch between plotting effective DPS and effective volley damage on the same graph.

A Follow-up PR will annotate the graph to show which ammo backs each segment of the envelope curve.


NicolasKion:
Obsolete now that https://github.com/pyfa-org/Pyfa/pull/2779 has been merged

PR #2771, opened 2026-05-30
Fix macOS local build crash when PYFA_VERSION is unset

benkwokcy:
## Problem
Building the macOS app from a local checkout crashes at the BUNDLE step:

    TypeError: unsupported type: <class 'NoneType'>
    ...
    File "dist_assets/mac/pyfa.spec", line 91, in <module>
        app = BUNDLE(...)
    plistlib.dump(info_plist_dict, plist_fh)

`dist_assets/mac/pyfa.spec` reads the bundle version from `os.getenv('PYFA_VERSION')` in three places. That env var is only set in CI, so outside CI the value is `None`, and `plistlib` cannot serialize `None` into `Info.plist`. The build fails right after COLLECT succeeds.

## Fix
Fall back to the `git describe --tags` label the spec already computes at the top (and writes to `.version`). CI is unchanged — `PYFA_VERSION` still wins when set; local builds now succeed and get the git-describe version in the plist instead of crashing.

Also ignore the generated `eve.db` (~99 MB, built by `db_update.py`), which was untracked and easy to commit by accident.

## Testing
- `PyInstaller -y --clean dist_assets/mac/pyfa.spec` with `PYFA_VERSION` set → builds (unchanged behavior).
- Same command with `PYFA_VERSION` unset → previously crashed, now builds; `CFBundleShortVersionString` = `v2.66.3-8-g0469e9733`.
- Resulting `pyfa.app` launches.

Built on macOS arm64 (M1), Python 3.11, PyInstaller 6.2.0.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


DarkFenX:
Thank you for the fix

PR #2776, opened 2026-06-08
Feature: Display Omega icons for hulls and modules

Skybladev2:
# Problem
Some hulls and modules can be used by Omega clones only, but it's not reflected in Pyfa item list.

This makes building Alpha-clone friendly fits difficult, as there is no clear indication if a module will work.

# Solution

Display Omega icons for hulls and modules both in item list and fits:

image_2026-06-08_09-47-11
image_2026-06-08_09-47-54

## Implementation details

The Omega modules/hulls are gated by skills only (proof: https://github.com/esi/esi-issues/issues/133#issuecomment-341815740)
The Omega attribute is calculated and persisted for better performance.

### DIFF
--- a/db_update.py
+++ b/db_update.py
@@ -664,6 +664,27 @@ def update_db():
             else:
                 attr.value = value
 
+    def processCloneGradeRestrictions():
+        print('processing clone grade restrictions')
+        alpha_caps = {}
+        for clone in eos.db.gamedata_session.query(eos.gamedata.AlphaClone).all():
+            for skill in clone.skills:
+                alpha_caps[skill.typeID] = skill.level
+        for item in eos.db.gamedata_session.query(eos.gamedata.Item).all():
+            if not item.reqskills:
+                continue
+            try:
+                reqskills = json.loads(item.reqskills)
+            except (ValueError, TypeError):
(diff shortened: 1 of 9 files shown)


DarkFenX:
I think ideally this thing should be done via some kind of a filter (done with an "alpha mode" which can be toggled somewhere). But, i'd keep it for pyfa-NG or w/e it's gonna be called.

This one I think is nice to have, but as an opt-in feature. Can you make it disabled by default, and make it so that it can be enabled via options?

The reason for that is that pyfa is definitely a power user tool in my opinion. People who get to using it usually have access to omega, and unnecessary (for them) visual clutter is not something nice-to-have.

Skybladev2:
> make it disabled by default, and make it so that it can be enabled via options

Done.

I would argue that Pyfa is more for powerful users. I think it's more valuable for newbros, but their use cases much simpler.

The original reason I introduced the feature was the situation when I fitted my ship with Omega modules at the end of the subscription.  
Then I logged off in space, logged in next day and found out my Omega modules were unusable.  
I would make the fit Alpha-friendly beforehand if it was indicated in Pyfa.

PR #2779, opened 2026-06-22
Replace Damage Envelope graph with Application Profile graph

GET-R3AL:
Swaps out the old Damage Envelope graph for a new Application Profile graph. Instead of graphing a single segment it draws each ammo as its own segment, so you can see where one ammo overtakes another and read the ammo type at each distance.

New options in the control panel (only show up for this graph):

- Ammo Style (None/Pattern/Color) for how segments are drawn
- Ammo Meta (T1/Navy/All) to limit which charges get considered
- The distance axis now defaults to the fit's actual weapon range instead of a flat 0–100km

There's also per-segment legend/coloring on the canvas, and the source list auto-switches to Pattern style when multiple fits use the same weapon class so you can still tell them apart. Added Apply Projected Effects / Ignore Target Resists context menu options for this graph too, with their own settings keys.

The new code lives in `graphs/data/fitApplicationProfile/` (getter/graph + calc helpers for the turret/launcher math, ammo selection, etc.), and the old `fitDamageEnvelope` module is removed.

pyfa_5nHBVNCkoG

pyfa_ofare0vD6s

pyfa_Uf0yBWFivJ

Added White Text Outlines & Offsetting to the X Axis Indicator for improved legibility

xaxis_indicator_text


Skybladev2:
I cannot distinguish between added fits, because there are only ammo names on graph. How to understand which graph belongs to which fit?

GET-R3AL:
> I cannot distinguish between added fits, because there are only ammo names on graph. How to understand which graph belongs to which fit?

Fits can be differentiated by the ammo they use. When using fits that use the same ammo, the graph should auto swap to pattern based ammo distinction where the lines are color coded similarly to the traditional graphs.

Issue #2740, opened 2026-04-01
Linux version crashes on startup


jasconius:
Linux version 2.66.1 AppImage crashes two or three seconds after startup as of 4/1

I strongly suspect that this is the update popup window causing the crash as the application was working fine prior to the release of 2.66.2

DarkFenX:
https://ci.appveyor.com/project/pyfa-org/pyfa/builds/54467077/job/pe242als8x4bq4qq/artifacts please try this build and let me know if it works

@tadeasf same request. It's the build with your changes

tadeasf:
<img width="1707" height="1402" alt="Image" src="https://github.com/user-attachments/assets/8cc0ac49-c183-4e1d-9edd-3d5197a69d71" />
works fine on fedora @DarkFenX would imagine this isn't anything distro-dependent

DarkFenX:
Well, the fix shouldn't be. The appimage itself is - e.g. it does not work on debian testing:
```
$ ./pyfa-v2.68.0+6-gab6bc94b0-linux.AppImage 
Failed loading SDL3 library.
Aborted                    ./pyfa-v2.68.0+6-gab6bc94b0-linux.AppImage
```
But this is old news. Issue is deeper than pyfa updater code, obviously. I will check it later.

tadeasf:
> Well, the fix shouldn't be. The appimage itself is - e.g. it does not work on debian testing:
> 
> ```
> $ ./pyfa-v2.68.0+6-gab6bc94b0-linux.AppImage 
> Failed loading SDL3 library.
> Aborted                    ./pyfa-v2.68.0+6-gab6bc94b0-linux.AppImage
> ```
> 
> But this is old news. Issue is deeper than pyfa updater code, obviously. I will check it later.

hmmm, that's interesting. i can potentially spin up a box during the weekend and test it out, but it's kinda a hassle so if you'll have infra for it already i'll happily leave it up to you

tadeasf:
@DarkFenX I think I found the chain:

1. wxPython uses `libwx_gtk3u_core-3.2.so.0` and this has a hard `DT_NEED` set for `libSDL2-2.0.so.0` (wxwidgerts uses SDL for wxSound) so it loads on `import wx`
[...]

PR #2791, opened 2026-07-26
Fix Linux startup crash caused by the update notification dialog

tadeasf:
Hi! I love pyfa - I've been using it for years. I'm primarily on Linux these days, and it would crash on startup, so I decided to fix that and give some love back to the project. More to follow if you're happy with my work (I was thinking about the dark mode request, but then again, on Linux it's already sort of there, so I'd need to switch machines... eh, we'll see).

## What / why

Addresses #2740 - the Linux AppImage dies ~2-3 s after launch with nothing in the Python log.

## Root cause

The update notification dialog (`gui/updateDialog.py`) is the only user of `wx.html2.WebView`. On wxGTK that's WebKitGTK, and:

1. **AppImage**: the builder bundles jammy's `libwebkit2gtk-4.0-37`. The library loads, but WebKit then tries to spawn its helper processes at the Ubuntu-hardcoded path `/usr/lib/x86_64-linux-gnu/webkit2gtk-4.0/WebKitNetworkProcess`, which doesn't exist on other distros - a fatal GLib error that aborts the app with no Python traceback. This is why it "appeared" with 2.66.2: `git diff v2.66.1..v2.66.3` touched nothing in the startup/update path - 2.66.2 was simply the first release new enough for the popup to run at all. Reproduced on Fedora 44 with the v2.64.2 AppImage (exit 133 = SIGTRAP):

   ```
   ** (pyfa.py:856537): ERROR **: 20:41:04.016: Unable to spawn a new child process:
   Failed to spawn child process "/usr/lib/x86_64-linux-gnu/webkit2gtk-4.0/WebKitNetworkProcess"
   (No such file or directory)
   ```

   In `~/.pyfa/pyfa.log`, every crashed launch ends exactly at `service.update: Starting Check Update Thread.`

2. **From source** on distros without webkit2gtk-4.0 (Fedora today): `import wx.html2` itself fails, so startup dies before the main window - `gui/mainFrame.py` imports the dialog at module level.

## Changes

- `gui/updateDialog.py`: never use the WebView on wxGTK (a native abort can't be caught with try/except) - release notes render in a read-only `wx.TextCtrl` instead; the Download button already opens the release page in a browser. Other platforms keep the WebView but check `IsBackendAvailable()` and guard `WebView.New()`/`SetPage()`, falling back to plain text on failure. `import wx.html2` is now lazy so startup doesn't depend on WebKitGTK being present. Two small guards while there: `releaseInfo['body']` may be `None` (would `TypeError`), and the dialog title contained a literal `{}` - it now shows the release tag.
- `gui/mainFrame.py`: `ShowUpdateBox` wrapped in try/except + logging so a dialog failure can never silently break startup again.
- `dist_assets/linux/AppImageBuilder.yml`: drop `libwebkit2gtk-4.0-37` - nothing loads it on Linux anymore. **Happy to split this into a separate PR if preferred.**

Behavior on Windows/macOS is unchanged (WebView2/WKWebView still used, now guarded).

## Verified

- Reproduced pre-fix: AppImage crash (signature above), and from-source startup failure at `import wx.html2`, both on Fedora 44.
- Post-fix: app starts and stays up from source (venv, wxPython 4.2.1 gtk3 wheels); faked an old version in `version.yml` to force the popup - app survives, dialog opens with plain-text notes; headless `UpdateDialog` construction against live release data passes; flake8 (tox pep8 flags) adds zero findings on changed files.
- Not verified: a rebuilt AppImage (needs appimage-builder on Ubuntu) - but the WebKitGTK code path is gone entirely on Linux.

## Notes

Flagging so that you know: `service/update.py` swallows the primary endpoint's error silently (lines 54-59) and assumes the response is a JSON list (line 61) - might be worth fixing later?

### DIFF
--- a/dist_assets/linux/AppImageBuilder.yml
+++ b/dist_assets/linux/AppImageBuilder.yml
@@ -30,7 +30,6 @@ AppDir:
       # wx dependencies
       - libgtk-3-0
       - librsvg2-common  # GTK3 recommendation; without it, search in char editor crashes
-      - libwebkit2gtk-4.0-37  # Needed for wx's HTML lib
       # Unknown
       - libpcre2-32-0  # https://github.com/pyfa-org/Pyfa/issues/2572
       - libnotify4  # https://github.com/pyfa-org/Pyfa/issues/2598
--- a/gui/mainFrame.py
+++ b/gui/mainFrame.py
@@ -242,8 +242,16 @@ class MainFrame(wx.Frame):
         return Fit.getCommandProcessor(fitID)
 
     def ShowUpdateBox(self, release, version):
-        with UpdateDialog(self, release, version) as dlg:
-            dlg.ShowModal()
+        # This runs right after startup, so any failure here must not be able
+        # to take the whole application down with it
+        try:
+            with UpdateDialog(self, release, version) as dlg:
+                dlg.ShowModal()
+        except (KeyboardInterrupt, SystemExit):
+            raise
+        except Exception as e:
+            pyfalog.error("Caught exception while showing the update notification dialog")
+            pyfalog.error(e)
 
     def LoadPreviousOpenFits(self):
         sFit = Fit.getInstance()
--- a/gui/updateDialog.py
+++ b/gui/updateDialog.py
@@ -21,11 +21,13 @@
 import wx
 # noinspection PyPackageRequirements
 import dateutil.parser
+from logbook import Logger
 from service.settings import UpdateSettings as svc_UpdateSettings
-import wx.html2
 import webbrowser
 import re
 import markdown2
+
+pyfalog = Logger(__name__)
 
 _t = wx.GetTranslation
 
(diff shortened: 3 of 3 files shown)


DarkFenX:
Don't do anything re dark mode (and pretty much anything else) for now. wxpython 4.3 was released 2 days ago, it will likely bring it automatically (it is backed by wxwidgets v3.3.x which should have that functionality on windows). I will work on uplifting pyfa's dependencies / python over this weekend and next week. It will result in lots of changes, likely making you to redo your work.

Will check this PR and merge if everything is fine later today.

DarkFenX:
I just checked it. Looks good, thanks.

tadeasf:
> Don't do anything re dark mode (and pretty much anything else) for now. wxpython 4.3 was released 2 days ago, it will likely bring it automatically (it is backed by wxwidgets v3.3.x which should have that functionality on windows). I will work on uplifting pyfa's dependencies / python over this weekend and next week. It will result in lots of changes, likely making you to redo your work.
> 
> Will check this PR and merge if everything is fine later today.

sure thing. I'll check next week if there is something else i might be able to pick up.
dark mode is a weird thing, on linux you can already set it up thanks to gtk iirc, so this is 100% a correct choice.

PR #2795, opened 2026-08-09
Generic Links

GET-R3AL:
Lets a fit simulate links at a chosen strength directly from the Command tab, instead of having to build and add a separate fit with command bursts and charges.

Right-click in the Command view > "Generic Links" to pick a link type and a strength from 0-5%/lvl (level V Skills assumed), with or without a mindlink. There's also a one-click "Max Links" option and a top-level "All Links".

They get injected into the fit's command bonuses through the same path real links fits use, so they interact correctly with normal links fits, can be toggled on/off, dragged around, undone/redone, etc.

Added a new commandLinks table (migration 50) to store the selections per fit

image


DarkFenX:
Thanks

PR #2798, opened 2026-08-29
Ignore volume of active fighters for fighterBayUsed

WilliamWeist:
Small logic change to ignore the volume of **active fighter squads** when calculating the volume of m3 used in the _Fighters Bay_.

In Eve the active squads are not in the _Fighters Bay_ but in the _Launch Tubes_ of the carrier or the supercarrier. When theory-crafting fits for those ships people have to take into consideration the fact that Pyfa store all the fighters in the _Fighters Bay_ and manually do the math to calculate the amount of fighters the ships need to have under optimal condition. Removing active fighter squads from the volume used in the _Fighters Bay_  let people easily simulate the action of moving fighters from the _Fighters Bay_ to the _Launch Tubes_.

### DIFF
--- a/eos/saveddata/fit.py
+++ b/eos/saveddata/fit.py
@@ -1320,7 +1320,8 @@ class Fit:
     def fighterBayUsed(self):
         amount = 0
         for f in self.fighters:
-            amount += f.item.attributes['volume'].value * f.amount
+            if not f.active:
+                amount += f.item.attributes['volume'].value * f.amount
 
         return amount
 


DarkFenX:
i thought of this long time ago, and unsure if i like it. Both simple solutions are wrong (simple being how pyfa handles it, and how you propose to handle it). Proper one would be making a tube UI where you can drag fighters, but it is more complex, and does not solve adjacent issues (more on that below).

Out of 2 methods, I prefer the one implemented in pyfa. It allows to plan fighter bay volume & fit your ship without breaking fighter bay limit with e.g. multifit. After you load fighters into the tubes, you can add some spares. This preference comes from managing a sizeable carrier-based home defense fleet for our corp.

Maybe an option to conditionally export active fighters similarly to charges can be added, then not considering active fighters can be fine. But it solves issue only partially: EFT interop format does not allow to define what's in tubes and what is not, so by default imported fit will have overloaded fighter bay, which is not nice.

So all-in-all it is an issue when it comes to interop, rather than the feature itself. And interop is something which is *very* hard to move. I did it only once when mutated items were introduced.

If you have any thoughts, do share.

